PR #3, opened 2017-03-29
Fix contributing/readme links

samneirinck:
The links to the service-fabric repo in the README.md seem to point to an invalid location. This should resolve that.


vturecek:
@samneirinck Thank you, good catch.

PR #10, opened 2017-05-18
Develop

johnbindel:
Added "Async" suffix on async methods
Minor grouping/ordering adjustments

### DIFF
--- a/src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs
+++ b/src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs
@@ -237,7 +237,7 @@ namespace Microsoft.ServiceFabric.Services.Client
             {
                 case ServicePartitionKind.Singleton:
                     {
-                        return this.ResolveAsyncHelperAsync(
+                        return this.ResolveHelperAsync(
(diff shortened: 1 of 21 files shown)


johnbindel:
This contains a breaking naming change to IActorManager's GetActorsFromStateProvider method. I renamed it to GetActorsFromStateProviderAsync to follow best practices around async method naming. If necessary, I can resubmit without that change.

amanbha:
Yes, lets not break public api surface area. Please resubmit with changes limited to internal and private methods only.

johnbindel:
Ok, I reverted that commit and addressed only private and internal methods.

src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs
```diff
@@ -506,7 +506,7 @@ public static ServicePartitionResolver GetDefault()
                 cancellationToken);
         }
 
-        private async Task<ResolvedServicePartition> ResolveAsyncHelper(
+        private async Task<ResolvedServicePartition> ResolveAsyncHelperAsync(
```
amanbha: rename to ResolveHelperAsync?

johnbindel: Done

Issue #9, opened 2017-05-18
StatefulService ctor doesn't check for null arguments


vturecek:
Service constructors should null-check arguments and throw ArgumentNullException.

Copied from Azure/service-fabric-issues#69

amanbha:
Fixed with PR https://github.com/Azure/service-fabric-services-and-actors-dotnet/pull/11

PR #11, opened 2017-05-21
Added null-check arguments to Service constructors.

Marusyk:
According to issue #9 I've added null-check arguments. Please review.


Marusyk:
Could you advise me whether I can use C # 7 syntax here?

`this.serviceContext = serviceContext ?? throw new ArgumentNullException(nameof(serviceContext));`

instead of

```
if (serviceContext == null)
{
       throw new ArgumentNullException(nameof(serviceContext));
}
this.serviceContext = serviceContext;
```

amanbha:
@Marusyk C# 7 syntax: not yet. I have merged this PR, thanks!

PR #12, opened 2017-05-23
Added implementation of IDisposable to classes that contains IDisposable members

Marusyk:
The `MockActorManager` and  `RuntimeContext` class containing IDisposable members does not itself implement IDisposable. 

Also the `traceId` field is used before it is initialized in constructor.

Please review.


amanbha:
@Marusyk Could you also please make StatelessServiceInstanceFactory and StatefulServiceReplicaFactory implement IDisposable and call dispose on RuntimeContext as they contain RuntimeContext. Also lets not change MockActorManager now, then the classes which contain it needs to be changed and that would change the public api surface area.

Marusyk:
@amanbha  Thanks you your suggestions. I've already committed it. Please check again.

PR #4, opened 2017-04-08
Created three xUnit Unit Test projects to get the ball rolling

lythix:
Created a few simple tests just to get a place for future unit tests to live and make sure they are structured correctly with right tools.

Issue #2, opened 2017-03-28
Solution Unit Tests


lythix:
I understand that this part of Service Fabric has only just been made Open Source, so it's possible any testing code may not yet have been made available. 
It's common for unit test projects to be along side the code being tested (ie in the same Solution). 

I was not sure where to ask this question so my apologies if this is the incorrect place to ask. Where are the Unit Tests located? Unit Tests are not only great for Testing, but also for new comers to help understand what individual units do and how they are supposed to work. 

Thank you.

amanbha:
Yes, you are correct, we are starting to Open Source are programming model, product source code is the first step. Tests which we have  currently are written using internal tools. Part of it is due to history. Service Fabric was developed as internal service, our development process uses tools that aren't public. To be publishable, such issues have to be resolved. Tools either need to be replaced or themselves published  We will incrementally modify tests and add them to this repo so that they can be run without any dependency on internal tools.

lythix:
Sounds excellent, looking forward to seeing the tests. Am interested in contributing and am especially interested in what is used to test, test style, mocking frameworks etc. 
Will keep an eye on this repo, I have an interest in Service Fabric but with no tests present making code changes sounds scary. 
Thank you

amanbha:
That's great! yes, of course you can add unit tests, please use NUnit for writing unit tests and do a PR into the develop branch for this repo. Thanks.

lythix:
Excellent, will do. 
I also noticed they use Moq in another Microsoft codebase I looked at (roslyn project-system), so will go with that. I've not yet figured out if each team use their own preference to tools or if Microsoft have a list of tools to try to be consistent. Visibility of that outside Microsoft isn't good.

amanbha:
Using Moq is fine, I also updated my previous post to use xUnit, we just created the develop branch. Thanks

vturecek:
@amanbha thanks!

@lythix yes you have a very valid point about standardized testing frameworks across Microsoft projects. A number of MS open source projects (across Azure and asp.net) are using either XUnit or MSTest (the test framework that ships with Visual Studio). So as @amanbha mentioned, we're jumping on the XUnit train as well.

lythix:
@amanbha Thanks for not letting me get away with anything. :)
I've learned a lot just with this one pull request. It was my first real attempt at a pull request (hopefully it didn't show too much). May there be many more better ones to come. 

[...]

Issue #1, opened 2017-03-10
Support for Xamarin


migueldeicaza:
Would it be possible to package this software in a form that can be consumed by Xamarin?

We have a customer asking for Xamarin support for Service Fabric, specifically these libraries:

`Microsoft.ServiceFabric` and  `Microsoft.ServiceFabric.Services.Wcf`

mi-hol:
@migueldeicaza @vturecek I wonder if this topic is getting discussed in an internal conversation? If yes, will you please share the result here?

amanbha:
@migueldeicaza 
What kind of support are you looking for, exactly? Can you explain the scenario in more detail?

migueldeicaza:
The packages would be compiled either as pure .NET standard, or compiled specifically for Xamarin.iOS and Xamarin.Android and published as NuGet packages.

amanbha:
it will be a pure net standard nuget package.

PR #14, opened 2017-05-24
Performance

johnbindel:
Reducing # of array resize operations in ServiceProxy.Create and FabricTransportServiceRemotingListener ctor- for InterfaceDescription, we know the size of the array at the beginning. For MethodDescription, we know the list will either be N or N-1 elements in length during normal operation. Benchmarked to demonstrate a performance gain- I can share that code as well if desired.

### DIFF
--- a/src/Microsoft.ServiceFabric.Services.Remoting/Description/InterfaceDescription.cs
+++ b/src/Microsoft.ServiceFabric.Services.Remoting/Description/InterfaceDescription.cs
@@ -67,10 +67,13 @@ namespace Microsoft.ServiceFabric.Services.Remoting.Description
             MethodReturnCheck methodReturnCheck)
         {
             EnsureValidMethods(remotedInterfaceKindName, remotedInterfaceType, methodReturnCheck);
-            var methods = remotedInterfaceType.GetMethods().Select(
-                methodInfo => MethodDescription.Create(remotedInterfaceKindName, methodInfo)).ToList();
-
-            return methods.ToArray();
+            var methods = remotedInterfaceType.GetMethods();
+            var methodDescriptions = new MethodDescription[methods.Length];
+            for (int i = 0; i < methods.Length; i++)
+            {
+                methodDescriptions[i] = MethodDescription.Create(remotedInterfaceKindName, methods[i]);
+            }
+            return methodDescriptions;
         }
 
         private static void EnsureValidMethods(
--- a/src/Microsoft.ServiceFabric.Services.Remoting/Description/MethodDescription.cs
+++ b/src/Microsoft.ServiceFabric.Services.Remoting/Description/MethodDescription.cs
@@ -61,10 +61,11 @@ namespace Microsoft.ServiceFabric.Services.Remoting.Description
      
         internal static MethodDescription Create(string remotedInterfaceKindName, MethodInfo methodInfo)
         {
-            var argumentList = new List<MethodArgumentDescription>();
+            var parameters = methodInfo.GetParameters();
+            var argumentList = new List<MethodArgumentDescription>(parameters.Length);
             var hasCancellationToken = false;
 
-            foreach (var param in methodInfo.GetParameters())
+            foreach (var param in parameters)
             {
                 if (hasCancellationToken)
                 {
--- a/src/Microsoft.ServiceFabric.Actors/Remoting/Runtime/ActorServiceRemotingDispatcher.cs
+++ b/src/Microsoft.ServiceFabric.Actors/Remoting/Runtime/ActorServiceRemotingDispatcher.cs
@@ -61,18 +61,18 @@ namespace Microsoft.ServiceFabric.Actors.Remoting.Runtime
         {
             if (messageHeaders.InterfaceId == ActorMessageDispatch.InterfaceId)
             {
-                return this.HandleActorMethodDispatch(messageHeaders, requestBodyBytes);
+                return this.HandleActorMethodDispatchAsync(messageHeaders, requestBodyBytes);
             }
 
             if (messageHeaders.InterfaceId == ActorEventSubscription.InterfaceId)
             {
-                return this.HandleSubscriptionRequests(requestContext, messageHeaders, requestBodyBytes);
+                return this.HandleSubscriptionRequestsAsync(requestContext, messageHeaders, requestBodyBytes);
             }
 
             return base.RequestResponseAsync(requestContext, messageHeaders, requestBodyBytes);
         }
 
-        private async Task<byte[]> HandleSubscriptionRequests(
+        private async Task<byte[]> HandleSubscriptionRequestsAsync(
             IServiceRemotingRequestContext requestContext,
             ServiceRemotingMessageHeaders messageHeaders,
             byte[] requestMsgBodyBytes)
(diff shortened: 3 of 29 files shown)


src/Microsoft.ServiceFabric.Services.Remoting/Description/MethodDescription.cs
```diff
@@ -61,10 +61,11 @@ public MethodInfo MethodInfo
      
         internal static MethodDescription Create(string remotedInterfaceKindName, MethodInfo methodInfo)
         {
-            var argumentList = new List<MethodArgumentDescription>();
+            var parameters = methodInfo.GetParameters();
+            var argumentList = new List<MethodArgumentDescription>(parameters.Length);
             var hasCancellationToken = false;
 
-            foreach (var param in methodInfo.GetParameters())
+            foreach (var param in parameters)
```
amanbha: Looking at the IL, the GetParameters() is done before loop the loop so it gets optimized already.

johnbindel: Sure. We can initialize argumentList size to parameters.Length to reduce list resize operations because we know the argumentList.Length will either be parameters.Length (if we have no CancellationToken parameter) or parameters.Length-1 (if we have a CancellationToken parameter). Otherwise, the underlying array in the list will resize from 0->4->8->16->etc. while the list is being added to.

src/Microsoft.ServiceFabric.Services.Remoting/Description/InterfaceDescription.cs
```diff
@@ -67,10 +67,13 @@ public MethodDescription[] Methods
             MethodReturnCheck methodReturnCheck)
         {
             EnsureValidMethods(remotedInterfaceKindName, remotedInterfaceType, methodReturnCheck);
-            var methods = remotedInterfaceType.GetMethods().Select(
-                methodInfo => MethodDescription.Create(remotedInterfaceKindName, methodInfo)).ToList();
-
-            return methods.ToArray();
+            var methods = remotedInterfaceType.GetMethods();
```
amanbha: Could you explain what causes perf gain here? Is the perf gain  because Linq .ToArray() is not performant. It would be great if you can share your perf numbers and the perf code as well.

johnbindel: Absolutely. In the existing code, we are:
1) getting the array of methods
2) projecting the array to an IEnumerable, with .Select(...)
3) creating a List (with resizes from 0->4->8->16->etc. on the internal array because ToList only honors the Count when it is an ICollection which in this case, it is not)
4) creating an array from the list (another array and copy operation)

Since we know the length of the array, we can make a single array of the projected type and avoid all extraneous resize and copy operations.

I can submit the code to look at performance this evening.

amanbha: I agree with #4, ToList() and then ToArray() is redundant. It can just be ToArray(), IL code is doing it. #1 and #2 are fine. 
So Its just #3 which causes list to be resized as IEnumerable returned from Select is not ICollection? If yes, then lets make the change.

johnbindel: Correct, it is just an IEnumerable, not an ICollection because elements are returned with yield return syntax. Here's a concrete example. If we have 5 methods on the interface:
1. GetMethod returns an array of length 5
2. Select will produce an IEnumerable (deferred)
3. ToList will create a new list with array length 0, then create an array of length 4 and fill it with the first 4 elements of the enumerable, then when adding the 5th: it will create a new array of length 8, copy the 4 elements from the other array, add the 5th element to the new array
4. ToArray will create a new array of length 5 and copy the 5 elements over

Combining the ToList and ToArray calls into just a ToArray call, would continue to have the same inefficiencies as IEnumerable.ToArray does not know the size until after iteration has happened. It would still create unnecessary arrays and copy on each grow.

Since we know the size of the output array I have recommended this change. It would work this way:
1. GetMethod returns an array of length 5
2. We create a new array of 5
3. We iterate and project the elements into the new array

amanbha: Sounds good @johnbindel , thanks for explaining with concrete example!

johnbindel: Happy to help out where I can! Thanks @amanbha

johnbindel: The performance test I wrote is very simple:
```
    public interface ITestService : IService
    {
        Task<string> Test1Async(object o, int[] i);
        Task<int> Test2Async(object o, int[] i);
        Task<long> Test3Async(object o, int[] i);
        Task<Guid> Test4Async(object o, int[] i);
        Task<object> Test5Async(object o, int[] i);
    }
```

    [TestClass]
    public class PerformanceBenchmarks
    {
        [TestMethod]
        public void DescriptionCreationBenchmark()
        {
            var stopwatch = new Stopwatch();
            stopwatch.Start();

            for (int i = 0; i < 100000; i++)
            {
                var description = ServiceInterfaceDescription.Create(typeof (ITestService));
            }

            stopwatch.Stop();
        }
    }

In my benchmarks, this led to 8-10% reduction in time for the ServiceInterfaceDescription.Create call. Each individual call is not expensive, but we can utilize memory and CPU slightly more optimally without convoluted code.

image

amanbha: Looks good @johnbindel

src/Microsoft.ServiceFabric.Services.Remoting/Description/MethodDescription.cs
amanbha: @johnbindel Should i merge this PR or wait for you to make the above change in this PR.

johnbindel: The change I described was done in the initial PR. It's ready to be merged.

amanbha: those changes are for InterfaceDescription.cs, the changes mentioned above should be in MethodDescription.cs. Chnages in MethodDescription.cs is moving methodInfo.GetParameters() before the loop which doesn't help with perf as compiler moves it outside of the loop as optimization.

johnbindel: My comment above about list growing was referencing this code change. In this example, we know argumentList will either be parameters.Length or parameters.Length-1. Since we know the maximum size, we can create our list with capacity of parameters.Length to avoid resizing from 0->4->8->16->etc.
The important change here is:
var argumentList = new List<MethodArgumentDescription>(**parameters.Length**);

amanbha: ah, my bad, i was missing this line var argumentList = new List(parameters.Length);

PR #33, opened 2017-07-20
Service Context propagation for downstream consumption by telemetry systems.

nizarq:
This change add the ability in the framework to communicate service context downstream for telemetry systems to use. 

The strategy of context propagation depends on the hosting model a service is using. If the service is using Exclusive model, the context would be propagated using environment variables. If on the other hand, the service is using shared model, a process wide mechanism such as env vars don't work, so we pass it on the logical call context. 

Further the user can specify a setting to disable any context propagation. 

This change will enable [Application Insights Service Fabric SDK](https://github.com/Microsoft/ApplicationInsights-servicefabric) to be able to automatically determine the right context that it can appent to all telemetry generated through various AI modules.


nizarq:
@suchiagicha @VipulM-MSFT @harahma

nizarq:
@brahmnes @karolz-ms

PR #34, opened 2017-08-05
Added support for ServiceBehavior(InstanceContextMode = InstanceConte…

tommyja:
…xtMode.PerSession) to WcfCommunicationListener

### DIFF
--- a/src/Microsoft.ServiceFabric.Services.Wcf/Communication/Wcf/Runtime/WcfCommunicationListener.cs
+++ b/src/Microsoft.ServiceFabric.Services.Wcf/Communication/Wcf/Runtime/WcfCommunicationListener.cs
@@ -81,6 +81,35 @@ namespace Microsoft.ServiceFabric.Services.Communication.Wcf.Runtime
 
         /// <summary>
         ///     Constructs a WCF based communication listener that uses specified listener binding and 
+        ///     endpoint address derived from the specified endpoint resource name.
+        /// </summary>
+        /// <param name="serviceContext">
+        ///     The context of the service for which this communication listener is being constructed.
+        /// </param>
+        /// <param name="wcfServiceType">
+        ///     Type of WCF service implementing the specified WCF service contract.
+        /// </param>
+        /// <param name="listenerBinding">
+        ///     The binding to use for the WCF endpoint. If the listenerBinding is not specified or it is null, a default listener binding is 
+        ///     created using <see cref="WcfUtility.CreateTcpListenerBinding"/> method.
+        /// </param>
+        /// <param name="endpointResourceName">
+        ///     The name of the endpoint resource defined in the service manifest that should be used to create the address for the listener. 
+        ///     If the endpointResourceName is not specified or it is null, its name is derived from the WCF service contract type using
+        ///     <see cref="Microsoft.ServiceFabric.Services.ServiceNameFormat.GetEndpointName"/> method.
+        ///     If matching endpoint resource is not found in the service manifest, a default endpoint resource definition with port zero is used.
+        /// </param>
+        public WcfCommunicationListener(
+            ServiceContext serviceContext,
+            Type wcfServiceType,
+            Binding listenerBinding = null,
+            string endpointResourceName = null)
+            : this(serviceContext, wcfServiceType, listenerBinding, null, endpointResourceName)
+        {
+        }
+
+        /// <summary>
+        ///     Constructs a WCF based communication listener that uses specified listener binding and 
         ///     endpoint address derived from the specified endpoint address.
         /// </summary>
         /// <param name="serviceContext">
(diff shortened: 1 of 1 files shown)


shabanovanton:
Hello,

Are there any updates on this matter? I really would like to have an opportunity to instantiate WCF services per call so I'm also interested when this request to be merged and released into nuget!

Apart from that, was it intended that the only supported mode is Single? Are there any limitations inside SF which don't allow to use PerCall mode currenly?

tommyja:
@shabanovanton. I see no reason why it shouldn't work. I am using a nuget package locally built from my clone of the repo and that works fine in the dev cluster.

tommyja:
closed by acciden

shabanovanton:
@tommyja This was a question to MS team, I wonder why they didn't initially do it the way you've done it now:) But thank you for your contribution, I think I don't have other option rather than using your solution for now!

shabanovanton:
@tommyja UPD: worked like a charm, I can now use InstanceContextMode.PerCall and utilise autofac to inject decencies into the service object.

@vturecek I see you have assigned this on yourself, is there any ETA it will be released officially?

tommyja:
The client already supported callbacks so it is odd that that the only supported mode is InstanceContextMode.Single on the server side...

kevinhillinger:
@tommyja @shabanovanton It looks like there's no issue tied to this pull request. That's the first step. Second, there are code changes, but no unit tests. Ideally, those would be there.

I will create a ticket to track this, and assist getting this in. @tommyja would you be willing to redo the PR and we can review?

tommyja:
I could not find any existing unit tests for the WcfCommunicationListener. It is not easy to come up with meaningful unit tests for this. It really needs integration tests or a way to mock the WCF stack.

VipulM-MSFT:
@tommyja - sorry this has taken this long, it should not have. 

One of the reason for having a singleton previously was to associate a single object with the single stateful or stateless (mainly stateful) service replica object. In this case the replica object should not be the one implementing the WCF service contract. How do you use it in your scenario?

tommyja:
@VipulM-MSFT The old constructors work as before so my changes should have no impact on the scenario you describe. The new constructors also allow any mode by applying the ServiceBehaviorAttribute to the service implementation class. In my scenario i push data from a server to a client through a callback which requires instancecontextmode persession.

VipulM-MSFT:
@tommyja - Thank you for making change in a compatible manner. I wanted to understand your scenario. What would be the pattern if someone wants to access the service object so that they can access state associated with the statefulservice replica in the WCF service instance?

tommyja:
@VipulM-MSFT I don't understand the question. Are you talking about the situation when a service goes down and a replica gets promoted to primary? If this happens, the WCF session goes down and the client must resolve the new wcf endpoint and establish a new session.

VipulM-MSFT:
@tommyja - if the WCF communication listener is a communication endpoint for a stateful service, you need to attach the stateful service to the instance that the host create. I was asking how you are doing than, one way to do this is to write instance provider and dispatch behavior, like below:

```csharp
    [ServiceContract]
    public interface ICounterService
    {
        [OperationContract]
        Task<long> GetCountAsync();
    }

    internal class WcfCounterService : ICounterService
    {
        private readonly IStatefulCounterService stateful;

        internal WcfCounterService(IStatefulCounterService stateful)
        {
            this.stateful = stateful;
        }

        public Task<long> GetCountAsync()
        {
            return this.stateful.GetCountAsync();
        }
    }

    internal class CounterServiceInstanceProvider : IInstanceProvider
    {
        private readonly IStatefulCounterService stateful;

        public CounterServiceInstanceProvider(IStatefulCounterService stateful)
        {
            this.stateful = stateful;
        }

        public object GetInstance(InstanceContext instanceContext)
        {
            return new WcfCounterService(stateful);
        }

        public object GetInstance(InstanceContext instanceContext, Message message)
        {
            return new WcfCounterService(stateful);
        }

        public void ReleaseInstance(InstanceContext instanceContext, object instance)
        {
            // no-op
        }
    }

    internal class CounterServiceInstanceProviderBehavior : IServiceBehavior
    {
        private readonly IStatefulCounterService stateful;

        internal CounterServiceInstanceProviderBehavior(IStatefulCounterService stateful)
        {
            this.stateful = stateful;
        }

        public void AddBindingParameters(
            ServiceDescription serviceDescription, 
            ServiceHostBase serviceHostBase, 
            Collection<ServiceEndpoint> endpoints, 
            BindingParameterCollection bindingParameters)
        {
        }

        public void ApplyDispatchBehavior(
            ServiceDescription serviceDescription, 
            ServiceHostBase serviceHostBase)
        {
            foreach (var dispatcher in serviceHostBase.ChannelDispatchers)
            {
                var channelDispatcher = dispatcher as ChannelDispatcher;
                if (channelDispatcher != null)
                {
                    foreach (var endpoint in channelDispatcher.Endpoints)
                    {
                        endpoint.DispatchRuntime.InstanceProvider = 
                            new CounterServiceInstanceProvider(stateful);
                    }
                }
            }

        }

        public void Validate(
            ServiceDescription serviceDescription, 
            ServiceHostBase serviceHostBase)
        {
        }
    }
```

Then this needs to be hooked up with the instance in a following way:

```csharp
    internal sealed class Stateful1 : StatefulService, IStatefulCounterService
    {
        public Stateful1(StatefulServiceContext context)
            : base(context)
        { }

       protected override IEnumerable<ServiceReplicaListener> CreateServiceReplicaListeners()
        {
            var serviceReplicaListener = new ServiceReplicaListener((context) =>
            {
                var wcfListener = new WcfCommunicationListener<ICounterService>(
                    wcfServiceType: typeof(WcfCounterService),
                    serviceContext: this.Context,
                    listenerBinding: WcfUtility.CreateTcpListenerBinding(),
                    endpointResourceName: "ServiceEndpoint");
                wcfListener.ServiceHost.Description.Behaviors.Add(new CounterServiceInstanceProviderBehavior(this));
                return wcfListener;
            });
            return new ServiceReplicaListener[1]
            {
                serviceReplicaListener
            };
        }

... 
      }
```

tommyja:
@VipulM-MSFT - Thanks for the insight. I bypassed most of that by setting ServicePackageActivationMode="ExclusiveProcess" in the application manifest, and then getting at the state through a singleton. I will test with your suggestion.

Issue #35, opened 2017-08-09
Customizable IActorTimers


Jm-Fox:
IActorTimer provides a nice abstraction over timers, but it is not customizable. ActorBase always uses the internal implementation ActorTimer:
https://github.com/Azure/service-fabric-services-and-actors-dotnet/blob/master/src/Microsoft.ServiceFabric.Actors/Runtime/ActorBase.cs#L274

There should be an overload which roughly does the same thing:
```
protected void RegisterTimer(IActorTimer timer)
{   
    if (this.GcHandler.IsGarbageCollected)
    {
        throw new ObjectDisposedException("actor");
    }

    if (this.timers == null)
    {
        this.timers = new List<IActorTimer>();
    }
	
    this.timers.Add(timer);
}
```

This issue is very important to me because System.Threading.Timer does not perform well at high frequencies, which I require for a game loop. Testing on my (weak) desktop indicates that under no circumstances can Timers ever reach 60 FPS with callbacks (seems to cap out at 42 FPS). Meanwhile using Monitor.Wait() or Thread.Sleep() allow me to keep 60 FPS on hundreds of simluated game loops.
My desktop processor is not nearly as powerful or good at multithreading as the processors my servers will have, but the performance difference is too large to discount.

Update: formatted code

VipulM-MSFT:
@Jm-Fox - Definitely this is something that can be customizable. Would it be possible for you to suggest the fix and a PR for the issue?

Jm-Fox:
Yes, I've forked the repo and created pull request #36.

vturecek:
@Jm-Fox thanks! Looks pretty good to me.
@harahma can you review real quick and merge if it looks good to you?

Jm-Fox:
I'm not attached to any implementations, I'll probably make several and use whichever turns out to have the best performance. I'll try both Monitor.Wait() and Thread.Sleep(), though I'm not fond of either of those.

amanbha:
PR #36

amanbha:
@harahma please look at pull request #36 for this.

VipulM-MSFT:
@Jm-Fox - do you still need this feature? @amanbha - can the PR be modified to provide the needed enhancement without sacrificing the turn based concurrency?

Jm-Fox:
Actually, no, I don't still need this feature. I'll go down one of the two following paths:

1. Implement my game loop critical code in a partitioned StatefulService instead of an Actor
[...]

PR #36, opened 2017-08-09
Update ActorBase.cs

Jm-Fox:
Added method for registering any IActorTimer. This way ActorBase implementations are not restricted to System.Threading.Timer for timers.

### DIFF
--- a/src/Microsoft.ServiceFabric.Actors/Runtime/ActorBase.cs
+++ b/src/Microsoft.ServiceFabric.Actors/Runtime/ActorBase.cs
@@ -277,6 +277,26 @@ namespace Microsoft.ServiceFabric.Actors.Runtime
         }
 
         /// <summary>
+        /// Registers any ITimer object for the actor.
+        /// </summary>
+        /// <param name="timer">
+        /// ITimer to register
+        /// </param>
+        protected void RegisterTimer(IActorTimer timer)
+        {   
+            if (this.GcHandler.IsGarbageCollected)
+            {
+                throw new ObjectDisposedException("actor");
+            }
+
+            if (this.timers == null)
+            {
+                this.timers = new List<IActorTimer>();
+            }
+            this.timers.Add(timer);
+        }
+
+        /// <summary>
         /// Registers a reminder with the actor.
         /// </summary>
         /// <param name="reminderName">The name of the reminder to register. The name must be unique per actor.</param>


Jm-Fox:
Should have included in the commit comment: #35

vturecek:
@Jm-Fox sorry for the delay. Looks pretty straight-forward to me. Any particular timer implementations you had in mind?

harahma:
@Jm-Fox Thanks for working on this. I had a look at your changes. 

When the custom timer will fire it will need some hook provided by framework to be able to dispatch its callback to execute under its owning actors' concurrency lock.

Do you have any idea in mind for this? 

I think we should split the current implementation of ActorTimer class into an abstract ActorTimerBase base class (which will provide a protected method to dispatch timer callback under concurrency lock) and a concrete class ActorTimer which derives from ActorTimerBase and implements IActorTimer.

If users wants to provide their custom implementation of IActorTimer, their implementation will inherit from ActorTimerBase class and implement IActorTimer interface and use base classes' method to dispatch timer callback.

Jm-Fox:
Apologies for the late reply; I did not have a lot of hobby time outside of work in the past few weeks.

For my purposes, I wasn't worried about the concurrency lock. My system will be able tolerate any inconsistencies that arise from race conditions involving custom timers.

That said, I do think that it's important to be able to use the concurrency lock, so I would make that optional. Maybe have the ActorTimerBase look something like this:

```
public abstract class ActorTimerBase
{
    private readonly Func<object, Task> asyncCallback;
    private readonly ActorMethodContext methodContext;
    private readonly ActorBase owner;

    public CancellationToken CancellationToken { get; set; } = CancellationToken.None;
    public bool IgnoreConcurrencyLock { get; set; } = false;
    public object State { get; set; } = null;

    protected ActorTimerBase(ActorBase owner, Func<object, Task> asyncCallback)
    {
        this.owner = owner;
        this.asyncCallback = asyncCallback;
        this.State = state;
        this.methodContext = ActorMethodContext.CreateForTimer(asyncCallback.GetMethodInfo().Name);
    }

    // Returns true if actor is disposed
    protected bool FireCallback()
    {
        try
        {
            if (this.IgnoreConcurrencyLock)
            {
                await this.asyncCallback.Invoke(State);
            }
            else
            {
                // Copy/pasted from current ActorTimer.cs (except CancellationToken)
                await this.owner.Manager.DispatchToActorAsync(
                    this.owner.Id,
                    this.callbackMethodContext,
                    false,
                    this.DispatchTimerCallback,
                    Guid.NewGuid().ToString(),
                    true,
                    this.CancellationToken);
            }
        }
        catch (ObjectDisposedException)
        {
            return true;
        }
        return false;
    }

    // Copy/pasted from current ActorTimer.cs
    private async Task<byte[]> DispatchTimerCallback(ActorBase actor, CancellationToken cancellationToken)
    {
        if (!ReferenceEquals(actor, this.owner))
        {
            throw new ObjectDisposedException("actor");
        }
        var callback = this.asyncCallback;
        if (callback != null)
        {
            await callback(this.callbackState);
        }
        return null;
    }
}
```

Things to note:
- State is mutable
- Optional mutable CancellationToken
- Base class does not handle rescheduling. Inherited classes will call FireCallback when necessary, and determine whether or not to reschedule on their own based on the return value (and maybe the state if they so choose)
- No actual timer logic is here; IDisposable not required. The existing ActorTimer can inherit the class and provide that implementation
- Constructor only requires immutable fields. Inherited classes can choose how picky they want to be with optional fields

amanbha:
@Jim-Fox Sorry for a delayed response on this one, we took a look at it and discussed about it and turn based concurrency is inherent to our Actor model and we can't relax this restriction as it will cause  confusion for users. Additionally it will make debugging difficult when providing support without code snippet from user as now concurrency can be bypassed for timers on demand.
So please update the PR without support for IgnoreConcurrencyLock.

amanbha:
Closing this PR, feel free to submit again with the suggestions.

PR #38, opened 2017-08-11
allow custom actor activator

grahamehorner:
changes to allow custom actor activator


vturecek:
@grahamehorner sorry for the delayed response. Is this a factory for the actor factory? What sort of use cases for this did you have in mind?

grahamehorner:
I'm think of a use case where the Actor Activation would be from an external trigger like a message queue or a file system operation.

grahamehorner:
I'm going to possibly have a rethink and write some custom attributes similar to the azure function  bindings

amanbha:
@grahamehorner Do you have any updates for this PR as per your last comment or can i close it for now?

PR #44, opened 2017-09-17
Fixed async/await misuses

Marusyk:
Fixed some async/await misuses. Please review. Thank you in advance

### DIFF
--- a/src/Microsoft.ServiceFabric.Actors/Client/ActorProxy.cs
+++ b/src/Microsoft.ServiceFabric.Actors/Client/ActorProxy.cs
@@ -325,7 +325,6 @@ namespace Microsoft.ServiceFabric.Actors.Client
             {
                 await this.SubscribeAsyncV2(eventType, subscriber, resubscriptionInterval);
                 return;
-                ;
             }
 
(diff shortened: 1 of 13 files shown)


Mardoxx:
Why were these misuses?

Marusyk:
Hi @Mardoxx 
Adding the async modifier comes at a price. There are some async methods where there is no need to use async/await keywords.  For excample 
[UnregisterReminderAsync](https://github.com/Marusyk/service-fabric-services-and-actors-dotnet/blob/cb812639e86d7cef6a06f225ccebaf59556a3c73/src/Microsoft.ServiceFabric.Actors/Runtime/ActorBase.cs#L235)
```C#
protected async Task UnregisterReminderAsync(IActorReminder reminder)
{
	this.CheckIfReminderOperationIsPossible(reminder.Name);
	await this.Manager.UnregisterReminderAsync(reminder.Name, this.Id, removeFromStateProvider: true);
}
```
The thing is, the state machine created. When a method only awaits one operation and does so at the end of the method, it's usually not worth it. Instead, we can directly return the task, as shown in the first method, and avoid the cost of the async state machine altogether.

PR #47, opened 2017-09-30
Merging code for 2.8.211 nuget packages.

amanbha:
Merging code for 2.8.211 nuget packages.


Scythen:
This merge introduces an issue in GetActorReference. The previous version used the public IActorProxy interface which allowed custom implementations to provide an actor proxy implementation.
Now this method uses the ActorProxy class, which does not allow a custom implementation.
Additionally, this method could be written such that IActorProxy would still be supported.

amanbha:
@Scythen This release is already out, I see that you have opened issue for it, will track it there for our next release.

PR #55, opened 2017-10-27
Make *ProxyFactory compat with V1 and V2 listeners (DO NOT MERGE)

Mardoxx:
See https://github.com/Azure/service-fabric-issues/issues/570

ActorProxyFactory needs fixing up too. Although I want discussion on this first please!

### DIFF
--- a/src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
+++ b/src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
@@ -91,65 +91,45 @@ namespace Microsoft.ServiceFabric.Services.Remoting.Client
         {
             var serviceInterfaceType = typeof(TServiceInterface);
 
-#if !DotNetCoreClr
-
             //Use provider to find the stack
-            if (this.proxyFactoryV1 == null && this.proxyFactoryV2 == null)
-            {
-                var provider = this.GetProviderAttribute(serviceInterfaceType);
-                if (provider.RemotingClient.Equals(RemotingClient.V2Client))
-                {
-                    //We are overriding listenerName since using provider we can have multiple listener configured(Compat Mode).
-                    this.overrideListenerName = true;
-                    this.proxyFactoryV2 =
-                        new V2.Client.ServiceProxyFactory(provider.CreateServiceRemotingClientFactoryV2,this.retrySettings);
-                }
-                else
-                {
-                    this.proxyFactoryV1 = new V1.Client.ServiceProxyFactory(provider.CreateServiceRemotingClientFactory,this.retrySettings);
-                }
-                
-           }
+            var provider = this.GetProviderAttribute(serviceInterfaceType);
 
-            if (this.proxyFactoryV1 != null)
+#if !DotNetCoreClr
+            if (provider.RemotingClient.Equals(RemotingClient.V1Client))
             {
+                // Create V1 proxyFactory
+                if (this.proxyFactoryV1 == null)
+                {
+                    this.proxyFactoryV1 =
+                        new V1.Client.ServiceProxyFactory(provider.CreateServiceRemotingClientFactory, this.retrySettings);
+                }
+
                 return this.proxyFactoryV1.CreateServiceProxy<TServiceInterface>(serviceUri,
                     partitionKey,
                     targetReplicaSelector,
                     listenerName);
             }
-            if (this.overrideListenerName && listenerName==null)
+#endif
+
+            // Create V2 proxyFactory
+            if (this.proxyFactoryV2 == null)
+            {
+                this.proxyFactoryV2 =
+                    new V2.Client.ServiceProxyFactory(provider.CreateServiceRemotingClientFactoryV2, this.retrySettings);
+            }
+
+            if (listenerName == null)
             {
                 return this.proxyFactoryV2.CreateServiceProxy<TServiceInterface>(serviceUri,
                     partitionKey,
                     targetReplicaSelector,
                     ServiceRemotingProviderAttribute.DefaultV2listenerName);
             }
+
             return this.proxyFactoryV2.CreateServiceProxy<TServiceInterface>(serviceUri,
                 partitionKey,
                 targetReplicaSelector,
                 listenerName);
-#else
-            if (proxyFactoryV2==null)
-            {
-                var provider = this.GetProviderAttribute(serviceInterfaceType);
-                //We are overriding listenerName since using provider we can have multiple listener configured(Compat Mode).
-                this.overrideListenerName = true;
-                this.proxyFactoryV2 =
-                    new V2.Client.ServiceProxyFactory(provider.CreateServiceRemotingClientFactoryV2);
-            }
-            if (this.overrideListenerName && listenerName == null)
-            {
-                return this.proxyFactoryV2.CreateServiceProxy<TServiceInterface>(serviceUri,
-                    partitionKey,
-                    targetReplicaSelector,
-                    ServiceRemotingProviderAttribute.DefaultV2listenerName);
-            }
-            return this.proxyFactoryV2.CreateServiceProxy<TServiceInterface>(serviceUri,
-              partitionKey,
-              targetReplicaSelector,
-              listenerName);
-#endif
         }
 
 


src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
```diff
@@ -91,65 +91,45 @@ public ServiceProxyFactory(OperationRetrySettings retrySettings = null)
         {
             var serviceInterfaceType = typeof(TServiceInterface);
 
-#if !DotNetCoreClr
-
             //Use provider to find the stack
-            if (this.proxyFactoryV1 == null && this.proxyFactoryV2 == null)
```
Mardoxx: Why this check? This way we break interop for wherer you need orchestration between V1 and V2 services. Either V1 factory is created, or V2 -- never both.

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
```diff
@@ -91,65 +91,45 @@ public ServiceProxyFactory(OperationRetrySettings retrySettings = null)
...
+            var provider = this.GetProviderAttribute(serviceInterfaceType);
+
+#if !DotNetCoreClr
```
Mardoxx: I assume V1 listener is not available on netcore. Is this correct?

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
```diff
@@ -91,65 +91,45 @@ public ServiceProxyFactory(OperationRetrySettings retrySettings = null)
...
-#else
-            if (proxyFactoryV2==null)
+#endif
+
+            // Create V2 proxyFactory
+            if (this.proxyFactoryV2 == null)
             {
-                var provider = this.GetProviderAttribute(serviceInterfaceType);
-                //We are overriding listenerName since using provider we can have multiple listener configured(Compat Mode).
-                this.overrideListenerName = true;
```
Mardoxx: What is `overrideListenerName` for? Looks like it was just used to determine if we were using V2 listener over V1.

It is not needed in my code so perhaps remove defn for it.

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
```diff
@@ -91,65 +91,45 @@ public ServiceProxyFactory(OperationRetrySettings retrySettings = null)
...
                 this.proxyFactoryV2 =
-                    new V2.Client.ServiceProxyFactory(provider.CreateServiceRemotingClientFactoryV2);
```
Mardoxx: Why were retry settings not passed?

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
```diff
@@ -91,65 +91,45 @@ public ServiceProxyFactory(OperationRetrySettings retrySettings = null)
         {
             var serviceInterfaceType = typeof(TServiceInterface);
 
-#if !DotNetCoreClr
-
             //Use provider to find the stack
-            if (this.proxyFactoryV1 == null && this.proxyFactoryV2 == null)
+            var provider = this.GetProviderAttribute(serviceInterfaceType);
```
Mardoxx: We get provider attribute each time service proxy factory is called. This is not a hot path, is it? So I don't think it matters for perf here?

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
```diff
@@ -91,65 +91,45 @@ public ServiceProxyFactory(OperationRetrySettings retrySettings = null)
...
-            return this.proxyFactoryV2.CreateServiceProxy<TServiceInterface>(serviceUri,
-                partitionKey,
-                targetReplicaSelector,
-                listenerName);
-#else
-            if (proxyFactoryV2==null)
+#endif
+
+            // Create V2 proxyFactory
+            if (this.proxyFactoryV2 == null)
```
Mardoxx: Here we default to V2 factory. I thought this was better to assume V2 client rather than doing, if V1 client, else if v2 client, else throw. Because the last branch "should never happen"  - famous last words..

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
amanbha: Its just checking if none of the V1 and V2 factory is provided,c reate from the attribute.

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
amanbha: This is called when creating the proxy.

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
amanbha: yes, this is correct

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
amanbha: Both the scenarios are equally possible, when users just upgrade to new nuget package for existing services V1 is default unless they specify the attribute. Starting with new templates in VS V2 will be made default by specifying the attribute.

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
amanbha: Its already possible to talk to a V1 or V2 listener https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-reliable-services-communication-remoting#how-to-upgrade-from-remoting-v1-to-remoting-v2

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
amanbha: This is to figure out wether to override the listenerName is a named listener is provided.

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
Mardoxx: Right! Which is fine, as long as this factory is only resued for services which use the same stacks (or have stacks that are compatible with each other).

However, ServiceProxy uses a [single instance](https://github.com/Mardoxx/service-fabric-services-and-actors-dotnet/blob/a66d4accddb40d214956bdb67db4ca9300fc9e62/src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxy.cs#L29).

The trouble with this is, if I do the following

```csharp
IV1Svc svc1 = ServiceProxy.Create<IV1Svc>(/*...*/);

// Internal DefaultServiceProxy is initialised with `proxyFactoryV1 != null`

IV2Svc svc2 = ServiceProxy.Create<IV2Svc>(/*...*/);

// When calling svc2.someMethod, it will not know to use v2 stack.
```

Do you see where I'm coming from here?

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
Mardoxx: Seeing as netcore only supports V2, it is more succinct to just assume that the default stack is V2 stack, rather than having another check for the provider. This way the only code that checks the attribute is in non-net-core code. Hope that makes sense?

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
Mardoxx: I think there's some confusion here, due to me not explaining why I've made these changes. `ServiceProxy` static methods use static instance of `ServiceProxyFactory`, if we use `ServiceProxy` to orchestrate between v1 and v2 services, you can not, as the internal DefaultProxyFactory will be initialised to either a v1 factory or a v2 factory. See below comment also!

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
Mardoxx: Ah, I see, it's so you can figure out if an implementaion has been given or not in the constructor. IF not, then it uses a default V2 implementation which requires a listener name of `ServiceRemotingProviderAttribute.DefaultV2listenerName`

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
Mardoxx: Fixed in https://github.com/Azure/service-fabric-services-and-actors-dotnet/pull/56

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
Mardoxx: Thanks (also thank for all your replies :smile:)

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
amanbha: Thanks for the PR, its merged.

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
amanbha: yes, i see your point

src/Microsoft.ServiceFabric.Services.Remoting/Client/ServiceProxyFactory.cs
amanbha: Thanks for explaining, yes, i see your point when you want to use the same factory to talk to one service with V1 and then talk to other service with V2. When this was designed, it assumed one factory for V1 and other for V2.

Mardoxx: That would make sense! You can close this PR if you like, unless you want me to do anything with it?

If V1 stack is being deprecated in the near future, I assume you have some sort of plan for the ServiceProxy class, respective factories (and similarly for Actors). It would seem at this point, any change would be a breaking change.

I'd perhaps deprecate `Microsoft.ServiceFabric.Services.Remoting.Client.ServiceProxyFactory`. Move the `IService` provider (stack selection) logic to `ServiceProxy.Create<>()` and have this use the appropriate default static instance of `Remoting.V2.Client.ServiceProxyFactory` (or v1 if not using netcore). I have a very naive view of SF inner workings though, so this may be a bad idea!

Thanks again for all your responses!

amanbha: Sure @Mardoxx . We would try to deprecate V1 in a non breaking way after the customers have moved the services and client to v2 stack. Thanks!.

Issue #83, opened 2018-02-27
IServiceRemotingResponseMessageBodySerializer should use IMessageBody and not OutgoingMessageBody/IncomingMessageBody


VipulM-MSFT:
```csharp
 OutgoingMessageBody Serialize(IServiceRemotingResponseMessageBody serviceRemotingRequestMessageBody);

IServiceRemotingResponseMessageBody Deserialize(IncomingMessageBody messageBody);
```

**should be**

```csharp
IMessageBody Serialize(IServiceRemotingResponseMessageBody serviceRemotingRequestMessageBody);

IServiceRemotingResponseMessageBody Deserialize(IMessageBody messageBody);

Marusyk:
Can be closed.
Thanks all

Issue #82, opened 2018-02-27
IServiceRemotingRequestMessageBodySerializer should use IMessageBody and not OutgoingMessageBody/IncomingMessageBody


VipulM-MSFT:
```csharp 
OutgoingMessageBody Serialize(IServiceRemotingRequestMessageBody serviceRemotingRequestMessageBody);
```

**should be** 
```csharp
IMessageBody Serializer(IServiceRemotingRequestMessageBody serviceRemotingRequestMessageBody);
```

and
```csharp
IServiceRemotingRequestMessageBody Deserialize(IncomingMessageBody messageBody);
```

**should be** 

```csharp
IServiceRemotingRequestMessageBody Deserialize(IMessageBody messageBody);
```

Marusyk:
Hello @VipulM-MSFT, can I create PR for that or it should be discussed first?

VipulM-MSFT:
@Marusyk - sure. We are currently fixing all of the coding style related issues to ensure consistency in the code. If you would like to wait until those are in that's fine as well as it will be easier to merge. We expect all style related fixes should be done in a week or so. We are fixing the documentation comments as part of this as well.

Marusyk:
@VipulM-MSFT Thanks for answer
So, if you are currently fixing all then will be fixed this as well? Are you sure that I can make PR?

Marusyk:
Can be closed. 
Thanks all

PR #86, opened 2018-03-07
Using interface IMessageBody instead of OutgoingMessageBody/IncomingMessageBody

Marusyk:
Please review. Made fix according to #82 and #83 
Thank you in advance


VipulM-MSFT:
Thank you making the change. @suchiagicha and @amanbha - do you have any feedback?

suchiagicha:
@Marusyk 

We are sending buffers(OutgoingMessageBody) and receiving stream(IncomingMessageBody). In order to make this explicit in serialization interface, we are thinking different way to fix this issue. 

Here are the recommended steps:

1. Remove the IncomingMessageBody and OutgoingMessageBody dependency on IMessageBody .
2. Make IncomingMessageBody  implement IDisposable .
3 Make OutgoingMessageBody  implement IDisposable .
4. Remove IMessageBody  interface.
5. Remove GetReceivedBuffer method from OutgoingMessageBody .
6. Remove GetSendBuffers method from IncomingMessageBody .

Let us know if you are interested in making these changes.

Marusyk:
Hello @suchiagicha,
That's a good point. Thanks. I've committed that changes. Please review

suchiagicha:
@Marusyk 
Thinking more about it. We found one more way to fix it :) 
Using concrete classes in interface doesn't look promising.
By Splitting IMessageBody into two interface IIncomingMessageBody(having GetRecievedBuffer)  and IOutgoingMessageBody(contain GetSendBuffer) . We don't need IMessageBody then.
and have concrete classes implementing those respectively instead of IMessageBody.
Serialization interface now depends on message interfaces instead of concrete classes.
That way interfaces look more pluggable .

Let me know your thoughts .

Marusyk:
@suchiagicha
Yeah, maybe that new interfaces makes sense for future unit tests. But it is a new abstractions and it seems that YAGNI applies. Could there be a case when we need to have a more than once implementations of OutgoingMessageBody/IncomingMessageBody? Interfaces are a tool for defining contracts between multiple subsystems of application, but we are using it only inside Remoting.  

I can think of two reasons for having an interface when there is only one implementation:
 - There is another mock implementation for unit tests.
 - The interface and the implementation are defined in different class libraries.

So, if it is for unit tests then Ok for me.  If you make decision to create a two new interfaces, I'll prepare a new changes, just let me know.

suchiagicha:
In future if we want to add implementation for IIncomingMessageBpdy and IOutgoingMessageBody for ex: another transport. it will be easier to do it if its not a concrete class.

Sure. Please go ahead with the change

Marusyk:
@suchiagicha 
Done. Please review

suchiagicha:
@Marusyk  Looks good to me. Thanks for the contribution. Will merge it soon .

amanbha:
Thanks @Marusyk  and @suchiagicha . This will be part of our 6.3 runtime release.

Issue #78, opened 2018-02-21
ActorService.RemotingListener never set


jfloodnet:
Hi,

I've just been testing an upgrade to V2 and was inspecting the property RemotingListener on the ActorService of the Actor.

This property kept telling me that the Listener was still on V1Listener, even though I was dead sure I'd configured it correctly.

Turns out the property is never used, and therefore misleading.

https://github.com/Azure/service-fabric-services-and-actors-dotnet/blob/develop/src/Microsoft.ServiceFabric.Actors/Runtime/ActorService.cs#L89

Cheers,
Joe

amanbha:
@jfloodnet Thanks for reporting it, indeed its not set. we will fix it in our next release.

amanbha:
@jfloodnet I discussed it with my teammates and it seems like don't intend to expose this property, could you please explain the scenario for which you want to use it?

amanbha:
I am closing this one as we don't intend to expose this property RemotingListener and will remove it in future releases. Please feel free to reopen if you have other concerns.

jfloodnet:
Sorry for the no reply. My usecase was for debugging, I was upgrading to V2 remoting, and was trying to confirm that the ActorService was indeed running on V2, and that property looked like it would do the trick.

PR #95, opened 2018-03-21
ActorService.RemotingListener never set #78

amanbha:
Setting ActorServiceRemotingListener


amanbha:
@suchiagicha Could you please review it.

Issue #81, opened 2018-02-27
The enumeration name for the remoting protocol versions should be named correctly


VipulM-MSFT:
Currently the name of the enumeration that selects the remoting version are "RemotingClient" and "RemotingListener". This is confusing as these types are not representing the actual client or listener, but they are version selector. They should be appropriately named.

Marusyk:
@VipulM-MSFT 
What do you think about this rename

```
public enum RemotingClientVersion
{
	V1 = 0,
	V2 = 1,
}
```

and

```
public enum RemotingListenerVersion
{
	V1 = 0,
	Compat = 1,
	V2 = 2,
}
```

Marusyk:
Hi @VipulM-MSFT, any news?

suchiagicha:
@Marusyk 

Looks good to me. Would you be interested in making the change?

Marusyk:
@suchiagicha Of course, I'll prepare PR. Thanks

suchiagicha:
We have made this fix in 3.2 release . Closing this issue.

Marusyk:
@suchiagicha OK, but can you explain why did you made this changes without my PR? I've tried to help to project that I'm using. But you just threw away my changes and apply yours that is the same.

PR #102, opened 2018-03-30
Change enumeration name for the remoting protocol versions 

Marusyk:
#81 
Please review.


Marusyk:
@amanbha But why you just made this change without my pull request? Why I created thid PR? Does opensource works in this way?

amanbha:
@Marusyk I understand your point of view. Actually this change comes with other work which the team did for fixing the V2 remoting stack, since this name change was small so we included it with V2 remoting fixes.

amanbha:
@Marusyk It was a mistake a on our behalf to not merge this PR first before making further changes on it. I want to assure you that we value your contributions and this will  not happen again. Please let me know on how this can be resolved so that credit where its due for this work is not lost.

PR #108, opened 2018-04-27
WebServiceHost support

wasabii:
Added a `WcfWebCommunicationListener` which uses a `WebServiceHost` instead of a `ServiceHost`.


wasabii:
This is mostly a copy of the existing `WcfCommunicationListener`, but only slightly altered to use a `WebServiceHost` instead of a `ServiceHost`.

I think it would be better if there was a `WcfCommunicationListenerBase` class, which used a generic `TServiceHost` type to set that. But I'm not going to move on that without the responsible party weighing in.

suchiagicha:
@wasabii 
Today servicehost setter is private. You could make the func CreateServiceHost virtual protected  .
Now you can inherit the existing class and then override the CreateServiceHost function.

I hope this helps you.

amanbha:
@wasabii Just checking if you have updates on the PR based on comments by @suchiagicha

amanbha:
@wasabii , closing this PR because of inactivity, please feel free to reopen with updates based on the comments by @suchiagicha

wasabii:
@suchiagicha I have implemented your suggestions. The `CreateServiceHost` and `ConfigureServiceHost` methods in the `WcfCommunicationListener` class have been made `protected virtual`, and are no longer `static`. These are then overridden by the `WcfWebCommunicationListener` class.

PR #117, opened 2018-06-04
Making Non Transient Exception Retry Count Configurable

suchiagicha:


src/Microsoft.ServiceFabric.Services.Wcf/Communication/Wcf/Client/WcfExceptionHandler.cs
```diff
@@ -177,7 +177,7 @@ public WcfExceptionHandler()
                     e,
                     false,
                     retrySettings,
-                    retrySettings.DefaultMaxRetryCount);
+                    retrySettings.DefaultMaxRetryCountForNonTransientErrors);
```
BharatNarasimman: >DefaultMaxRetryCountForNonTransientErrors [](start = 34, length = 41)

Transient error

suchiagicha: This is non-transient error , hence using non transient retry count.

PR #123, opened 2018-06-08
Doing Lazy Initialization for Random

suchiagicha:


src/Microsoft.ServiceFabric.Services/Communication/Client/CommunicationClientFactoryBase.cs
```diff
@@ -661,6 +661,11 @@ private int NextRandom(int upperBound)
             var rand = 0;
             lock (this.randomLock)
             {
+                if (this.random == null)
+                {
+                    this.random = new Random(this.GenerateSeed());
+                }
```
BharatNarasimman: Can you please add comments on why the default seed is not sufficient?

Issue #129, opened 2018-06-29
Please add CreateNonIServiceProxy to the IServiceProxyFactory interface


Expecho:
The public interface `IServiceProxyFactory` [here](https://github.com/Microsoft/service-fabric-services-and-actors-dotnet/blob/0e8d0ed6ebd3fa34248fbb050780468f1ce9d744/src/Microsoft.ServiceFabric.Services.Remoting/Client/IServiceProxyFactory.cs ) does not contain the method `CreateNonIServiceProxy` making it hard to use the interface for unit testing purposes if you depend on the `CreateNonIServiceProxy` method.

suchiagicha:
Sure. Would you be interested in submitting PR?

Expecho:
Sure, no problem. I'll get to it tomorrow.

Expecho:
I think we are running into problems. The V1 remoting stack does not support the `CreateNonIServiceProxy` method. Most interfaces and classes have a separate V1 and V2 version, except for the single `IServiceProxyFactory` interface that is shared by the internal classes `Microsoft.ServiceFabric.Services.Remoting.V2.Client.ServiceProxyFactory`  and `Microsoft.ServiceFabric.Services.Remoting.V1.Client.ServiceProxyFactory` and the public class Microsoft.ServiceFabric.Services.Remoting.Client.ServiceProxyFactory.

Now, since `Microsoft.ServiceFabric.Services.Remoting.V1.Client.ServiceProxyFactory` is internal we could opt for just create an implementation of `IServiceProxyFactory.CreateNonIServiceProxy` that throws a NotImplementedException for example since it won't be accessed anywhere and especially not from outside the assembly.

Any thoughts on this?

suchiagicha:
Sounds good  :)

PR #131, opened 2018-07-04
Extend IServiceProxyFactory with CreateNonIServiceProxy method

Expecho:
Fix for #129

### DIFF
--- a/src/Microsoft.ServiceFabric.Services.Remoting/V1/Client/ServiceProxyFactory.cs
+++ b/src/Microsoft.ServiceFabric.Services.Remoting/V1/Client/ServiceProxyFactory.cs
@@ -45,6 +45,11 @@ namespace Microsoft.ServiceFabric.Services.Remoting.V1.Client
             this.retrySettings = retrySettings;
         }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.ServiceFabric.Services.Remoting/V1/Client/ServiceProxyFactory.cs
```diff
@@ -45,6 +45,11 @@ internal class ServiceProxyFactory : IServiceProxyFactory
             this.retrySettings = retrySettings;
         }
 
+        public TServiceInterface CreateNonIServiceProxy<TServiceInterface>(Uri serviceUri, ServicePartitionKey partitionKey = null, TargetReplicaSelector targetReplicaSelector = TargetReplicaSelector.Default, string listenerName = null)
+        {
+            throw new NotImplementedException("CreateNonIServiceProxy is not supported in remoting V1");
```
suchiagicha: Please change the exception type to NotSupportedException and
add the string in the Resource String list. Here is an example on it :  [](url)

Expecho: Sure, no problem.

PR #168, opened 2018-12-31
RegisterReminderAsync DueTime validation

arturek:
The RegisterReminderAsync method should accept "negative one (-1) milliseconds" as a valid value of its dueTime parameter. This is both specified in the the [documentation](https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicefabric.actors.runtime.actorbase.registerreminderasync?view=azure-dotnet) and in the message of an exception thrown when this validation fails (TimeSpan TotalMilliseconds specified value must be between -1 and 922337203685477, Parameter name: DueTime). Other code which uses the DueTime property seems to be aware of -1 value and handles it correctly.


amanbha:
We should update the documentation for this. If the value of duetime is -1, reminder will not fire even once and its timer will not be armed for the next firing as specified by perioed.

PR #175, opened 2019-02-25
Setting the Client in Cache after client open call succeeds

suchiagicha:


suchiagicha:
@amanbha 
Sure . Will do it.

Issue #177, opened 2019-03-08
ClientRetryTimeout not honored with CancellationToken


mattrowmsft:
**Describe the bug**
The ClientRetryTimeout is ignored if a real CancellationToken is used.  It defeats the purpose of the config to not use it in most cases.

**To Reproduce**
Pass in a real CancellationToken and talk to an endpoint with retryable failure.

**Expected behavior**
Should timeout when ClientRetryTimeout or cancellationToken expires, whichever comes sooner.

**Additional context**
I'll send PR with proposed solution.

PR #178, opened 2019-03-08
Change retry logic to cancel after ClientRetryTimeout or cancellation…

mattrowmsft:
…Token.IsCancellationRequested, whichever comes sooner. Prevents cases where ClientRetryTimeout is not honored. #177


mattrowmsft:
Actually hold off on merging this for now. I think I found a more subtle bug, so at a minimum I'll add some unit tests to the ServicePartitionClient class.

mattrowmsft:
I'm not sure why this closed based on my fork being updated.

PR #181, opened 2019-03-18
Change retry logic to cancel after ClientRetryTimeout or cancellation…

mattrowmsft:
…Token.IsCancellationRequested, whichever comes sooner. Prevents cases where ClientRetryTimeout is not honored. #177

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,7 @@ bld/
 *.vsp
 *.vspx
 /drop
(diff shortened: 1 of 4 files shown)


mattrowmsft:
Sorry, my local branch was setup incorrectly previously and caused me to end up pushing commit. Reverted commit and opened PR again.

PR #184, opened 2019-03-27
ServicePartitionClient fix to return on cancellation token canceled i…

mattrowmsft:
…nstead of invoking user exception handler.


mattrowmsft:
This will fix unit test failures as well.  The exception handler was incorrectly being invoked in these cases, which was causing test failure.

PR #186, opened 2019-04-05
Tracing health for Communication listener close & MissingMethodException when method is not found

amanbha:
Summary of Changes:
Adding traces and health reporting around Communication listeners.
Returning MissingMethodException when method is not found on the service side.


amanbha:
Fix for issues #188 & #189

Issue #67, opened 2018-01-17
No transactions in WCF service hosted by WcfCommunicationListener


dronkoff:
I have a contract and implementation
```CSharp
    [ServiceContract(Namespace = "http://TestInterfaces.org")]
    public interface ICalculator
    {
        [OperationContract]
        [TransactionFlow(TransactionFlowOption.Allowed)]
        string GetTransactionInfo();
    }
//..........
    public class CalcService : ICalculator
    {
        [OperationBehavior(TransactionScopeRequired = true, TransactionAutoComplete = true)]
        public string GetTransactionInfo()
        {
            Console.WriteLine("Received GetTransactionInfo");
            var res = "No transaction context found";
            if(Transaction.Current != null)
            {
                res = $"LocalID: {Transaction.Current.TransactionInformation.LocalIdentifier}\n";
                res += $"DistributedID: {Transaction.Current.TransactionInformation.DistributedIdentifier}\n";
                res += $"IsolationLevel: {Transaction.Current.IsolationLevel}\n";
                res += $"CreationTime: {Transaction.Current.TransactionInformation.CreationTime}, Status: {Transaction.Current.TransactionInformation.Status}";
            }
            Console.WriteLine("Return: {0}", res);
            return res;
        }
    }

```

They are packaged as stateless service

```CSharp
    internal sealed class CalcSvc : StatelessService
    {
        public CalcSvc(StatelessServiceContext context)
            : base(context)
        { }

        protected override IEnumerable<ServiceInstanceListener> CreateServiceInstanceListeners()
        {
            const int bufferSize = 512000; //500KB
            NetTcpBinding binding = new NetTcpBinding(SecurityMode.None)
            {
                SendTimeout = TimeSpan.FromSeconds(30),
                ReceiveTimeout = TimeSpan.FromSeconds(30),
                CloseTimeout = TimeSpan.FromSeconds(30),
                MaxConnections = 1000,
                MaxReceivedMessageSize = bufferSize,
                MaxBufferSize = bufferSize,
                MaxBufferPoolSize = bufferSize * Environment.ProcessorCount,
                TransactionFlow = true
            };

            ServiceInstanceListener listener = new ServiceInstanceListener(context =>
                {
                    var wcfListener = new WcfCommunicationListener<TestInterfaces.ICalculator>(
                        context, new TestImpl.CalcService(), binding, "CalcServiceEndpoint");

[...]

PR #199, opened 2019-05-09
Resolve issue #198 - WCF Host fails to apply transaction configuration of WCF configuration for service

### LINKED ISSUE #198: ServiceHost for WCF service hosted by WcfCommunicationListener fails to enable Transaction support
**Describe the bug**
This identifies the root cause for closed (no fix) issue #67.
When `WcfCommunicationListener` is invoked to internally delegate to the WCF ServiceHost, it fails to include the Service implementation instance / Service Type when building the `ServiceEndpoint`.

The consequence of this is that the WCF ServiceHost is unaware of the specific Service Implementation (that contains essential attributes for example transaction configuration for the service defined by applying `[ServiceBehavior(...)]` attribute on Service Implementation classes) and it only uses the Contract Type (typically an interface which only allows `[ServiceOperation(...)]` attributes and doesn't allow `[ServiceBehavior()]` attributes). 

When the ServiceHost then stands up the service, it doesn't setup the requested transaction configuration (and similarly so for other service implementation configuration attributes that may have been defined).

The [`WcfCommunicationListener.CreateServiceEndpoint()`](https://github.com/microsoft/service-fabric-services-and-actors-dotnet/blob/41a2e7fdb8483420bf1b4dc6ad8af877de897410/src/Microsoft.ServiceFabric.Services.Wcf/Communication/Wcf/Runtime/WcfCommunicationListener.cs#L306) method has the relevant code that is problematic:

```
        private static ServiceEndpoint CreateServiceEndpoint(Type contractType, Binding binding, EndpointAddress address)
        {
            var endpoint = new ServiceEndpoint(
                ContractDescription.GetContract(contractType),
                binding,
                address);
            endpoint.Behaviors.Add(new ListenUriEndpointBehavior());

            return endpoint;
        }
```
The `ContractDescription.GetContract()` method has an overload available that allows for providing the instance of the Service implementation (or the type). This needs to be populated accordingly (depending on which constructor of the `WcfCommunicationListener` is used of course).

**To Reproduce**
Refer to issue #67 from user @dronkoff. The same implementation demonstrates a typical WCF Listener implementation with a WCF Service with transactional configuration.

**Expected behavior**
If `[ServiceBehavior(...)]` attributes are applied to a service type (not the contract interface type), then it should be honoured with Service Fabric hosts the WCF service. 

To achieve this, the `WcfCommunicationListener` constructors that accept Service Implementation Types or Instances (in addition to the Contract Interface Type generic parameter) should be passed down the call chain to the `CreateServiceEndpoint(...)` static method and passed into the `ContractDescription.GetContract(contractType, [serviceInstance|serviceType])` call.

**Additional context**
Please advise expected timeframe / timeline for this change. This limitation is deal breaker for us as we cannot string together multiple WCF calls into a single UoW transaction.

Jaans:
PR to resolve #198 & #67 

Code change passes service implementation instance (or type depending on constructor used) down the call chain where it is used to create service endpoint instances which in turn is used by service host to configure and activate WCF service.

Local test confirms fix with WCF Service Host within Service Fabric applies configuration declared by `[ServiceBehavior(...)]` attribute, like implicit transaction scope.

### DIFF
--- a/src/Microsoft.ServiceFabric.Services.Wcf/Communication/Wcf/Runtime/WcfAsyncThreadExceptionHandler.cs
+++ b/src/Microsoft.ServiceFabric.Services.Wcf/Communication/Wcf/Runtime/WcfAsyncThreadExceptionHandler.cs
@@ -17,7 +17,7 @@ namespace Microsoft.ServiceFabric.Services.Communication.Wcf.Runtime
         {
             ServiceTrace.Source.WriteWarning(
                 this.traceType,
-                "Excepion {0}  Stacktrace {1} occured in Wcf Service Background Threads",
+                "Exception {0}  Stacktrace {1} occured in Wcf Service Background Threads",
                 exception.Message,
                 exception.StackTrace);
 
--- a/src/Microsoft.ServiceFabric.Services.Wcf/Communication/Wcf/Runtime/WcfCommunicationListener.cs
+++ b/src/Microsoft.ServiceFabric.Services.Wcf/Communication/Wcf/Runtime/WcfCommunicationListener.cs
@@ -196,7 +196,7 @@ namespace Microsoft.ServiceFabric.Services.Communication.Wcf.Runtime
                 this.publishAddress = serviceContext.PublishAddress;
             }
 
-            this.endpoint = CreateServiceEndpoint(typeof(TServiceContract), listenerBinding, address);
+            this.endpoint = CreateServiceEndpoint(typeof(TServiceContract), wcfServiceObject, listenerBinding, address);
             this.host = CreateServiceHost(wcfServiceObject, this.endpoint);
         }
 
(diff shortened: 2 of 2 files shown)


Jaans:
Hi @suchiagicha 
I was expecting the latest NuGet packages for Microsoft.ServiceFabric.Services.Wcf v3.4.641 would include the above change, but it seems that it does not. Am I mistaken?
Which version would you expect will include this fix?

Jaans:
@suchiagicha & @amanbha Could you please give us an update on this?

amanbha:
@Jaans  I just noticed that this was checked into develop branch, which would put it in our next release(it will be 3.5.*)
To fasttrack it, I have cherry picked this into release_3.4 branch, so that it can be in our next  CU release 3.4.*

Jaans:
Excellent! That is great news. Thank you so much!

Just for my understanding, should I have created the PR for a different branch?

amanbha:
@Jaans You created the PR in the correct branch to target next major/minor update, its just that the release on our end got delayed.

Jaans:
@amanbha I'm guessing that it didn't make it to the next 3.4 CU release v3.4.676 ?

amanbha:
Not this one, the release was already going through the pipeline when I merged this change.

amanbha:
Please pick nuget pkg version 3.4.677

PR #203, opened 2019-05-24
Update EventSource classes to check whether the EventLevel is enabled…

mattrowmsft:
… before paying the cost to format the event. This performance optimization is most dramatic with noise level traces.

### DIFF
--- a/src/Microsoft.ServiceFabric.Actors/ActorEventSource.cs
+++ b/src/Microsoft.ServiceFabric.Actors/ActorEventSource.cs
@@ -38,13 +38,18 @@ namespace Microsoft.ServiceFabric.Actors
         [NonEvent]
         internal void WriteErrorWithId(string type, string id, string format, params object[] args)
         {
+            if (!this.IsEnabled(EventLevel.Error, Keywords.Default))
+            {
+                return;
+            }
+
             if (args == null || args.Length == 0)
             {
-                Instance.ErrorText(id, type, format);
+                this.ErrorText(id, type, format);
             }
             else
             {
-                Instance.ErrorText(id, type, string.Format(CultureInfo.InvariantCulture, format, args));
+                this.ErrorText(id, type, string.Format(CultureInfo.InvariantCulture, format, args));
             }
         }
 
@@ -57,13 +62,18 @@ namespace Microsoft.ServiceFabric.Actors
         [NonEvent]
         internal void WriteWarningWithId(string type, string id, string format, params object[] args)
         {
+            if (!this.IsEnabled(EventLevel.Warning, Keywords.Default))
+            {
+                return;
+            }
+
             if (args == null || args.Length == 0)
             {
-                Instance.WarningText(id, type, format);
+                this.WarningText(id, type, format);
             }
             else
             {
-                Instance.WarningText(id, type, string.Format(CultureInfo.InvariantCulture, format, args));
+                this.WarningText(id, type, string.Format(CultureInfo.InvariantCulture, format, args));
             }
         }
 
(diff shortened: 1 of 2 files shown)


amanbha:
@mattrowmsft Could you please validate that with this change events are still traced, if yes, we can merge it.

mattrowmsft:
Sorry I never got around to checking this.  Let me try to close on this by EOW.

mattrowmsft:
Actually this isn't the right way to solve this.  The filtering is meant to be controlled via the Config package for the service. So while this works, I think it will break expectations on how things should work.  I think I can instead change our apps to override this configuration value to reduce amount traced.

mattrowmsft:
Adding following to Settings.xml for SF service would set those traces to Error/Warning only. (Warning = 3)
<Settings>
  <Section Name="Tracing/ServiceFramework">
    <Parameter Name="Level" Value="3" />
  </Section>
</Settings>

juho-hanhimaki:
> Adding following to Settings.xml for SF service would set those traces to Error/Warning only. (Warning = 3)
> 
> ```
> <Settings>
>   <Section Name="Tracing/ServiceFramework">
>     <Parameter Name="Level" Value="3" />
>   </Section>
> </Settings>
> ```

Is this documented somewhere or is it still in development? I am interested in filtering out Information level events for performance gains.

mattrowmsft:
Glad we are open source now so I can just link to the code :-).

It is supported and working today.  I'm guessing it has been there for several releases too.  It isn't well documented but you can find all the code at following paths.

// Anything that derives from this base EventSource implementation would support this.
https://github.com/microsoft/service-fabric/blob/master/src/prod/src/managed/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs

// Config options set here.
https://github.com/microsoft/service-fabric/blob/master/src/prod/src/managed/Microsoft.ServiceFabric.Diagnostics/Tracing/Config/TraceConfig.cs

juho-hanhimaki:
Great stuff. I have wondered where that bit of code is. But it is in completely different git repo :)

Regarding this PR: Purely from perf standpoint wouldn't it still make sense to filter events before formatting (in the ActorEventSource.cs and ServiceEventSource.cs)?

So without this PR the events are still formatted but then later ignored based on the configured level in the ServiceFabricEventSource.cs? Or maybe I haven't yet understood the whole picture?

amanbha:
@juho-hanhimaki I just want to add that if you reduce the level of tracing by the service, we will not be able top provide support for the remoting calls made through the service code, users will have to rely on tracing from the service code to do any analysis when the calls fail.

juho-hanhimaki:
@amanbha That is understandable. I don't get why it is something we should be super concerned about? We don't get support for using HttpClient or other communication method we might use between services either. Are there some major risks or is remoting fragile in some way that I don't understand?

We have operated clusters for few years now and haven't really encountered any notable problems specific to remoting. That is also why we choose to use it. It has proven to be easy to use, stable and effective for the purpose of communication between services.

In our application we have analyzed that there are some types of frequent (hot path) remoting calls where there is no value (that we can think of) for us to log each or even any one of them. Of course there are other types of calls we might want to log. That is where the current remoting stack doesn't deliver because there is no option to really configure the tracing per method or otherwise.

In the end isn't it totally up to the design, complexity and requirements of the customer application to decide the appropriate approach to tracing? The used platform/tools (SF + remoting) can have some role in that decision but I don't think it should be the main factor.

amanbha:
@juho-hanhimaki Since Remoting is shipped as a component for writing SF services by our team,, there are some customers whose services lack proper tracing and diagnostics and file support tickets for SF Remoting related issues and our team has to provide support for those issues. So it varies from customer to customer. If you are ok with this trade off and can handle any issues by your tracing and diagnostics, then its fine.
I mentioned it explicitly for benefit of others, so if other users stumble upon this thread they are well aware of it.

juho-hanhimaki:
Yep that makes a lot of sense.

Right now it's always a somewhat of a tradeoff whatever way one decides to go since the only way to control tracing is at the service level. If there were more granular control people could make better decisions for their specific application. Service Fabric (cluster and programming models) have quite steep learning curve and are definitely tough on beginners. But I would expect the majority of the users to be able to make smart decisions if provided the right tools and information.

mattrowmsft:
Hey, sorry I missed explaining rest of change.

Ideally the change would still make sense as you mentioned. You avoid the formatting cost.  The reason why this won't help in practice is because SF listens to these events no matter what.  We would need to add support for configuring the trace session to allow this optimization to work.  You can see the trace session using 'logman FabricTraces -ets'.

amanbha:
@juho-hanhimaki Yes, I agree it could have been more granular. The change suggested by @mattrowmsft above will only disable the tracing done by Services and remoting layer which use this [event source](https://github.com/microsoft/service-fabric-services-and-actors-dotnet/blob/develop/src/Microsoft.ServiceFabric.Services/Runtime/ServiceFrameworkEventSource.cs). All the tracing done by runtime for this service will still be available.

Issue #234, opened 2019-12-02
Multiple IService implementations causing 'System.DivideByZeroException' in ServiceRemotingMessageDispatcher


mbirkes:
When two or more IService implementations share the same interface name from different namespaces, and have at least one Operation with the same name, a System.DivideByZeroException occurs during the listener startup for any StatelessService that implements both IService interfaces.  

I originally posted this here, but realize it is actually a services SDK issue: https://github.com/Azure/service-fabric-issues/issues/1610

## Expected Behavior
Allow a single stateless service instance to implement multiple IService implementations. 

## Current Behavior
ServiceRemotingMessageDispatcher initializer throws exception: System.DivideByZeroException: 'Attempted to divide by zero.'  See stack trace in following section. 

## Steps to Reproduce
1. Define two IService interfaces with the same name in different namespaces
2. At least one method in each Interface should have the same operation Name (parameters can be different)
3. Create a StatelessService implementing the two interfaces
4. Setup a FabricTransportServiceRemotingListener for the service. 
5. During FabricTransportServiceRemotingListener constructor, the Exception will be thrown. 

The following sample code can reproduce the behavior. Removing/Renaming one of the 'OperationA' methods will allow the service to start successfully. 

```csharp
namespace MyServiceA
{
    public interface IMyInterface : IService
    {
        Task<string> OperationA(string request);
        Task OperationC(int request);
    }
}

namespace MyServiceB
{
    public interface IMyInterface : IService
    {
        Task<int> OperationA(int request);
        Task OperationD(int request);
    }
}

namespace MyService
{
   
    internal sealed class MyService : StatelessService, MyServiceA.IMyInterface, MyServiceB.IMyInterface
    {
        public MyService(StatelessServiceContext context) : base(context)
        {
        }

        protected override IEnumerable<ServiceInstanceListener> CreateServiceInstanceListeners()
        {
            return new[]
            {
                new ServiceInstanceListener(
                    context => new FabricTransportServiceRemotingListener(context, this), "MyServiceListener"),
            };
        }

        Task<string> MyServiceA.IMyInterface.OperationA(string request)
        {
            return Task.FromResult("test");
        }

        Task MyServiceA.IMyInterface.OperationC(int request)
        {
            return Task.CompletedTask;
        }

        Task<int> MyServiceB.IMyInterface.OperationA(int request)
        {
            return Task.FromResult(1);
        }

[...]

PR #235, opened 2019-12-02
Fix System.DivideByZeroException in PerformanceCounterInstanceNameBuilder.ComputeTruncatedMethodName for overloaded service methods

mbirkes:
See issue: https://github.com/microsoft/service-fabric-services-and-actors-dotnet/issues/234
Multiple IService implementations causing 'System.DivideByZeroException' in ServiceRemotingMessageDispatcher


amanbha:
@BharatNarasimman Could you please help review this change?

PR #254, opened 2020-09-10
dispose actor reminder in case of update

oshvartz:
I encounter issue in our project we are updating existing reminder by calling RegisterReminderAsync and it works well. The issue we notice is memory usage increase after investigating we saw that the RC was ActorReminders waiting in the finilizer queue.
Reviewing the code I saw that on update reminder we create new reminder and just call CancelTimer on the "old" one. I suggest that on update we can just Dispose the "old" reminder which will suppress finalizers.

### DIFF
--- a/src/Microsoft.ServiceFabric.Actors/Runtime/ActorManager.cs
+++ b/src/Microsoft.ServiceFabric.Actors/Runtime/ActorManager.cs
@@ -1254,7 +1254,7 @@ namespace Microsoft.ServiceFabric.Actors.Runtime
                 actorReminder,
                 (k, v) =>
                 {
-                    v.CancelTimer();
+                    v.Dispose();
                     return actorReminder;
                 });
 


src/Microsoft.ServiceFabric.Actors/Runtime/ActorManager.cs
```diff
@@ -1254,7 +1254,7 @@ private async Task LoadRemindersAsync(CancellationToken cancellationToken)
                 actorReminder,
                 (k, v) =>
                 {
-                    v.CancelTimer();
```
amanbha, line 1257: CancelTimer call does Dispose internally, could you please explain a bit more how this Dispose here helps?

oshvartz: the logic is the same the only difference is that on Dispose we call - GC.SuppressFinalize
and we have extreme case - of ~20k actors we reminders that are triggered once in ~5 minutes and each time inside  ReceiveReminderAsync we are - calling RegisterReminderAsync - this causes some stress on the finalization thread handling spikes of ActorReminders finilizers

amanbha (approved):

PavanKunapareddyMSFT (approved):

PR #259, opened 2020-12-08
Safe some memory and also string comparisions on Acquire and Call Context

KDSBest:
Safe some memory and also string comparisions on Acquire and Call Context


olegsych:
Abandoned

PR #264, opened 2021-02-15
Xsd support for startup services - Added clear statement in config

sunil-indoria:
Xsd support for startup services - Added clear statement in nuget.config to avoid security analysis failure.


sunil-indoria:
Merged another PR with only config changes. This is reflecting already merged changes. Hence this PR is no longer required.

PR #269, opened 2021-03-16
Fix of ChangeRole being stuck in RunAsync

zilongj:
The PR consists of the fix for the ChangeRole issue posted at [https://github.com/microsoft/service-fabric/issues/1058](https://github.com/microsoft/service-fabric/issues/1058)

The ChangeRole issue happens when is Primary replica is becoming None (i.e., being closed). In this case, the ChangeRole procedure can get stuck, if the ChangeRole procedure is triggered when the RunAsync() is committing a transaction. In this case, the Native stack can successfully complete ChangeRole. However, the managed stack will not complete ChangeRole until RunAsync is canceled. As RunAsync detects that the role of native stack is not Primary, the RunAsync can never complete its commit and will not be canceled successfully. This creates a deadlock situation.

We reproduced the ChangeRole issue on a local machine. Based on multiple test runs, the code changes made in the PR can effectively fix the above issue.

### DIFF
--- a/src/Microsoft.ServiceFabric.Services/Runtime/StatefulServiceReplicaAdapter.cs
+++ b/src/Microsoft.ServiceFabric.Services/Runtime/StatefulServiceReplicaAdapter.cs
@@ -132,6 +132,7 @@ namespace Microsoft.ServiceFabric.Services.Runtime
                 newRole);
 
             await this.CloseCommunicationListenersAsync(cancellationToken);
+            await this.stateProviderReplica.ChangeRoleAsync(newRole, cancellationToken);
 
             if (newRole == ReplicaRole.Primary)
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.ServiceFabric.Services/Runtime/StatefulServiceReplicaAdapter.cs
```diff
@@ -223,6 +221,8 @@ async Task IStatefulServiceReplica.CloseAsync(CancellationToken cancellationToke
 
                 this.stateProviderReplica = null;
             }
+
+            await this.CancelRunAsync();
         }
 
         void IStatefulServiceReplica.Abort()
```
zuhairp, line 228: Add change to Abort path as well

PR #276, opened 2021-05-12
StartupServices Cleanup from Actor Repo

sunil-indoria:


sunil-indoria:
Updated Microsoft.ServiceFabric version should be added.

anantshankar17 (approved):

sunil-indoria:
it should use latest Microsoft.ServiceFabric nuget verison once winfab repo publishes that.

PavanKunapareddyMSFT (approved):

PR #289, opened 2021-11-12
Check if runtimeconfig.dev.json exists

pregress:
.NET 6 doesn't create runtimeconfig.dev.json by default  See:  https://github.com/dotnet/sdk/issues/16818
This crashes the build of Service fabric projects that use .NET 6 services/actors.
I've created the issues at : https://github.com/microsoft/service-fabric/issues/1289


pregress:
@PavanKunapareddyMSFT can you review? This blocks migrating to .NET 6

PR #293, opened 2021-12-09
Updating Migration Worker Framework 

jyotimisramsft:
Adding consuming response stream from kvs web API.
Also adding changes requested in [previous PR ](https://github.com/jyotimisramsft/service-fabric-services-and-actors-dotnet/pull/1)


src/Microsoft.ServiceFabric.Actors.Migration/MigrationWorker.cs
```diff
@@ -39,44 +39,102 @@ public MigrationWorker(KVStoRCMigrationActorStateProvider stateProvider, ActorTy
...
+                        while (responseLine != null)
+                        {
+                            List<KeyValuePair> kvsData = new List<KeyValuePair>();
+                            using (Stream memoryStream = new MemoryStream())
+                            {
+                                byte[] data = Encoding.UTF8.GetBytes(responseLine);
+                                memoryStream.Write(data, 0, data.Length);
+                                memoryStream.Position = 0;
+
+                                using (var reader = XmlDictionaryReader.CreateTextReader(memoryStream, XmlDictionaryReaderQuotas.Max))
```
PavanKunapareddyMSFT, line 100: You need to make changes in KVS controller as well to serialize to string rather than bytes?

src/Microsoft.ServiceFabric.Actors.Migration/MigrationOrchestrator.cs
```diff
@@ -52,168 +52,189 @@ public async Task StartMigration(CancellationToken cancellationToken)
...
-                    using (var tx = this.stateProvider.GetStateManager().CreateTransaction())
-                    {
-                        for (int i = 0; i < workerCount; i++)
-                        {
-                            var workerStatusValue = await this.metadataDict.TryGetValueAsync(tx, MigrationConstants.GetCopyWorkerStatusKey(i));
-                            MigrationStatus.TryParse(Encoding.ASCII.GetString(workerStatusValue.Value), out MigrationStatus status);
-                            isWorkerTaskIncomplete[i] = status != MigrationStatus.Completed;
-                        }
-                    }
+            int.TryParse(Encoding.ASCII.GetString(copyPhaseWorkerCountValue.Value), out var workerCountForCopyPhase);
```
PavanKunapareddyMSFT, line 91: What happens if TryParse fails ? If we are using TryParse/TryGetValueAsync, we need to make sure we are handling TryParse returning false and TryGetValueAsync returning ConditionalValue with null value. 
Or you can simply call Parse and let the service fail with framework specific error.

PavanKunapareddyMSFT (changes requested):

src/Microsoft.ServiceFabric.Actors.Migration/MigrationWorker.cs
jyotimisramsft: Updated the KVS API

PR #297, opened 2022-01-03
Handling ambiguous actor ids

roshanroutray:


src/Microsoft.ServiceFabric.Actors.Migration/MigrationOrchestrator.cs
```diff
@@ -217,6 +227,67 @@ public async Task StartMigration(CancellationToken cancellationToken)
...
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            var startSN = long.Parse(await Client.GetStringAsync(this.userSettings.KvsEndpoint + MigrationConstants.GetStartSNEndpoint));
+            var endSN = long.Parse(await Client.GetStringAsync(this.userSettings.KvsEndpoint + MigrationConstants.GetEndSNEndpoint));
+
+            if (endSN - startSN < this.userSettings.ItemsPerEnumeration || this.userSettings.ItemsPerEnumeration == -1)
+            {
+                this.ambiguousActorIdHandler.AddRange(await this.GetStringActorIdsFromKvsAsync(startSN, endSN - startSN, cancellationToken));
+            }
```
pakunapamsft, line 240: Instead of enumerating by sequence number, use enumeration based on continuation token. Refer to GetActorAsync or GetRemindersAsync. 
You could use chunkStartLsn and chunkEndLsn to ensure we are not overstepping by checking the version field.

src/Microsoft.ServiceFabric.Actors.Migration/AmbiguousActorIdHandler.cs
```diff
@@ -0,0 +1,260 @@
...
+            // drop last token. which could be statename or reminder name
+            if (stringActorId.Contains("_"))
+            {
+                var lastIndexOfUnderscore = stringActorId.LastIndexOf('_');
+                tempStr = stringActorId.Substring(0, lastIndexOfUnderscore);
+            }
+
+            for (int iterations = 0; iterations < stringActorId.Split('_').Count(); iterations++)
+            {
+                count += this.metadataCache.Select(n => n.Contains(tempStr)).Count();
```
pakunapamsft, line 116: We could break if the count is greater than zero.

src/Microsoft.ServiceFabric.Actors.Migration/AmbiguousActorIdHandler.cs
```diff
@@ -0,0 +1,260 @@
...
+
+                var lastIndexOfUnderscore = tempStr.LastIndexOf('_');
+                tempStr = tempStr.Substring(0, lastIndexOfUnderscore);
+            }
+
+            if (count > 1)
+            {
+                foreach (Type resolver in this.resolvers)
```
pakunapamsft, line 124: Resolver are used while fetching the actual state/reminders. Cache population is with presence keys.

src/Microsoft.ServiceFabric.Actors.Migration/AmbiguousActorIdHandler.cs
```diff
@@ -0,0 +1,260 @@
...
+                    tempStr = tempStr.Substring(0, lastIndexOfUnderscore);
+                }
+
+                if (count <= 1)
+                {
+                    return new ActorId(actorId);
+                }
+                else
+                {
+                    foreach (Type resolver in this.resolvers)
```
pakunapamsft, line 176: The cache is guaranteed to have non ambiguous actor ids, meaning only one actor id will match.

src/Microsoft.ServiceFabric.Actors.Migration/AmbiguousActorIdHandler.cs
```diff
@@ -0,0 +1,260 @@
...
+                            break;
+                        }
+                    }
+                }
+
+                return new ActorId(actorId);
+            }
+            else
+            {
+                this.ReportPartitionHealthAndThrow();
```
pakunapamsft, line 198: Throw exception. ActorService will take care of reporting partition health.

pakunapamsft (changes requested):

PR #327, opened 2022-05-12
Manual mode migration bug fix

jyotimisramsft:
Bug description:
For manual mode migration, if a failover happens during downtime phase, while resuming migration catchup iterations are resumed since the flag _isDowntimeAllowed = true_ is stored in-memory and lost during the failover

Fix:
_isDowntimeAllowed_ is stored in metadata dict and its value is checked if phase==catchup while resuming migration after a failover.

### DIFF
--- a/src/Microsoft.ServiceFabric.Actors.KVSToRCMigration/TargetMigrationOrchestrator.cs
+++ b/src/Microsoft.ServiceFabric.Actors.KVSToRCMigration/TargetMigrationOrchestrator.cs
@@ -241,7 +241,7 @@ namespace Microsoft.ServiceFabric.Actors.KVSToRCMigration
             {
                 await this.metadataDict.TryAddAsync(
                     tx,
-                    IsDowntimeAllowed,
+                    IsDowntimeInvoked,
                     true.ToString(),
                     DefaultRCTimeout,
                     cancellationToken);
@@ -419,7 +419,14 @@ namespace Microsoft.ServiceFabric.Actors.KVSToRCMigration
 
             try
             {
-                workloadRunner = await this.NextWorkloadRunnerAsync(MigrationPhase.None, cancellationToken);
+                var migrationCurrentPhase = await ParseMigrationPhaseAsync(
+                () => this.MetaDataDictionary.GetValueOrDefaultAsync(
+                    () => this.Transaction,
+                    MigrationCurrentPhase,
+                    DefaultRCTimeout,
+                    cancellationToken),
+                this.TraceId);
+                workloadRunner = await this.NextWorkloadRunnerAsync(migrationCurrentPhase, cancellationToken);
 
                 PhaseResult currentResult = null;
                 while (workloadRunner != null)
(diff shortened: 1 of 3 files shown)


src/Microsoft.ServiceFabric.Actors.KVSToRCMigration/TargetMigrationOrchestrator.cs
```diff
@@ -588,19 +599,43 @@ private async Task<IMigrationPhaseWorkload> NextWorkloadRunnerAsync(PhaseResult
         {
             var endSN = await this.GetEndSequenceNumberAsync(cancellationToken);
             var delta = endSN - currentResult.EndSeqNum;
+
             if (currentResult.Phase == MigrationPhase.Catchup)
             {
-                if (delta < this.MigrationSettings.DowntimeThreshold
-                    && this.downtimeAllowed == true)
```
roshanroutray, line 594: why are we complicating the logic here with more if..else blocks. Cant' we update downtimeAllowed from MetaDataDictionary and keep the original logic.

roshanroutray (changes requested):
why are we complicating the logic here with more if..else blocks. Cant' we update downtimeAllowed from MetaDataDictionary and keep the original logic.

src/Microsoft.ServiceFabric.Actors.KVSToRCMigration/TargetMigrationOrchestrator.cs
```diff
@@ -588,19 +599,43 @@ private async Task<IMigrationPhaseWorkload> NextWorkloadRunnerAsync(PhaseResult
...
+                if (!this.IsAutoStartMigration())
```
pakunapamsft, line 605: Simplify the logic
1) set downtimeallowed(local to method) = true when user triggered downtime api for manual mode or delta met threshold for auto mode
2) Common logic to invoke downtime phase if downtime is set, catchup otherwise

- Also remove downtimeAllowed class variable

pakunapamsft (changes requested):

src/Microsoft.ServiceFabric.Actors.KVSToRCMigration/TargetMigrationOrchestrator.cs
jyotimisramsft: Cleaned up the logic as suggested

src/Microsoft.ServiceFabric.Actors.KVSToRCMigration/TargetMigrationOrchestrator.cs
```diff
@@ -588,21 +594,33 @@ private async Task<IMigrationPhaseWorkload> NextWorkloadRunnerAsync(PhaseResult
         {
             var endSN = await this.GetEndSequenceNumberAsync(cancellationToken);
             var delta = endSN - currentResult.EndSeqNum;
+            var isDowntimeInvoked = false;
```
roshanroutray, line 597: NIT: Please use isDowntimeInvoked or IsDowntimeAllowed for local as well as meta data store., to avoid confusion later.

roshanroutray (approved):

src/Microsoft.ServiceFabric.Actors.KVSToRCMigration/TargetMigrationOrchestrator.cs
jyotimisramsft: Renamed all occurrences to isDowntimeInvoked

pakunapamsft (approved):

PR #349, opened 2023-03-20
Pass cancellation token to the resolver

### LINKED ISSUE #348: Resolver should respect the cancellation token parameter
`ResolveHelperAsync` currently ignores the cancellation token parameter in the call to `resolveFunc`. As a result, when the caller requests that the call is cancelled through the passed token, the request will still hang.
https://github.com/microsoft/service-fabric-services-and-actors-dotnet/blob/4c399431fe60ac350dd0d99c139679bac507b53f/src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs#L569-L573

yashagarwal23:
Pass the cancellation token to the resolver func.
Fixes #348

### DIFF
--- a/src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs
+++ b/src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs
@@ -621,13 +621,15 @@ namespace Microsoft.ServiceFabric.Services.Client
                             // Registering for Notification only for the first request for a service uri.
                             if (this.UseNotification && !this.registrationCache.ContainsKey(serviceUri))
                             {
-                                ServiceNotificationFilterDescription filter = new ServiceNotificationFilterDescription(
-                                    name: serviceUri,
-                                    matchNamePrefix: true,
-                                    matchPrimaryChangeOnly: false);
-                                await client.ServiceManager.RegisterServiceNotificationFilterAsync(filter, totaltime.GetRemainingTime(), cancellationToken);
-
-                                this.registrationCache.TryAdd(serviceUri, true);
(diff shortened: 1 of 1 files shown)


src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs
```diff
@@ -621,15 +621,13 @@ public static ServicePartitionResolver GetDefault()
...
-                                var added = this.registrationCache.TryAdd(serviceUri, true);
-                                if (added)
-                                {
-                                    ServiceNotificationFilterDescription filter = new ServiceNotificationFilterDescription(
-                                       name: serviceUri,
-                                       matchNamePrefix: true,
-                                       matchPrimaryChangeOnly: false);
-                                    await client.ServiceManager.RegisterServiceNotificationFilterAsync(filter, totaltime.GetRemainingTime(), CancellationToken.None);
-                                }
+                                ServiceNotificationFilterDescription filter = new ServiceNotificationFilterDescription(
```
pakunapamsft, line 624: Why did the order of instructions change? i.e, why TryAdd moved to after registration?

pakunapamsft (approved):

src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs
yashagarwal23: Actually I just saw, we are removing the service uri entry from the registrationCache in case the RegisterServiceNotificationFilterAsync fails. So the previous flow would have worked also.
Let me change back to that.

PR #352, opened 2023-05-31
update build to include visual studio 2022 or 2019

edwardgarza:


build.ps1
```diff
@@ -38,24 +38,30 @@ if($MSBuildFullPath -eq "")
 {
     if (Test-Path "env:\ProgramFiles(x86)")
```
edwardgarza, line 39: What is the point of this? A lot of machines have both an x86 and non x68 program file path. Should this be inverted to look like 

if (programfiles) use programfiles
else if(programfilesx86) use programfilesx86

olegsych:
Abandoned

PR #360, opened 2023-12-26
Actor reminders related tracing improvements

yashagarwal23:
Adding reminder related tracing for following paths - 
- Register Reminder
  - Saving reminder in state provider
- Unregister Reminder 
- Firing Reminder 
  - Arming Reminder
- Reminder Next Due Time Computation 
- Loading Reminders when actor primary starts/moves

### DIFF
--- a/src/Microsoft.ServiceFabric.Actors/Runtime/ActorManager.cs
+++ b/src/Microsoft.ServiceFabric.Actors/Runtime/ActorManager.cs
@@ -436,6 +436,13 @@ namespace Microsoft.ServiceFabric.Actors.Runtime
 
             try
             {
+                ActorTrace.Source.WriteInfoWithId(
+                    TraceType,
+                    this.traceId,
+                    "Firing reminder ({0}) for actor ({1})",
+                    reminder.Name,
+                    reminder.OwnerActorId.GetStorageKey());
+
                 using (var actorScope = this.GetActor(reminder.OwnerActorId, true, false))
                 {
                     var actorBase = actorScope.Actor;
@@ -1233,12 +1240,17 @@ namespace Microsoft.ServiceFabric.Actors.Runtime
 
         private async Task LoadRemindersAsync(CancellationToken cancellationToken)
         {
+            ActorTrace.Source.WriteInfoWithId(
+                    TraceType,
+                    this.traceId,
+                    $"Loading reminders.");
+
             var reminders = await this.StateProvider.LoadRemindersAsync(cancellationToken);
 
             ActorTrace.Source.WriteInfoWithId(
                     TraceType,
                     this.traceId,
-                    $"Loading {reminders.Count} reminders.");
+                    $"Found {reminders.Count} reminders.");
 
             if (reminders.Count > 0 && !this.actorService.ActorTypeInformation.IsRemindable)
             {
@@ -1306,7 +1318,7 @@ namespace Microsoft.ServiceFabric.Actors.Runtime
                 TraceType,
                 this.traceId,
                 "Registering reminder for actor: ({0}), reminderName: ({1}), remainingDueTime: ({2}),  saveState {3}",
-                actorReminder.OwnerActorId,
+                actorReminder.OwnerActorId.GetStorageKey(),
                 actorReminder.Name,
                 remainingDueTime,
                 saveState);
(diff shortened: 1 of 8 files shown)


src/Microsoft.ServiceFabric.Actors/Runtime/ActorReminder.cs
```diff
@@ -124,6 +124,14 @@ internal void ArmTimer(TimeSpan newDueTime)
             {
                 try
                 {
+                    this.actorManager.TraceSource.WriteInfoWithId(
+                        TraceType,
+                        this.actorManager.GetActorTraceId(this.OwnerActorId),
+                        "Arming timer for reminder ({0}), actor: ({1}), newDueTime: ({2})",
+                        this.Name,
+                        this.OwnerActorId.GetStorageKey(),
```
pakunapamsft, line 132: Why storage key? Let us ensure we trace actor id every where not confuse with storage keys in someplaces.

src/Microsoft.ServiceFabric.Actors/Runtime/ActorReminderState.cs
```diff
@@ -34,10 +37,26 @@ internal ActorReminderState(ActorReminderData reminder, TimeSpan currentLogicalT
             if (reminderCompletedData != null)
             {
                 this.nextDueTime = ComputeRemainingTime(currentLogicalTime, reminderCompletedData.LogicalTime, reminder.Period);
+                ActorTrace.Source.WriteInfo(
```
pakunapamsft, line 40: Emit current logical time as well which provides insights into how next due time is calculated

src/Microsoft.ServiceFabric.Actors/Runtime/ActorReminderState.cs
```diff
@@ -34,10 +37,26 @@ internal ActorReminderState(ActorReminderData reminder, TimeSpan currentLogicalT
...
+                    reminder.ActorId.GetStorageKey(),
+                    reminderCompletedData.LogicalTime,
+                    reminder.Period,
+                    this.nextDueTime);
             }
             else
             {
                 this.nextDueTime = ComputeRemainingTime(currentLogicalTime, reminder.LogicalCreationTime, reminder.DueTime);
+                ActorTrace.Source.WriteInfo(
+                    TraceType,
```
pakunapamsft, line 53: Same as above

src/Microsoft.ServiceFabric.Actors/Runtime/ActorManager.cs
```diff
@@ -436,6 +436,13 @@ public async Task FireReminderAsync(ActorReminder reminder)
 
             try
             {
+                ActorTrace.Source.WriteInfoWithId(
```
pakunapamsft, line 439: - Emit a log before acquiring the actor lock
- Emit a log after acquiring the actor lock 
(Basically add traces for the lock events which are going into diagnostics events). 
This will help us answer the queries around "My reminder did not fire on time" related questions.

src/Microsoft.ServiceFabric.Actors/Runtime/ActorManager.cs
```diff
@@ -436,6 +436,13 @@ public async Task FireReminderAsync(ActorReminder reminder)
 
             try
             {
+                ActorTrace.Source.WriteInfoWithId(
+                    TraceType,
+                    this.traceId,
```
pakunapamsft, line 441: Add log to emit time taken to load reminders in the context of runasync.

src/Microsoft.ServiceFabric.Actors/Runtime/ActorReminder.cs
yashagarwal23: Storage key is just {Actor Type}_{Actor Id}, this might help us keep track of the actor id type for the particular actor.

src/Microsoft.ServiceFabric.Actors/Runtime/ActorManager.cs
yashagarwal23: I have added traces at the start and the end of the ActorManager.LoadRemindersAsync method (this is called by ActorService.RunAsync when it is trying to load the reminders). We can get the total time taken to load the reminders by these 2 traces.

src/Microsoft.ServiceFabric.Actors/Runtime/ActorReminderState.cs
yashagarwal23: Done

pakunapamsft (approved):

PR #367, opened 2024-04-09
Enable some new test assemblies to be friends

ChrisBurgessMS:
Testing of Service Fabric requires the ability to reach into this assembly and use internal classes.  There are new binaries that need access to this assembly's internals so the assembly needs new entries to make those internal accessible.


olegsych (approved):

olegsych:
For some reason the build isn't getting triggered like it did for the release_7.0 PR. I'll try to close and reopen the PR.

olegsych (approved):

PR #368, opened 2024-04-09
Make build.yml trigger for all release_* branches

olegsych:
To enable PR validation builds for all release_* branches in addition to the develop branch.

Also remove v* tag trigger because it's not being used.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,14 +4,12 @@ on:
   push:
     branches:
       - develop
-      - release_4*
-    tags:
-      - v*
+      - 'release_*'
     
   pull_request:
     branches: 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -4,14 +4,12 @@ on:
   push:
     branches:
       - develop
```
ChrisBurgessMS, line 6: There is also a master branch. Do you want to include that?

ChrisBurgessMS (approved):

.github/workflows/build.yml
olegsych: Since it's 7 years behind, I think we should just delete it.

PR #375, opened 2024-06-19
Remove state Actor migration support

olegsych:
The KVS to RC migration projects were created years ago based on the direction from the Service Fabric Data team, who didn't want to support automatic migration. Since then, the strategy has changed, the migration effort was abandoned and these projects haven't been maintained.

In addition, to being abandoned, these projects have a dependency on the unsafe packages, such as Newtonsoft.Json, which are now reported by the component governance. Since there is no plan to resume this effort, the abandoned code is removed.


olegsych:
@pakunapamsft what do you think about removing the KVS to RC migration code from this repo? Since this effort was abandoned and now has been flagged by the component governance, I thought it would make more sense to remove it rather than maintain it.
https://dev.azure.com/msazure/One/_componentGovernance/service-fabric-services-and-actors-dotnet

When I started this experiment, I didn't realize that we also have public API in the `Microsoft.ServiceFabric.Actors.Migration` namespace. If I understand correctly, it would be no longer needed after removal of the KVS to RC migration code. If that's correct, removing it would be great as well (with deprecation in the previous versions, of course).

pakunapamsft (approved):

olegsych:
This is too stale now, I'll ask the Copilot to redo this.

PR #376, opened 2024-07-19
Update PublicKey of EventsValidationTest

gobradovic-MSFT:
EventsValidationTest project has been changed which caused the change in its PublicKey.

Updating the PublicKey so that EventsValidationTest can still access the internals of these classes.


olegsych:
Thanks @gobradovic-MSFT! Could you update the PR description to explain why we're making this change? https://cbea.ms/git-commit/#why-not-how

olegsych (approved):

PR #381, opened 2024-09-03
Fix CancelOnTimerWithLargeRetryDelay test flakiness due to Stopwatch imprecision

gobradovic:
CancelOnTimerWithLargeRetryDelay fails on some environments due 
to the fact that Stopwatch class may report incorrect timing results on 
different processors, as noted in official documentation: 
https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch?view=net-8.0

### DIFF
--- a/test/unittests/Microsoft.ServiceFabric.Services.Tests/ServicePartitionClientTests.cs
+++ b/test/unittests/Microsoft.ServiceFabric.Services.Tests/ServicePartitionClientTests.cs
@@ -109,12 +109,20 @@ namespace Microsoft.ServiceFabric.Services.Tests
             var retryCount = (int)(2 * (clientRetryTimeout.Ticks / DefaultRetryDelay.Ticks));
             var retryDelay = TimeSpan.FromTicks(clientRetryTimeout.Ticks * 2);
 
+            // Set processor affinity to avoid bug in time measurement by Stopwatch class.
+            #pragma warning disable CA1416
+            Process proc = Process.GetCurrentProcess();
+            long affinityMask = 0x0001;
+            proc.ProcessorAffinity = (IntPtr)affinityMask;
+            #pragma warning restore CA1416
+
             var sw = new Stopwatch();
             sw.Start();
             var result = await this.SetupCancelTestAsync(
                 clientRetryTimeout,
                 retryCount,
                 retryDelay);
+            sw.Stop();
 
             sw.ElapsedMilliseconds.Should().BeGreaterThan((long)clientRetryTimeout.TotalMilliseconds, "Should be longer than the ClientRetryTimeout.");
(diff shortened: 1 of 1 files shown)


test/unittests/Microsoft.ServiceFabric.Services.Tests/ServicePartitionClientTests.cs
```diff
@@ -109,12 +109,20 @@ public async Task CancelOnTimerWithLargeRetryDelay()
             var retryCount = (int)(2 * (clientRetryTimeout.Ticks / DefaultRetryDelay.Ticks));
             var retryDelay = TimeSpan.FromTicks(clientRetryTimeout.Ticks * 2);
 
+            // Set processor affinity to avoid bug in time measurement by Stopwatch class.
+            #pragma warning disable CA1416
+            Process proc = Process.GetCurrentProcess();
+            long affinityMask = 0x0001;
+            proc.ProcessorAffinity = (IntPtr)affinityMask;
```
olegsych, line 116: I noticed we have several tests that compare sw.ElapsedMilliseconds with clientRetryTimeout.TotalMilliseconds. I suspect that the CancelAfterVall test subtracts 10ms from it to deal with the exact same problem. Interestingly enough, the CancelOnTimer test doesn't have this logic, but I don't remember it failing consistently.

We should probably make all three of these tests use a consistent approach, otherwise the third one might start failing one day too. The -10 approach seems to be simplest. I would only extract it to a constant with a descriptive name, like StopwatchPrecision, and a link to the doc page in a comment. I like the ProcessorAffinity workaround less because it has global effect and could change behavior of any other test that happens to run at the same time. It's not unusual for async tests to use multiple threads, so restricting processor affinity to just one could cause false results or deadlocks.

gobradovic: Done (I changed the approach and renamed the PR). One unknown is whether the Stopwatch imprecision can occur only in one direction. If it can report longer times as well, we need to also adjust `BeLessThan` checks. Another question is whether the error can be larger than 10ms. If it can, tests can still show flakiness (hopefully not, as then the Stopwatch would be pretty useless then). I do agree with the argument that restricting the affinity can affect the tested behavior (to lessen this, I was setting processor rather than thread affinity to hopefully still have multithreading).

olegsych (approved):

test/unittests/Microsoft.ServiceFabric.Services.Tests/ServicePartitionClientTests.cs
olegsych: As we discussed in scrum today, these are good points. A 10ms discrepancy is significant in a context of a 1s time interval. I don't know if many of the time intervals used by customers would be this aggressive though. 1s interval looks like an attempt to make test execution time as short as possible to compensate for the lack of appropriate time abstractions that would allow mocking current time and timer callbacks in tests. Your call.

PR #386, opened 2024-09-15
Avoid locking on each call to ServicePartitionResolver.GetDefault()

gobradovic:
Each call to ServicePartitionResolver.GetDefault() acquires a lock due to lazy 
initialization. We could use double-checked locking to avoid this, but as the 
object is not heavy, and its creation is not expensive we can simply 
use eager initialization.

ServicePartitionResolver.SetDefault() can update this field but reads and 
writes of reference types are atomic. If needed, field can be declared as volatile.

### DIFF
--- a/src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs
+++ b/src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs
@@ -42,9 +42,14 @@ namespace Microsoft.ServiceFabric.Services.Client
 
         private static readonly string TraceType = typeof(ServicePartitionResolver).Name;
 
-        private static readonly object StaticLock = new object();
-        private static ServicePartitionResolver defaultResolver;
-        private static RandomGenerator randomGenerator = new RandomGenerator();
+        private static ServicePartitionResolver DefaultResolver = new ServicePartitionResolver(
+                        () => new FabricClient(
+                            new FabricClientSettings()
+                            {
+                                PartitionLocationCacheBucketCount = 4096,
+                                PartitionLocationCacheLimit = 4096,
+                            }));
+        private static readonly RandomGenerator RandomGenerator = new RandomGenerator();
         private readonly object thisLock = new object();
         private readonly CreateFabricClientDelegate createFabricClient;
         private readonly CreateFabricClientDelegate recreateFabricClient;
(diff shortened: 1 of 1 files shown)


src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs
```diff
@@ -146,10 +151,7 @@ public ServicePartitionResolver(
         /// <param name="defaultServiceResolver">The new default value</param>
         public static void SetDefault(ServicePartitionResolver defaultServiceResolver)
         {
-            lock (StaticLock)
-            {
-                defaultResolver = defaultServiceResolver;
-            }
+            DefaultResolver = defaultServiceResolver;
```
olegsych, line 154: I don't think this is likely, but technically, I think locking was necessary because object variables are not guaranteed to be atomic: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/variables#96-atomicity-of-variable-references. I would use a lighter form of locking though: https://learn.microsoft.com/en-us/dotnet/api/system.threading.interlocked.exchange?view=net-8.0#system-threading-interlocked-exchange-1(-0@-0)

src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs
```diff
@@ -165,21 +167,7 @@ public static void SetDefault(ServicePartitionResolver defaultServiceResolver)
...
-                            new FabricClientSettings()
-                            {
-                                PartitionLocationCacheBucketCount = 4096,
-                                PartitionLocationCacheLimit = 4096,
-                            }));
-                }
-
-                return defaultResolver;
-            }
+            return DefaultResolver;
```
olegsych, line 170: I think this is a breaking change for this API. Before, it was possible to SetDefault(null) and next GetDefault() would return a new defaultResolver. Now, the GetDefault() would return null, which could result in NullReferenceException thrown from calling code. I don't think this is a safe change to make.

src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs
gobradovic: As discussed offline, as this is a reference type, per the specification variable assignment is atomic, so this part should not be a problem.

src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs
gobradovic: That's a quite good catch, I didn't consider that scenario. I'll close this PR.

gobradovic:
Closing this PR as it would introduce a breaking change.

PR #389, opened 2024-10-01
Resolving vulnerable transitive dependencies

plave0:
Overriding versions of transitive dependencies that were flagged as vulnerable on nuget.org.

Internal ADO work item: https://dev.azure.com/msazure/One/_workitems/edit/29603592


gobradovic (approved):

olegsych (commented):
publish review

PR #392, opened 2024-10-04
Updating versions of transitive dependencies

plave0:
This PR addresses unaddressed comments form #389.

Internal ADO work item: https://dev.azure.com/msazure/One/_workitems/edit/29603592

### DIFF
--- a/src/Microsoft.ServiceFabric.Actors.KVSToRCMigration/Microsoft.ServiceFabric.Actors.KVSToRCMigration.csproj
+++ b/src/Microsoft.ServiceFabric.Actors.KVSToRCMigration/Microsoft.ServiceFabric.Actors.KVSToRCMigration.csproj
@@ -34,8 +34,9 @@
       defined by the direct dependencies are marked as vulnerable packages on nuget.org.
     -->
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
-    <PackageReference Include="System.Security.Cryptography.Xml" Version="9.0.0-rc.1.24431.7" />
-    <PackageReference Include="System.Text.Encodings.Web" Version="9.0.0-rc.1.24431.7" />
+    <PackageReference Include="System.Security.Cryptography.Xml" Version="8.0.1" />
+    <PackageReference Include="System.Text.Encodings.Web" Version="8.0.0" />
+    <PackageReference Include="System.Formats.Asn1" Version="8.0.1" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


src/Microsoft.ServiceFabric.Actors.KVSToRCMigration/Microsoft.ServiceFabric.Actors.KVSToRCMigration.csproj
```diff
@@ -29,9 +29,13 @@
...
     <!-- Overriding versions of transitive dependencies -->
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
-    <PackageReference Include="System.Security.Cryptography.Xml" Version="4.7.1" />
-    <PackageReference Include="System.Text.Encodings.Web" Version="4.5.1" />
+    <!-- 
+      Some versions of transitive dependencies have to be overwritten, since the default versions
+      defined by the direct dependencies are marked as vulnerable packages on nuget.org.
+    -->
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
+    <PackageReference Include="System.Security.Cryptography.Xml" Version="9.0.0-rc.1.24431.7" />
```
olegsych, line 37: We should stick to the latest stable versions of the packages. I'm not sure if we can use the 9.0 versions at this point, but since they are still pre-release, it's a moot point. 

Here is what pre-release means: https://semver.org/#spec-item-9

olegsych (changes requested):

src/Microsoft.ServiceFabric.Actors.KVSToRCMigration/Microsoft.ServiceFabric.Actors.KVSToRCMigration.csproj
plave0: @olegsych thank you for this insight. My reasoning for using the prerelease versions was that using the latest stable version of `System.Security.Cryptography.Xml` introduced another transitive vulnerable dependency: `System.Formats.Asn1:8.0.0`, which is not present in the prerelease version of `System.Security.Cryptography.Xml`. 

I reverted all packages to their latest table version, and resolved this vulnerable transitive dependency by, again, overriding it's verion:
```
<PackageReference Include="System.Formats.Asn1" Version="8.0.1" />
```

Nevertheless, it is very strange to me that we are encountering so many of these vulnerable transitive dependencies.

olegsych (approved):

PR #393, opened 2024-11-18
Implement central package versioning

plave0:
Many dependencies are shared between projects. Central package management (CPM) prevents discrepancies of dependency versions between projects and reduces the duplication of version information across projects. 

Previously, versions of transitive dependencies which have been marked as vulnerable have been overwritten to safe versions. CPM enables us to overwrite versions of vulnerable transitive dependencies in a centralized manner, which makes maintaining them easier. This change also enables us to exclude information about overwritten transitive dependencies from each individual project, which reduces clutter in project files.

Remove unused NuGet projects.

After implementing CPM, NuGet projects (all projects under the `nuproj` directory, except `SF.ActorsServices.Internal`) started causing warning on build. Since they are no longer in use, and it is not planned to use them in the future, they have been deleted form the repository.

Update versions of ASP.NET Core NuGet packages

Update the following packages:
- `Microsoft.AspNetCore.Diagnostics` has been downgraded to version `2.1.1`.
- `Microsoft.AspNetCore.Hosting` has been downgraded to version `2.1.1`.
- `Microsoft.AspNetCore.HttpsPolicy` has been downgraded to version `2.1.1`.
- `Microsoft.AspNetCore.Mvc` has been downgraded to version `2.1.3`.
- `Microsoft.AspNetCore.Server.HttpSys` has been downgraded to version `2.1.12`.

These packages have been downgraded in order to reflect the support policy of NuGet based ASP.NET Core 2.1 packages which are being used for building our .NET Framework 4.6 packages (https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core).


gobradovic (approved):

Directory.Packages.props
```diff
@@ -0,0 +1,44 @@
+<Project>
```
olegsych, line 1: Could you expand the PR description to explain that we are implementing central package management to pin transient dependencies to secure versions as well as to reduce the duplication of version information and the risk of discrepancies? https://cbea.ms/git-commit/#why-not-how

Also, could you change the title to imperative mood? https://cbea.ms/git-commit/#imperative

PR #401, opened 2025-04-03
Move Microsoft.ServiceFabric.Diagnostics project to GitHub

ilijagrbic:
We will need to consume SF metrics API on GitHub. This is setup work to enable us to move Microsoft.ServiceFabric.Telemetry project to GitHub as well.


ilijagrbic:
> @ilijagrbic please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement
@microsoft-github-policy-service agree company="Microsoft"

src/Microsoft.ServiceFabric.Diagnostics/Microsoft.ServiceFabric.Diagnostics.csproj
```diff
@@ -0,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+      <Description>Microsoft.ServiceFabric.Diagnostics Library</Description>
```
olegsych, line 3: Could you leave only those properties that are actually being used. The Description is optional, and the current text doesn't add any new information to the assembly name. I'm not sure about the PackageId. At some point we may need it, but shouldn't while we still build this package in the WindowsFabric repo. Let's try to remove it for now.

src/Microsoft.ServiceFabric.Diagnostics/Microsoft.ServiceFabric.Diagnostics.csproj
```diff
@@ -0,0 +1,28 @@
...
+      <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+      <AssemblyName>Microsoft.ServiceFabric.Diagnostics</AssemblyName>
+      <PackageId>Microsoft.ServiceFabric.Diagnostics</PackageId>
+      <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
```
olegsych, line 7: I don't see the Microsoft.ServiceFabric.Diagnostics.dll being included in the SF.ActorsServices.Internal.nupkg. I think the problem is that right now the service_fabric_managed_common.props does this:

    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>

We need to set this property back to true in this project because thats what the CopyProdBinaries_* targets expect.

olegsych (changes requested):
Thanks for the PR @ilijagrbic. I've added a few requests in the comments. The main criteria for completion of this PR is the appearance of the Microsoft.ServiceFabric.Diagnostics.dll in the SF.ActorsServices.Internal.nupkg produced by the build.

src/Microsoft.ServiceFabric.Diagnostics/Microsoft.ServiceFabric.Diagnostics.csproj
```diff
@@ -0,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
ilijagrbic: Resolved.

src/Microsoft.ServiceFabric.Diagnostics/Microsoft.ServiceFabric.Diagnostics.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <Import Project="..\..\properties\service_fabric_managed_netstandard.props" />
+  </Target>
+  <Target Name="ImportNetCoreAppProps" Condition="'$(TargetFramework)' == 'net462'">
+    <Import Project="..\..\properties\service_fabric_managed_netframework.props" />
+  </Target>
+  
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ServiceFabric" />
+  </ItemGroup>
+  <!--Publishing Directives-->
```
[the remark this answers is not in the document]
ilijagrbic: I left the code commented out since I started by copying csproj from an existing project in WindowsFabric, and I was unsure if we need these directives. I will remove it now.

src/Microsoft.ServiceFabric.Diagnostics/Microsoft.ServiceFabric.Diagnostics.csproj
ilijagrbic: I will remove the properties you mentioned as unused. I saw in other projects that we have some other properties, but I was not sure which ones do we need in the beginning. For example, here are the properties from Microsoft.ServiceFabric.Services project:
```
    <ProjectGuid>{F416C4C3-250D-48F3-822C-1C589E8F851C}</ProjectGuid>
    <RootNamespace>Microsoft.ServiceFabric.Services</RootNamespace>
    <RootNamespace>$(AssemblyName)</RootNamespace>
    <AssemblyTitle>$(AssemblyName)</AssemblyTitle>
    <DocumentationFile>$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
    <TargetFramework>net462</TargetFramework>
```
I assume DocumentationFile and AssemblyTitle are probably not needed right now. But I will investigate these properties a bit more today. :)

src/Microsoft.ServiceFabric.Diagnostics/Microsoft.ServiceFabric.Diagnostics.csproj
```diff
@@ -0,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+      <Description>Microsoft.ServiceFabric.Diagnostics Library</Description>
+      <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+      <AssemblyName>Microsoft.ServiceFabric.Diagnostics</AssemblyName>
+      <PackageId>Microsoft.ServiceFabric.Diagnostics</PackageId>
+      <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
+  </PropertyGroup>
+
+  <Target Name="ImportNetStandardProps" Condition="'$(TargetFramework)' == 'netstandard2.0'">
```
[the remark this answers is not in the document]
ilijagrbic: I'm not sure how should I build the project to test/confirm this. Currently I'm building with `dotnet.exe build` command. And I don't get any output regarding properties, regardless of Condition parameter (even if I remove condition, I still don't get any output). It is the same for other projects when I build them, so I'm probably doing something wrong. 

I also haven't found anything regarding the -bl flag in the [documentation here ](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-build)

I will change the imports section to:
```
<Import Condition="'$(TargetFramework)' == 'netstandard2.0'" Project="..\..\properties\service_fabric_managed_netstandard.props" />
  <Import Condition="'$(TargetFramework)' == 'net462'" Project="..\..\properties\service_fabric_managed_netframework.props" />
```
And hopefully I will figure out a way to confirm if props are being used.

src/Microsoft.ServiceFabric.Diagnostics/Microsoft.ServiceFabric.Diagnostics.csproj
ilijagrbic: I'm not sure about this, since I see the dlls in both the artefacts folder
image

And when I use nuget package explorer for the nuget packet produced by the pipeline I also see the dlls:
image

I have updated [service_fabric_managed_netframework.targets](https://github.com/microsoft/service-fabric-services-and-actors-dotnet/pull/401/files#diff-bbd971383f42c7fad0dfce5153fcc31e17a9e7dafc6e9707ff1e88788a4f2a85) and [service_fabric_managed_netstandard.targets](https://github.com/microsoft/service-fabric-services-and-actors-dotnet/pull/401/files#diff-a185c32d59b2e7973e9f2e78242f80baa558744620b1ebc9ec4669e1f50d7d90) files in the original commit to copy dlls from the good location, at least that was my intention, maybe I didn't do it properly.

src/Microsoft.ServiceFabric.Diagnostics/Microsoft.ServiceFabric.Diagnostics.csproj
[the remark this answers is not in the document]
ilijagrbic: I think I managed to dig up where properties are checked in binlog, and it seems that the new version of import works. Here is a screenshot:
image

We can talk on scrum to confirm this is the way we want the build to behave.

src/Microsoft.ServiceFabric.Diagnostics/Microsoft.ServiceFabric.Diagnostics.csproj
ilijagrbic: Sorry for the previous deleted comment Oleg. Before service_fabric_managed_common.props was not properly imported into the project, but now that I fixed the import, this also needs to be added, as you correctly pointed out. I added it now.

src/Microsoft.ServiceFabric.Diagnostics/Microsoft.ServiceFabric.Diagnostics.csproj
plave0: @ilijagrbic we might need the `ProjectGuid` property here so we can include the project in `code.sln`

ilijagrbic: @plave0 I have already included it without this property, and I think Visual Studio generated an Guid for this in code.sln. Do you think I should just copy that generated Guid to this csproj file?

plave0: Sorry, I was looking at changes from one commit only, and I did not see that you already made changes to `code.sln`. In case it works without specifying the `ProjectGuid` property, I would say there is no need to use it then.

plave0 (approved):

src/Microsoft.ServiceFabric.Diagnostics/Microsoft.ServiceFabric.Diagnostics.csproj
ilijagrbic: Thanks Pavle! I will resolve this comment then and won't introduce any more changes for now. :)

olegsych (approved):
The changes on the GitHub side look good @ilijagrbic. Could you cross link it with the WindowsFabric PR? It would be great to review them both at the same time before we merge.

PR #404, opened 2025-04-14
Remove unused DebugUtility class

gobradovic:
`DebugUtility` class is internal and not referenced anywhere.

Removing it as unused.


olegsych (approved):
Thanks @gobradovic. I did a quick search in the WindowsFabric and AspNetCore repos and didn't find any usages of this class.

PR #407, opened 2025-04-16
Assure cache lock release and reporting improvements

gobradovic:
Release semaphore in finally block to ensure proper cleanup.

Add info level traces for start of the cleanup process and situations where we didn't acquire lock in time. Both situations should be fairly rare (not verbose).


src/Microsoft.ServiceFabric.Services/Communication/Client/CommunicationClientCache.cs
```diff
@@ -251,30 +256,45 @@ public void CleanupCacheEntries(out int totalNumberOfItems, out int numberOfEntr
...
                 {
                     ++totalNumberOfItems;
-                    if (!entry.Value.Semaphore.Wait(this.cacheEntryLockWaitTimeForCleanup))
-                    {
-                        // If the wait cannot be satisfied in a short time, then it indicates usage for
-                        // this cache entry, so it is ok to skip this entry in the current cleanup run.
-                        continue;
-                    }
 
                     if (!entry.Value.IsCommunicationClientValid())
```
olegsych, line 260: Since the IsCommunicationClientValid changes the CommunicationClientCacheEntry.client, I would assume that the Semaphore needs to be acquired first. I would undo this particular change, leaving the traces and the try-finally.

olegsych (approved):

olegsych:
Thanks @gobradovic. Everything looks good. Could you double-check the PR description and commit message for accuraty before you merge? I think it may not be entirely accurate anymore.

PR #408, opened 2025-04-23
Remove not needed double projects

ilijagrbic:
We used to have multiple projects that would essentially just serve to build the same code in NET Framework and NET Standard flavors. 
Since we use SDK style projects, there is no need for this, since they allow 1 project to be built with multiple frameworks. 
This PR removes old projects, and updates configuration for the original projects to be able to be built using NET Standard.


src/Microsoft.ServiceFabric.Services/Microsoft.ServiceFabric.Services.csproj
```diff
@@ -1,19 +1,27 @@
...
-    <TargetFramework>net462</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
+    <AppendTargetFrameworkToOutputPath>true</AppendTargetFrameworkToOutputPath>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.ServiceFabric" />
     <PackageReference Include="Microsoft.ServiceFabric.Data" />
     <PackageReference Include="Microsoft.ServiceFabric.Diagnostics.Internal" />
   </ItemGroup>
+  <!-- Exclude AssemblyInfo.cs for netstandard builds -->
```
olegsych, line 20: It looks like we only have [CLSCompliant(true)] attributes in the AssemblyInfo files. This is rather pointless because System.Fabric is not CLS-compliant. CLS compliance was a big thing 20 years ago when we were trying to move the Visual Basic developers to .NET. However, as far as I know, we don't support Visual Basic for SF. So, instead of pretending that we do in some of the SDK assemblies, let's just remove these files and this conditional logic.

src/Microsoft.ServiceFabric.Services/Microsoft.ServiceFabric.Services.csproj
```diff
@@ -1,19 +1,27 @@
...
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
```
olegsych, line 21: When I tried removing the exclusion of the AssemblyInfo.cs and built this project, it reported a number of errors due to types being non-CLS-compliant. I think this indicates that we have some net462 code now included in the netstandard2.0 build. Could you double-check? I think we need to understand which types we're now adding to the netstandard2.0 assemblies. I can't think of a reason why this would be a problem, but let's understand what's changing.

src/Microsoft.ServiceFabric.Services.Remoting/Microsoft.ServiceFabric.Services.Remoting.csproj
```diff
@@ -1,21 +1,31 @@
...
   <ItemGroup>
     <PackageReference Include="Microsoft.ServiceFabric.FabricTransport.Internal" />
     <PackageReference Include="System.Diagnostics.DiagnosticSource" />
+    <PackageReference Include="System.Reflection.Emit.ILGeneration" />
+    <PackageReference Include="System.Reflection.Emit" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.ServiceFabric.Services\Microsoft.ServiceFabric.Services.csproj" />
   </ItemGroup>
+  <!-- Exclude AssemblyInfo.cs for netstandard builds -->
```
olegsych, line 24: Let's delete the AssemblyInfo.cs and remove this conditional logic.

src/Microsoft.ServiceFabric.Actors/Microsoft.ServiceFabric.Actors.csproj
```diff
@@ -23,6 +27,10 @@
   <ItemGroup>
     <Reference Include="System.Net.Http" />
   </ItemGroup>
+  <!-- Exclude AssemblyInfo.cs for netstandard builds -->
```
olegsych, line 30: Let's delete the AssemblyInfo.cs and remove this conditional logic.

src/Microsoft.ServiceFabric.Actors.KVSToRCMigration/Microsoft.ServiceFabric.Actors.KVSToRCMigration.csproj
```diff
@@ -31,4 +34,8 @@
     <Reference Include="System.ComponentModel.DataAnnotations" />
     <Reference Include="System.Web" />
   </ItemGroup>
+  <!-- Exclude AssemblyInfo.cs for netstandard builds -->
```
olegsych, line 37: Let's delete the AssemblyInfo.cs and remove this conditional logic.

code.sln
```diff
@@ -15,18 +15,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ServiceFabric.Act
...
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ServiceFabric.Services_netstandard", "src\netstandard\Microsoft.ServiceFabric.Services\Microsoft.ServiceFabric.Services_netstandard.csproj", "{E93A1CC5-C164-4C4D-8314-A35B4B3AF633}"
-EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ServiceFabric.Services.Remoting_netstandard", "src\netstandard\Microsoft.ServiceFabric.Services.Remoting\Microsoft.ServiceFabric.Services.Remoting_netstandard.csproj", "{C6FF32EC-DCB6-4684-ACB5-4D8A9B6BB590}"
-EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ServiceFabric.Actors_netstandard", "src\netstandard\Microsoft.ServiceFabric.Actors\Microsoft.ServiceFabric.Actors_netstandard.csproj", "{997C67FC-4571-4DAC-8C2F-9BB62756AFFA}"
-EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "FabActUtil_netstandard", "src\netstandard\FabActUtil\FabActUtil_netstandard.csproj", "{BD88FA2D-1F61-4107-8113-8D78FB5697F3}"
-EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "dotnetframework", "dotnetframework", "{861D6F40-D3CE-46CF-B8AD-0215B9F38BE1}"
```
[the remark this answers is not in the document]
ilijagrbic: Makes sense, I will make sure to do that.

src/Microsoft.ServiceFabric.Services/Microsoft.ServiceFabric.Services.csproj
ilijagrbic: Yeah, this makes sense. I notices we were not using CLS compliance in NET Standard projects, but I was not sure if there was a good reason to keep it on NET Framework. Anyways, I will remove the files.

code.sln
```diff
@@ -46,8 +36,6 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ServiceFabric.Ser
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ServiceFabric.Actors.KVSToRCMigration", "src\Microsoft.ServiceFabric.Actors.KVSToRCMigration\Microsoft.ServiceFabric.Actors.KVSToRCMigration.csproj", "{B5E3B9D7-E91E-4B76-959D-D418E973FCAF}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ServiceFabric.Actors.KVSToRCMigration_netstandard", "src\netstandard\Microsoft.ServiceFabric.Actors.KVSToRCMigration\Microsoft.ServiceFabric.Actors.KVSToRCMigration_netstandard.csproj", "{E5B52C61-66BE-4D71-88D2-E50F78011C72}"
-EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ServiceFabric.Actors.StateMigration.Tests", "test\unittests\Microsoft.ServiceFabric.Actors.StateMigration.Tests\Microsoft.ServiceFabric.Actors.StateMigration.Tests.csproj", "{1BDC4681-FDBA-4E55-A247-5F779627A4D7}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "netfx", "netfx", "{02EA681E-C7D8-13C7-8484-4AC65E1B71E8}"
```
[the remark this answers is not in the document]
ilijagrbic: Ah, I think I missed the 2 Test projects that were also redundant. I will make sure to fix this tomorrow.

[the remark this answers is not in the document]
ilijagrbic: Do we even need to build test projects with 2 different frameworks (in this case NET9.0 and NET462)? Maybe we could just remove NET462 projects altogether?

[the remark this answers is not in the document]
olegsych: I think we still need to run the tests on net462 and net9.0 as long as the binaries we build for netfx and netcore are different. I.e. if they have different sets of source files or conditional compilation. We've replaced net8.0 with net9.0 when we tried to support this STS version, so it stands for net8.0 we do support. net462 was added for SF 11 when we needed to implement netfx specific logic in the remoting provider attributes. @plave0 can tell you more.

src/Microsoft.ServiceFabric.Services/Microsoft.ServiceFabric.Services.csproj
ilijagrbic: I have double checked this and indeed all of the .cs files were previously included in the NET Standard projects. So, removing the AssemblyInfo.cs should be sufficient for this issue.

src/Microsoft.ServiceFabric.Actors.KVSToRCMigration/Microsoft.ServiceFabric.Actors.KVSToRCMigration.csproj
```diff
@@ -27,7 +30,7 @@
     <PackageReference Include="Microsoft.AspNetCore.HttpsPolicy" />
     <PackageReference Include="Microsoft.AspNetCore.Server.HttpSys" />
   </ItemGroup>
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
```
ilijagrbic, line 33: @olegsych I'm not sure if these references are needed since the project builds even without them. I guess we have a transitive dependency to these from somewhere. Anyways, I have only included them for now only for net462, since they are present in NET Standard by default.

src/Microsoft.ServiceFabric.Actors/Microsoft.ServiceFabric.Actors.csproj
```diff
@@ -20,8 +23,8 @@
     <ProjectReference Include="..\Microsoft.ServiceFabric.Services.Remoting\Microsoft.ServiceFabric.Services.Remoting.csproj" />
     <ProjectReference Include="..\Microsoft.ServiceFabric.Services\Microsoft.ServiceFabric.Services.csproj" />
   </ItemGroup>
-  <ItemGroup>
-    <Reference Include="System.Net.Http" />
+  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
```
ilijagrbic, line 26: I'm not sure if these references are needed since the project builds even without them. I guess we have a transitive dependency to these from somewhere. Anyways, I have only included them for now only for net462, since they are present in NET Standard by default.

code.sln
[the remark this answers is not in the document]
ilijagrbic: Thanks for the answer and additional context. I have removed the net462 test project, and we are building them from the original projects, while only including .cs files that were previously used to build net462.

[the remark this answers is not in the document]
olegsych: Thanks @ilijagrbic. I love the new experience of running tests in the VS test explorer for two frameworks from a single project. Hopefully the complexity of the project file will go away once retrofit more tests to .NET Framework. We'll still have some #ifs in the .cs files, but hopefully no MSBuild magic in the .csproj files.

src/Microsoft.ServiceFabric.Actors/Microsoft.ServiceFabric.Actors.csproj
olegsych: If we can remove them without breaking the build, let's do that

src/Microsoft.ServiceFabric.Actors.KVSToRCMigration/Microsoft.ServiceFabric.Actors.KVSToRCMigration.csproj
ilijagrbic: Makes sense, I went with a safe option first, but I will remove them now.

src/Microsoft.ServiceFabric.Actors/Microsoft.ServiceFabric.Actors.csproj
ilijagrbic: Same as above, I will remove the references.

src/Microsoft.ServiceFabric.Services/Microsoft.ServiceFabric.Services.csproj
```diff
@@ -1,19 +1,23 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
-  <Import Project="..\..\properties\service_fabric_managed_netframework.props" />
+  <Import Condition="'$(TargetFramework)' == 'netstandard2.0'" Project="..\..\properties\service_fabric_managed_netstandard.props" />
+  <Import Condition="'$(TargetFramework)' == 'net462'" Project="..\..\properties\service_fabric_managed_netframework.props" />
+
```
olegsych, line 5: I feel that adding blank lines to project files to improve visibility is a losing battle. I prefer to edit projects by hand and find blank lines useful myself. Unfortunately, Visual Studio strips them when you do things like adding a reference through the Solution Explorer. Hand-crafted XML is also more difficult to preserve when running scripts for updating multiple projects. It's not the case here, but in the WindowsFabric repo, Edward does this regularly. 

If we agree on editing the project files by hand, we'll need to enforce the coding style through code review. Hopefully the number of projects and people touching them will remain small enough that we can maintain it.

src/FabActUtil/FabActUtil.csproj
```diff
@@ -1,18 +1,27 @@
...
   </PropertyGroup>
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
+    <None Include="App.config" />
+    <None Include="Microsoft.ServiceFabric.Actors_netframework.targets" Link="Microsoft.ServiceFabric.Actors.targets">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <None Include="App.config" />
```
olegsych, line 23: Consider combining the two item groups and placing Conditions on None items to reduce the duplication.

olegsych (approved):
I love how much simpler the new solution/code structure is now. At some point I would also love to eliminate the unittests subfolder from both the directory structure and the solution file. We can do it in a separate PR. Your call.

test/unittests/Microsoft.ServiceFabric.Actors.Tests/Microsoft.ServiceFabric.Actors.Tests.csproj
```diff
@@ -1,11 +1,13 @@
...
-  <Import Project="..\..\..\properties\service_fabric_managed_netstandard.props" />  
+ <Import Condition="'$(TargetFramework)' == 'net9.0'" Project="..\..\..\properties\service_fabric_managed_netstandard.props" />
+ <Import Condition="'$(TargetFramework)' == 'net462'" Project="..\..\..\properties\service_fabric_managed_netframework.props" />
   <PropertyGroup>
     <ProjectGuid>{63B58D31-66BB-4879-B4F1-0969FA3F4464}</ProjectGuid>
     <AssemblyName>Microsoft.ServiceFabric.Actors.Tests</AssemblyName>
     <RootNamespace>$(AssemblyName)</RootNamespace>
-    <TargetFramework>net9.0</TargetFramework>
+    <TargetFrameworks>net9.0;net462</TargetFrameworks>
+    <AppendTargetFrameworkToOutputPath>true</AppendTargetFrameworkToOutputPath>
```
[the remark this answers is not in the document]
ilijagrbic: We don't have it in the projects that only build on net462. Currently those are: **Microsoft.ServiceFabric.Actors.Wcf** and **Microsoft.ServiceFabric.Services.Wcf**. However, I think we should add it to shared .props, there is no harm and we can make the Wcf projects follow the same structure.

src/Microsoft.ServiceFabric.Services/Microsoft.ServiceFabric.Services.csproj
ilijagrbic: I was not aware of some of the issues with this, so this is good insight. I feel that since we are a small team, and this repo is small, we can try to enforce this through code review. But that might be something we need to decide if it is worthwhile.

ilijagrbic: If you agree, I will keep the blank line for now since it doesn't do much harm and we can talk about this in the future.

src/FabActUtil/FabActUtil.csproj
ilijagrbic: Yeah there is no need for 2 ItemGroups. I will fix this.

test/unittests/Microsoft.ServiceFabric.Actors.Tests/Microsoft.ServiceFabric.Actors.Tests.csproj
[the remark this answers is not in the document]
ilijagrbic: I have moved the AppendTargetFrameworkToOutputPath to common props and modified the "Wcf" projects to work with it.

properties/service_fabric_managed_common.props
```diff
@@ -9,7 +9,7 @@
     <WarningLevel>4</WarningLevel>
     <NoWarn></NoWarn>
     <!-- Don't append TargetFramework to output path. -->
```
olegsych, line 11: Could you remove the comment, since it's no longer accurate?
Also, could you check if the AppentTargetFrameworkToOutputPath is true by default? If it is, I think we should remove it instead of setting it explicitly. Over time things like property names change and we often end up carrying misleading and obsolete code.

src/FabActUtil/FabActUtil.csproj
```diff
@@ -1,18 +1,25 @@
...
     <DocumentationFile>$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
-    <TargetFramework>net462</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
     <OutputType>Exe</OutputType>
   </PropertyGroup>
-  <ItemGroup>
+  <ItemGroup >
     <None Include="App.config" />
-    <None Include="Microsoft.ServiceFabric.Actors.targets">
+    <None Condition="'$(TargetFramework)' == 'net462'" Include="Microsoft.ServiceFabric.Actors_netframework.targets"
```
olegsych, line 17: I think the Content items are copied to the output directory when newer by default. The App.config has special handling and needs to be included as None, but the .targets files could take advantage of this to reduce unnecessary metadata.

olegsych (approved):
Thanks @ilijagrbic! I left a couple of suggestions for your consideration. Please don't wait on my additional approvals unless you decide to make significant other changes. I trust you, so Goran's or Pavle's approvals should be just fine.

gobradovic (approved):

PR #413, opened 2025-05-22
Document GUIDs and manifests of the EventSource classes

olegsych:
Create initial tests for the EventSource classes to document the dynamically-generated GUIDs in the code and save the instrumentation manifests for inspection and registration with the ETL tools, like Event Viewer and PerfView.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -49,11 +49,11 @@ end_of_line = crlf
 # Organize usings
 dotnet_sort_system_directives_first = true
 
-# Use this.
-dotnet_style_qualification_for_field = true:warning
-dotnet_style_qualification_for_property = true:warning
-dotnet_style_qualification_for_method = true:warning
-dotnet_style_qualification_for_event = true:warning
+# Avoid this., reduce the code size and complexity to make it irrelevant instead
+dotnet_style_qualification_for_field = false:suggestion
+dotnet_style_qualification_for_property = false:suggestion
+dotnet_style_qualification_for_method = false:suggestion
+dotnet_style_qualification_for_event = false:suggestion
 
 # Use language keywords instead of BCL types
 dotnet_style_predefined_type_for_locals_parameters_members = true:suggestion
--- a/test/unittests/Microsoft.ServiceFabric.Actors.Tests/ActorEventSourceTest.cs
+++ b/test/unittests/Microsoft.ServiceFabric.Actors.Tests/ActorEventSourceTest.cs
@@ -1,0 +1,38 @@
+using System.Diagnostics.Tracing;
+using System.IO;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.ServiceFabric.Actors
+{
+    public abstract class ActorEventSourceTest
+    {
+        readonly ActorEventSource sut = ActorEventSource.Instance;
+
+        public sealed class Guid : ActorEventSourceTest
+        {
+            [Fact]
+            public void RemainsUnchangedForBackwardCompatibilityWithCollectionTools()
+            {
+                Assert.Equal(new System.Guid("e2f2656b-985e-5c5b-5ba3-bbe8a851e1d7"), sut.Guid);
+            }
+        }
+
+        public sealed class Manifest : ActorEventSourceTest
+        {
+            readonly ITestOutputHelper output;
+
+            public Manifest(ITestOutputHelper output) => this.output = output;
+
+            [Fact]
+            public void CanBeSavedForRegistrationWithExternalTools()
+            {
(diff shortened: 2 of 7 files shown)


test/unittests/Microsoft.ServiceFabric.Actors.Tests/ActorEventSourceTest.cs
```diff
@@ -0,0 +1,38 @@
+using System.Diagnostics.Tracing;
+using System.IO;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.ServiceFabric.Actors
+{
+    public abstract class ActorEventSourceTest
```
olegsych, line 8: Note the hierarchical test class structure, where each top-level class matches the SUT (system under test) class and each nested class matches a the member being tested. With the test methods collapsed, it's supposed to read as a specification. 

ActorEventSourceTest
    Guid
        RemainsUnchangedForBackwardCompatibilityWithCollectionTools
    Manifest
        CanBeSavedForRegistrationWithExternalTools

It's a bit of an overkill with just one test per member right now, but I tend to add a lot of test and came to a conclusion that creating this structure from the start is more productive than starting with a flat structure and having to refactor it soon after.

.editorconfig
```diff
@@ -49,11 +49,11 @@ end_of_line = crlf
 # Organize usings
 dotnet_sort_system_directives_first = true
 
-# Use this.
-dotnet_style_qualification_for_field = true:warning
-dotnet_style_qualification_for_property = true:warning
-dotnet_style_qualification_for_method = true:warning
-dotnet_style_qualification_for_event = true:warning
+# Avoid this., reduce the code size and complexity to make it irrelevant instead
```
olegsych, line 52: Hopefully self-explanatory. This used to be one of the unfortunate styles forced on everyone by the misguided StyleCop rules 20 years ago.

test/unittests/Microsoft.ServiceFabric.Actors.Tests/ActorEventSourceTest.cs
```diff
@@ -0,0 +1,38 @@
...
+            readonly ITestOutputHelper output;
+
+            public Manifest(ITestOutputHelper output) => this.output = output;
+
+            [Fact]
+            public void CanBeSavedForRegistrationWithExternalTools()
+            {
+                string manifest = EventSource.GenerateManifest(sut.GetType(), sut.GetType().Assembly.Location);
+                string manifestFile = Path.ChangeExtension(Path.Combine(Path.GetDirectoryName(sut.GetType().Assembly.Location), sut.Name), "man");
+                File.WriteAllText(manifestFile, manifest);
```
ilijagrbic, line 32: Can we assert that this method doesn't throw an exception? Additionally, it could be useful to assert that manifest is not empty, as that would be unexpected.

test/unittests/Microsoft.ServiceFabric.Actors.Tests/ActorEventSourceTest.cs
ilijagrbic: This is good to know. I like the structure and agree with the approach.

ilijagrbic (approved):

test/unittests/Microsoft.ServiceFabric.Actors.Tests/ActorEventSourceTest.cs
olegsych: The absence of exceptions is implicitly asserted, because if an exception is thrown the test fails.

Let me think about how to make these tests more useful. I wanted to create them primarily to allow inspecting and understanding them. Asserting that the manifest content is not empty seems performative because it would be a validation of the EventSource.GenerateManifest and not of one of EventSource descendant. I think we should assert on something defined in our code that would be important to validate through the generated manifest. The only thing I can think of is the event source name, but it would be simpler to verify it through the EventSource.Name property.

ilijagrbic: Yeah, I understand that the exception is implicitly asserted, but from my perspective when there are no asserts in the tests, it is harder to understand why and what is being tested. 

Also, I'm not sure if we have a specific guideline for naming of TestCases - but in my experience Given-When-Then approach was quite good in making the intent of the test clear. So, for example in this situation - GivenManifestAndContent_WhenSaving_ThenNoExceptions(), this might not be the best name for the test, I'm more giving an example. Also, if the test was named in this fashion, it would be much easier to think of what needs to be asserted, and what doesn't. For example, it makes no sense to assert empty manifest string (as I initially suggested) if it is already clear that manifest is "given". :)

PR #414, opened 2025-05-22
Make ServiceFabricEventSource event size assert more informative

olegsych:
Move the event size assert from the ServiceFabricEventSource to the EventDataArrayBuilder to include actual sizes

Also, make the the modified files easier to read:
    - Remove comments that don't add new information
    - Remove excessive line breaks
    - Move misplaced directives


src/Microsoft.ServiceFabric.Diagnostics/Tracing/Writer/EventDataArrayBuilder.cs
```diff
@@ -37,45 +38,21 @@ internal void AddEventData(Variant variant)
...
             return this.eventData;
         }
 
-        internal bool IsValid()
+        internal bool Validate(int eventId)
         {
-            return this.totalEventSize <= EventDataArrayBuilder.TraceEventMaximumSize;
+            bool valid = totalEventSize <= TraceEventMaximumSize;
+            if (!valid)
+                Debug.Fail($"Event {eventId} data size {totalEventSize} exceeds the maximum {TraceEventMaximumSize}.");
```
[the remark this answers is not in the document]
olegsych: Changed .editorconfig to stop recommending braces for single-line blocks.

.editorconfig
```diff
@@ -131,7 +131,7 @@ csharp_style_var_when_type_is_apparent = true:suggestion
 csharp_style_var_elsewhere = true:suggestion
 
 # Expression-level preferences
-csharp_prefer_braces = true:suggestion
```
ilijagrbic, line 134: This could just be my personal preference - but I like always having braces. I don't see any disadvantages to always using them, except perhaps for the file length. But it's a minor thing really, It's fine either way.

src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs
```diff
@@ -301,19 +217,9 @@ private string GetTypeFromIndex(int index, Variant v0, Variant v1, Variant v2, V
...
-        Variant v1 = default(Variant),
-        Variant v2 = default(Variant),
-        Variant v3 = default(Variant),
-        Variant v4 = default(Variant),
-        Variant v5 = default(Variant),
-        Variant v6 = default(Variant),
-        Variant v7 = default(Variant),
-        Variant v8 = default(Variant))
+        unsafe void VariantWrite(int eventId, int argCount, Variant v0 = default, Variant v1 = default, Variant v2 = default, Variant v3 = default, Variant v4 = default, Variant v5 = default, Variant v6 = default, Variant v7 = default, Variant v8 = default)
         {
```
ilijagrbic, line 223: I think we should do the same for the Linux specific method above: **VariantWriteViaNative(...)**. I expect us to keep it and build with Linux, so it makes sense keep the coding style consistent.

ilijagrbic (approved):

.editorconfig
olegsych: I'll add a comment explaining my reasoning. My personal preference for braces has changed. I started programming professionally with DBase which didn't support single line code blocks and required using End X keywords, so I was fine with always using braces in C-style languages with the K&R [indentation style](https://en.wikipedia.org/wiki/Indentation_style). However, the Allman style, so prevalent at Microsoft, is the least logical of the [possible styles](https://learning.oreilly.com/library/view/code-complete-2nd/0735619670/ch31s03.html). When it is combined with the "always use braces" rule, I think it desensitizes people to code size and complexity. During the 90s it became popular to criticize the old C code for being cryptic because some influential people in the 70s and 80s valued and praised the ability to write super compact code. I think the adoption of the Allman style with braces required was our performative way of making the code _look_ less cryptic. A page of dense K&R code looks obviously complex, but a page of Allman with braces looks OK. In reality, any function longer than a dozen lines and more than one level of nested braces is already too complex. I feel that discouraging braces around single-line blocks helps to make this more evident.

src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs
olegsych: Good idea.

olegsych (commented):
Implemented @ilijagrbic's suggestions.

ilijagrbic (approved):
Looks good, thanks @olegsych !

PR #416, opened 2025-05-29
Build projects with net8

ilijagrbic:
**Motivation**

This PR introduces changes that are needed for multiple reasons. 
Firstly, we will migrate away from netstandard2.0->net8.0 during the effort to support Mariner for SF. So, we are making similar migration here.

**Other changes**

- Build all of the projects in this repo with **net8.0** (instead of **netstandard2.0**)
- Clean up property files
- Other minor cleanup such as removing unnecessary properties, moving common properties to common files etc.

### DIFF
--- a/test/unittests/Microsoft.ServiceFabric.TestFramework/Microsoft.ServiceFabric.TestFramework.csproj
+++ b/test/unittests/Microsoft.ServiceFabric.TestFramework/Microsoft.ServiceFabric.TestFramework.csproj
@@ -1,13 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
- <Import Condition="'$(TargetFramework)' == 'net9.0'" Project="..\..\..\properties\service_fabric_managed_netstandard.props" />
- <Import Condition="'$(TargetFramework)' == 'net462'" Project="..\..\..\properties\service_fabric_managed_netframework.props" />
   <PropertyGroup>
     <ProjectGuid>{B8F12A3E-4C5D-4F6A-8B7C-9D1E2F3A4B5C}</ProjectGuid>
     <RootNamespace>Microsoft.ServiceFabric</RootNamespace>
     <TargetFrameworks>net9.0;net462</TargetFrameworks>
-    <AppendTargetFrameworkToOutputPath>true</AppendTargetFrameworkToOutputPath>
   </PropertyGroup>
+  <Import Project="..\..\..\properties\service_fabric_managed_common.props" />
   <ItemGroup>
     <PackageReference Include="Fuzzy" />
     <PackageReference Include="Microsoft.ServiceFabric" />


properties/service_fabric_managed_netcore_windows.props
```diff
@@ -2,7 +2,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="service_fabric_managed_common.props" />
   <PropertyGroup>
-    <!-- Define Constants used in code for set standard. -->
     <DefineConstants>$(DefineConstants);DotNetCoreClr</DefineConstants>
+    <NoWarn>$(NoWarn);SYSLIB0011;SYSLIB0050;SYSLIB0037;SYSLIB0012</NoWarn>
```
ilijagrbic, line 6: This is for multiple warnings we have regarding BinaryFormater (which I would preffer not to remove in this PR) and AssemblyName/Assembly (which we could consider removin in this PR).

properties/service_fabric_managed_common.props
```diff
@@ -25,11 +20,6 @@
   <PropertyGroup>
     <HighEntropyVA>true</HighEntropyVA>
   </PropertyGroup>
-
-  <!-- Set AllowUnsafeBlocks to false -->
-  <PropertyGroup>
-    <AllowUnsafeBlocks>false</AllowUnsafeBlocks>
-  </PropertyGroup>
```
ilijagrbic, line 32: AssemblyClsCompliant should be false since we no longer support it. If it's not present default is false. AllowUnsafeBlocks is false by default so removing it.

src/FabActUtil/FabActUtil.csproj
```diff
@@ -1,22 +1,18 @@
...
   <PropertyGroup>
     <ProjectGuid>{F6E091C3-9136-4058-91CF-57CDF383DF74}</ProjectGuid>
-    <RootNamespace>$(AssemblyName)</RootNamespace>
-    <AssemblyTitle>$(AssemblyName)</AssemblyTitle>
-    <DocumentationFile>$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
-    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
+    <TargetFrameworks>net8.0;net462</TargetFrameworks>
     <OutputType>Exe</OutputType>
   </PropertyGroup>
+  <Import Project="..\..\properties\service_fabric_managed.props" />
```
ilijagrbic, line 8: I moved the import after TargetFramework property, since otherwise it would not be defined in the props file that is imported. This would result in conditional Imports in the props file to not be resolved correctly.

properties/service_fabric_managed_netcore.targets
```diff
@@ -0,0 +1,34 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">  <!-- Copy prod binaries for net standard-->
+  <Target Name="CopyProdBinaries_netcore">
+    <!-- Publish FabActUtil so that all binaries from referenced nuget package are available for creation of nuget package-->
+    <!-- FabActUtil is invoked in VS tooling to autogenerate and update manifests for Actors. -->
+    <Exec Command="dotnet publish &quot;$(RepoRoot)src\FabActUtil\FabActUtil.csproj&quot; -f &quot;net8.0&quot; -o &quot;$(OutputPath)Publish&quot; --no-restore --nologo" />
+
+    <ItemGroup>
+      <!-- Windows binaries (net8.0-windows) -->
+      <BinairesNetStandard_Win Include="$(RepoRoot)src\Microsoft.ServiceFabric.Diagnostics\$(OutputPath)\net8.0-windows\Microsoft.ServiceFabric.Diagnostics.*"/>
```
olegsych, line 10: Looks like we're still using the term NetStandard here

properties/service_fabric_managed_netcore.targets
```diff
@@ -0,0 +1,34 @@
...
+  <Target Name="CopyProdBinaries_netcore">
+    <!-- Publish FabActUtil so that all binaries from referenced nuget package are available for creation of nuget package-->
+    <!-- FabActUtil is invoked in VS tooling to autogenerate and update manifests for Actors. -->
+    <Exec Command="dotnet publish &quot;$(RepoRoot)src\FabActUtil\FabActUtil.csproj&quot; -f &quot;net8.0&quot; -o &quot;$(OutputPath)Publish&quot; --no-restore --nologo" />
+
+    <ItemGroup>
+      <!-- Windows binaries (net8.0-windows) -->
+      <BinairesNetStandard_Win Include="$(RepoRoot)src\Microsoft.ServiceFabric.Diagnostics\$(OutputPath)\net8.0-windows\Microsoft.ServiceFabric.Diagnostics.*"/>   
+      
+      <!-- Linux binaries (net8.0) -->
```
olegsych, line 12: Hmm.... This is confusing. We only have linux-specific binaries for the Diagnostics.dll, correct?

properties/service_fabric_managed_netcore_windows.props
olegsych: This is fine, but instead of NoWarn, let's use WarningsNotAsErrors to make them noticeable and fix soon. https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/errors-warnings#warningsaserrors-and-warningsnotaserrors

ilijagrbic: Good point, I will update this.

properties/service_fabric_managed_netcore.targets
ilijagrbic: That was the original goal, however having different build targets proved to be a bit confusing either way - and it seems like I haven't managed to clear it up. I guess I should have called this "Other platform/Universal binaries". 

Anyways, if I decided to only have Linux-specific binary (build with **net8.0** target) for Diagnostics, that would mean I should build the regular Windows library with **net8.0-windows** target. This raises another question, which target should I use to build the other projects? From my perspective it made sense to use **net8.0** for other projects, since they didn't need a distinction between Linux and Windows. Which is how I ended up in this situation.

Either way, I will remove these changes for now to limit the scope of this PR, as you suggested. I will open another PR for only Linux changes. I like the direction you gave to move away from the compile directives, so we can try to explore that.

src/Microsoft.ServiceFabric.Diagnostics/Microsoft.ServiceFabric.Diagnostics.csproj
```diff
@@ -1,12 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-  <Import Condition="'$(TargetFramework)' == 'netstandard2.0'" Project="..\..\properties\service_fabric_managed_netstandard.props" />
-  <Import Condition="'$(TargetFramework)' == 'net462'" Project="..\..\properties\service_fabric_managed_netframework.props" />
+  <PropertyGroup>
+      <TargetFrameworks>net8.0;net8.0-windows;net462</TargetFrameworks>
+  </PropertyGroup>
+  <Import Project="..\..\properties\service_fabric_managed.props" />
   <PropertyGroup>
```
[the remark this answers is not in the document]
ilijagrbic: I have fixed this.

properties/service_fabric_managed.props
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
ilijagrbic: I see your point. Honestly, I got caught up in trying to clean everything up from the csproj files, that I just moved the mess to prop files. I tried to consolidate everything now in **service_fabric_managed_common.props** (we already have this file). Hopefully it is better now. :)

properties/service_fabric_managed_common.props
```diff
@@ -11,9 +11,20 @@
     <Platforms>AnyCPU</Platforms>
   </PropertyGroup>
 
-  <!-- Cls Compliant -->
+  <!-- Assembly name, title and documentation -->
   <PropertyGroup>
-    <AssemblyClsCompliant>true</AssemblyClsCompliant>
+      <RootNamespace>$(AssemblyName)</RootNamespace>
```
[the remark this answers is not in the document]
ilijagrbic: Thanks for noticing this, I will fix it.

properties/service_fabric_managed_common.props
```diff
@@ -11,9 +11,20 @@
...
+    <AssemblyTitle>$(AssemblyName)</AssemblyTitle>
+    <DocumentationFile Condition="!$(MSBuildProjectDirectory.Contains('\test\unittests\'))">$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
+  </PropertyGroup>
+
+  <!-- Properties specific to certain TargetFrameworks -->
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0' or '$(TargetFramework)' == 'net9.0'"> 
+    <DefineConstants>$(DefineConstants);DotNetCoreClr</DefineConstants>
+    <WarningsNotAsErrors>$(WarningsNotAsErrors);SYSLIB0011;SYSLIB0050;SYSLIB0037;SYSLIB0012</WarningsNotAsErrors >
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net462'">
```
olegsych, line 26: Why is the OutputType conditional for net462 only?

olegsych (approved):

olegsych (approved):

olegsych (approved):

olegsych (approved):

gobradovic (approved):

PR #417, opened 2025-05-30
Make Actor and ServiceFramework event sources publish unstructured traces on Linux

olegsych:
- Introduce `UnstructuredTracePublisher` to encapsulate publishing of unstructured traces to the Service Fabric telemetry pipeline used on Linux. This class inherits from `EventListener` for integration with any `EventSource`.
- Make the `ActorFrameworkEventSource` and `ServiceFrameworkEventSource` use the `UnstructuredTracePublisher` on Linux. These classes couldn't be simply changed to inherit from the `ServiceFabricEventSource`, which also supports unstructured tracing, because its hard-coded implementation supports only 8 event parameters and because it doesn't support unstructured tracing for the `WriteEventCore` method used by these event sources with up to 12 parameters.
- Bump `SF.ActorsServices.Internal` package version to `9.4.1`


olegsych:
@ilijagrbic Unfortunately, I haven't gotten as far as I hoped today. I have created the ServiceFabricUnstructuredTracePublisher I described in yesterday's scrum and have integrated in into the ServiceFrameworkEventSource, I just need to verify that it registers the expected EventListener. Hopefully this gives you an idea of where the design is going.

I'll undo the changes in the Microsoft.ServiceFabric.Diagnostics and integrate the ServiceFabricUnstructuredTracePublisher into the ActorFrameworkEventSource tomorrow. If this design works, the ServiceFabricUnstructuredTracePublisher should probably be moved to a better permanent location, at least to Microsoft.ServiceFabric.Diagnostics, possibly to System.Fabric, but that's a bridge to cross another day.

olegsych:
@ilijagrbic I think this PR is ready. Could you review when you get a chance?

test/unittests/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/ActorFrameworkEventSourceTest.cs
```diff
@@ -1,20 +1,82 @@
...
+using Moq;
 using Xunit;
 using Xunit.Abstractions;
 
 namespace Microsoft.ServiceFabric.Actors.Diagnostics
 {
     public abstract class ActorFrameworkEventSourceTest
     {
-        readonly ActorFrameworkEventSource sut = new ActorFrameworkEventSource();
+#if NET // Remove #if once on net472+ where IsOSPlatform is available
```
ilijagrbic, line 15: Lets use `!NETFRAMEWORK`, since I think `NET` only applies to **.NET5+**, while our projects are being build on **NetStandard** in main and develop branches.

nuprojs/SF.ActorsServices.Internal/SF.ActorsServices.Internal.nuproj
```diff
@@ -8,7 +8,7 @@
       <Description>This package provides Service Fabric Actors and Services libraries for consumption by other Service Fabric repos.</Description>
       <Summary>This package provides Service Fabric Actors and Services libraries for consumption by other Service Fabric repos.</Summary>
       <Tags>ServiceFabric Microsoft Azure Fabric</Tags>
-      <TargetFrameworkVersion>v4.8.1</TargetFrameworkVersion>
+      <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
```
ilijagrbic, line 11: Why are we changing this?

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorFrameworkEventSource.cs
```diff
@@ -17,12 +19,20 @@ namespace Microsoft.ServiceFabric.Actors.Diagnostics
     [EventSource(Name = "Microsoft-ServiceFabric-Actors", LocalizationResources = "Microsoft.ServiceFabric.Actors.SR", Guid = "0e1ec353-9f02-55d7-fbb8-f3857458acbd")]
     internal sealed class ActorFrameworkEventSource : EventSource
     {
-        // Although the documentation claims it exists, EventKeywords.All does not
-        // seem to be defined in the assembly. So just use a constant with all possible
-        // bits set.
-        private const long AllKeywords = 0x7FFFFFFFFFFFFFFF;
+#if !NETFRAMEWORK // Remove #if once on net472+ where IsOSPlatform is available
+        static Func<OSPlatform, bool> isOSPlatform = RuntimeInformation.IsOSPlatform;
```
ilijagrbic, line 23: nit: isOSPlatform should start with an uppercase letter (Pascal case), since currently that is our style convention.

nuprojs/SF.ActorsServices.Internal/SF.ActorsServices.Internal.nuproj
olegsych: This wasn't working in my local environment because I didn't have the 4.8.1 framework targeting pack installed. The 4.6.2. targeting pack is already required because we build net462 assemblies, so it should be already present.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorFrameworkEventSource.cs
```diff
@@ -17,12 +19,20 @@ namespace Microsoft.ServiceFabric.Actors.Diagnostics
...
-        // Although the documentation claims it exists, EventKeywords.All does not
-        // seem to be defined in the assembly. So just use a constant with all possible
-        // bits set.
-        private const long AllKeywords = 0x7FFFFFFFFFFFFFFF;
+#if !NETFRAMEWORK // Remove #if once on net472+ where IsOSPlatform is available
+        static Func<OSPlatform, bool> isOSPlatform = RuntimeInformation.IsOSPlatform;
+
+        public ActorFrameworkEventSource()
+        {
+            if (isOSPlatform(OSPlatform.Linux))
```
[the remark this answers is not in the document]
olegsych: I agree. It doesn't meet the rule of three yet, but I also thought that moving this logic to the ServiceFabricEventSource base class would be the right solution.

https://blog.codinghorror.com/rule-of-three/

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorFrameworkEventSource.cs
olegsych: Let's circle back to it. I feel our current style has too much of the StyleCop-inspired rules that don't have good reasons behind them. This rule in particular is strange because camelCase is used for private instance fields, which I do find helpful to distinguish private implementation details from public APIs or dependencies.

ilijagrbic: We should consider removing the rule then. I also find it weird, but still noticeable since IDE highlight it. :)

PR #418, opened 2025-06-03
Add new System and Fabric known exceptions

gobradovic:
Add following exceptions classes to lists of System and Fabric known exceptions:

* `System.Threading.Tasks.TaskCanceledException`
* `System.Fabric.FabricInsufficientMaxLoadCapacityException`
* `System.Fabric.FabricSkipRestoreOperationException`
* `System.Fabric.FabricBackupNotFoundException`

Bump minor version (`12.3.0` -> `12.4.0`).

Several minor fixes (proper order of using directives and name spaces, several typos fixed).

### DIFF
--- a/properties/service_fabric_common.props
+++ b/properties/service_fabric_common.props
@@ -26,7 +26,7 @@
     <!-- Major Version for Microsoft.ServiceFabric.Diagnostics binary - Bump together with normal MajorVersion -->
     <!-- We migrated Diagnostics package from WindowsFabric repo and we need to keep the versioning consistent -->
     <MajorVersion Condition="'$(MSBuildProjectName)' == 'Microsoft.ServiceFabric.Diagnostics'">12</MajorVersion>
-    <MinorVersion>3</MinorVersion>
+    <MinorVersion>4</MinorVersion>
     <BuildVersion>0</BuildVersion>
     <Revision>0</Revision>
 
--- a/test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/ExceptionConvertors/FabricExceptionConvertorTest.cs
+++ b/test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/ExceptionConvertors/FabricExceptionConvertorTest.cs
@@ -72,6 +72,9 @@ namespace Microsoft.ServiceFabric.Services.Remoting.Tests.V2.ExceptionConvertors
             new FabricPeriodicBackupNotEnabledException("FabricPeriodicBackupNotEnabledException"),
             new FabricValidationException("FabricValidationException"),
             new FabricTransportCallbackNotFoundException("FabricTransportCallbackNotFoundException"),
+            new FabricBackupNotFoundException("FabricBackupNotFoundException"),
+            new FabricSkipRestoreOperationException("FabricSkipRestoreOperationException"),
+            new FabricInsufficientMaxLoadCapacityException("FabricInsufficientMaxLoadCapacityException"),
         };
 
         [Fact]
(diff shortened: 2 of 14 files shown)


properties/service_fabric_common.props
```diff
@@ -26,7 +26,7 @@
     <!-- Major Version for Microsoft.ServiceFabric.Diagnostics binary - Bump together with normal MajorVersion -->
     <!-- We migrated Diagnostics package from WindowsFabric repo and we need to keep the versioning consistent -->
     <MajorVersion Condition="'$(MSBuildProjectName)' == 'Microsoft.ServiceFabric.Diagnostics'">12</MajorVersion>
-    <MinorVersion>3</MinorVersion>
+    <MinorVersion>4</MinorVersion>
```
ilijagrbic, line 29: Why are we bumping minor version for this change? We haven't integrated 9.3.0 anywhere anyways, so it might be alright to leave this as is.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/ExceptionConvertors/FabricExceptionConvertorTest.cs
```diff
@@ -72,6 +72,9 @@ public class FabricExceptionConvertorTest
             new FabricPeriodicBackupNotEnabledException("FabricPeriodicBackupNotEnabledException"),
             new FabricValidationException("FabricValidationException"),
             new FabricTransportCallbackNotFoundException("FabricTransportCallbackNotFoundException"),
+            new FabricBackupNotFoundException("FabricBackupNotFoundException"),
+            new FabricSkipRestoreOperationException("FabricSkipRestoreOperationException"),
+            new FabricInsufficientMaxLoadCapacityException("FabricInsufficientMaxLoadCapacityException"),
```
olegsych, line 77: The testing approach looks like an afterthought. I doubt we have adequate coverage because I don't see the validation of the inner exceptions, error codes. We're comparing HResults but aren't setting them anywhere. We should put this under proper tests at some point.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/ExceptionConvertors/SystemExceptionConvertorTest.cs
```diff
@@ -95,6 +95,7 @@ public class SystemExceptionConvertorTest
             new AbandonedMutexException("AbandonedMutexException"),
             new SemaphoreFullException("SemaphoreFullException"),
             new SynchronizationLockException("SynchronizationLockException"),
+            new TaskCanceledException("TaskCanceledException"),
```
olegsych, line 98: Well, on the plus side, these tests actually have some parameter validation logic, but there are several anti-patterns. 
- Instead of loops, we can use xUnit Theory (parameterized test methods)
- Instead of ifs, we should split the tests into functions so that conditions aren't needed.

olegsych (approved):
Thanks @gobradovic. I appreciate the opportunistic cleanup. I don't think the testing approach we have inherited is adequate and would love to retrofit a solid replacement, but we should probably keep this out of the current scope.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/ExceptionConvertors/FabricExceptionConvertorTest.cs
gobradovic: I agree, we should follow up on this.

PR #419, opened 2025-06-03
Remove Linux compiler directives from Microsoft.ServiceFabric.Diagnostics

ilijagrbic:
**1. Motivation/Goals**

- We are removing Linux specific compiler conditions from Diagnostics project (**DotNetCoreClrLinux**)
- We will keep the same behavior but use Runtime-level checks to determine if Linux specific logic is needed
- We are not addressing other issues with Diagnostics in this PR

**2. Testing/Other**

- This PR adds some unit tests to test the changes - focusing on behavior around previous Linux specific compilation
- This PR is being built with netstandard2.0 still, however in the future we can use newer API for platform checks.


src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs
```diff
@@ -278,9 +233,67 @@ unsafe void VariantWrite(int eventId, int argCount, Variant v0 = default, Varian
...
 #endif
         }
+
+// Methods needed only on Linux, and are included only in NetCore build
+#if DotNetCoreClr
+        [NonEvent]
+        public void VariantWriteViaNative(int eventId, int argCount, Variant v0 = default, Variant v1 = default, Variant v2 = default, Variant v3 = default, Variant v4 = default, Variant v5 = default, Variant v6 = default, Variant v7 = default, Variant v8 = default)
+        {
+
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
ilijagrbic, line 251: I'm considering extracting this into a separate method or even a separate class so we can test it without actually running a Linux build.

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/TestServiceFabricEventSource.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                return;
+            }
+
+            var eventSource = new TestEventSource();
+
+            var eventDescriptorsField = typeof(ServiceFabricEventSource).GetField("eventDescriptors", BindingFlags.NonPublic | BindingFlags.Instance);
+            var eventDescriptors = (ReadOnlyDictionary<int, TraceEvent>)eventDescriptorsField.GetValue(eventSource);
```
ilijagrbic, line 48: There is probably a better way to do this - investigate further.

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/TestServiceFabricEventSource.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            Assert.True(eventDescriptors.ContainsKey(eventId));
+            var traceEvent = eventDescriptors[eventId];
+            Assert.Equal(expectedHasId, traceEvent.hasId);
+            Assert.Equal(expectedTypeFieldIndex, traceEvent.typeFieldIndex);
+        }
+#else
+        [Fact]
+        public void VariantWriteViaNative_NotAvailableOnNetFramework()
```
ilijagrbic, line 57: This test verifies that VariantWriteViaNative is present only in net8 since in net462 is windows-only.

src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs
```diff
@@ -157,11 +164,17 @@ private ReadOnlyDictionary<int, TraceEvent> GenerateEventDescriptors(Type eventS
             return new ReadOnlyDictionary<int, TraceEvent>(eventDescriptors);
         }
 
-#if DotNetCoreClrLinux
+#if DotNetCoreClr
 
         [NonEvent]
         public void VariantWriteViaNative(int eventId, int argCount, Variant v0 = default, Variant v1 = default, Variant v2 = default, Variant v3 = default, Variant v4 = default, Variant v5 = default, Variant v6 = default, Variant v7 = default, Variant v8 = default)
```
olegsych, line 170: I did a search in ADO and didn't find any references of VariantWriteViaNative except this class and it's copy&paste in the SFRP repo. 

https://dev.azure.com/msazure/One/_search?action=contents&text=VariantWriteViaNative&type=code&lp=code-Project&filters=ProjectFilters%7BOne%7DRepositoryFilters%7BSF-RP-Common*WindowsFabric%7D&pageSize=25&result=DefaultCollection/One/SF-RP-Common/GBmaster//src/Common/Tracing/ServiceFabricEventSource.cs

Since structured tracing was never implemented on Linux, I think we should either make this method private or, ideally, replace it with the UnstructuredTraceListener. That way this class could serve as a base class for the ActorFramework and ServiceFrameworkEventSource classes and the variant optimization could be tested completely independently of the linux support. 

We could fully-unit test the Linux support and leave the variant testing out of scope for now, since we would only be removing obsolete code.

What do you think?

olegsych (changes requested):
Thanks for taking a stab at this @ilijagrbic. I have a few suggestions.

src/Microsoft.ServiceFabric.Diagnostics/Tracing/Writer/TraceEvent.cs
```diff
@@ -50,7 +50,7 @@ internal TraceEvent(
             this.message = message;
             this.UpdateSinkEnabledStatus();
             configMgr.OnFilterUpdate += this.UpdateSinkEnabledStatus;
-#if DotNetCoreClrLinux
+#if DotNetCoreClr
```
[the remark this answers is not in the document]
ilijagrbic: Agreed, I will clean up this file.

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/TestServiceFabricEventSource.cs
```diff
@@ -0,0 +1,65 @@
...
+using Xunit;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Tests
+{    
+    public class TestServiceFabricEventSource
+    {
+#if DotNetCoreClr
+        private IPlatformInformation GetPlatformInformation(OSPlatform platform)
+        {
+            var mockPlatform = new Mock<IPlatformInformation>();
```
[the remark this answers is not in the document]
ilijagrbic: I agree this is better, however I sometimes prefer longer option, since it is usually more understandable for people not familiar with the API. Anyways, I will use the option you provided.

src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs
```diff
@@ -133,12 +137,15 @@ private ReadOnlyDictionary<int, TraceEvent> GenerateEventDescriptors(Type eventS
 
                 var hasId = false;
                 int typeFieldIndex = -1;
-#if DotNetCoreClrLinux
-                var methodParams = eventMethod.GetParameters().ToList();
-                if (methodParams.Any())
+#if DotNetCoreClr
```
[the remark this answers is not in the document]
ilijagrbic: We use them in VariantWriteViaNative [here ](https://github.com/microsoft/service-fabric-services-and-actors-dotnet/blob/cf6a9cea3d64a37511f240e9cb79358463f8e916/src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs#L163), where we have logic that uses these two fields.

src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs
ilijagrbic: Well firstly, I think at minimum we should make the `VariantWriteViaNative` method private, that's a good catch. 

I review-ed your PRs for unstructured tracing, and I like the approach for Linux tracing especially because it is way more testable than the current setup. I'm just not sure whether we should incorporate similar changes into `ServiceFabricEventSource` in this PR, or open a new PR later on, that would do that and refactor `ServiceFrameworkEventSource` and `ActorFrameworkEventSource` to inherit from `ServiceFabricEventSource`.

src/Microsoft.ServiceFabric.Diagnostics/Tracing/Writer/TraceEvent.cs
```diff
@@ -19,7 +19,7 @@ internal class TraceEvent
         private readonly TraceConfig configMgr;
         private readonly string message;
         private bool filterState;
-#if DotNetCoreClrLinux
+#if DotNetCoreClr
```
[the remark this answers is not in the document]
ilijagrbic: Considering how we are planning to refactor `ServiceFabricEventSource`, we should eventually remove this anyways.

olegsych (approved):

PR #421, opened 2025-06-04
Make Actor and ServiceFrameworkEventSource classes inherit from ServiceFabricEventSource

olegsych:
ServiceFabricEventSource traces are collected on Linux by the ServiceFabric telemetry infrastructure. The additional traces are needed to troubleshoot Azure Linux/Mariner test failures.

Also, increment SF.ActorsServices.Internal package version to 9.2.4.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,11 +4,13 @@ on:
   push:
     branches:
       - develop
+      - main
       - 'release_*'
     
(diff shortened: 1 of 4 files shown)


ilijagrbic (approved):
Just so that we don't forget, we need to include these changes in develop as well.

gobradovic (approved):

PR #423, opened 2025-06-06
Reintroduce CLS compliance

ilijagrbic:
**1. Motivation/Goals**

- Goal of this PR is strictly to revert changes where we removed CLS compliance in [this PR](https://github.com/microsoft/service-fabric-services-and-actors-dotnet/pull/408)
-  Intent is to return to the original state for project in this repo regarding CLS compliance, so we are unblocked and can proceed without waiting on approval to remove CLS compliance in WindowsFabric repo.
- This situation is temporary, and long-term goal is to comprehensively remove CLS compliance from this and WindowsFabric repo, so I don't want to invest too much time into making this as clean as possible

**2. Why no CLS AssemblyInfo in netstandard2.0?**
- It was not present in the original separate projects used to build netstandard2.0 assemblies
- We could investigate this further - there seem to be some differences in how CLS works on netstandard/netframework - however this is not the goal of this PR

### DIFF
--- a/src/FabActUtil/FabActUtil.csproj
+++ b/src/FabActUtil/FabActUtil.csproj
@@ -24,11 +24,8 @@
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.ServiceFabric.Actors\Microsoft.ServiceFabric.Actors.csproj" />
   </ItemGroup>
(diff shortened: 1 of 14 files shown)


olegsych:
Let's hold off on this. I got Sridhar's agreement to proceed with the CLS compliance removal in Windows Fabric

PR #426, opened 2025-06-10
User ServiceFabricEventSource for Actors/Services

ilijagrbic:
TODO


test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/ServiceFabricEventSourceTest.cs
```diff
@@ -7,13 +7,18 @@
...
+using Xunit.Abstractions;
+using System.Diagnostics.Eventing.Reader;
+using EventLevel = System.Diagnostics.Tracing.EventLevel;
 
 
 namespace Microsoft.ServiceFabric.Diagnostics.Tests
-{    
+{
+#if DotNetCoreClr
     public abstract class ServiceFabricEventSourceTest
```
ilijagrbic, line 20: Diff for this file is really messy since most of the tests changed.
I reused the tests we had for FrameworkEventSource classes for this general test of `ServiceEventSource`.

ilijagrbic:
@olegsych, I was thinking if we need additional testing for these changes and I decided not to test them manually (with a script test for example). My reasoning was that it would require quite some time to build the Linux SF build and set up everything, while we do have pretty good Unit test coverage over these components. If you feel like we should test it, let me know.

src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs
```diff
@@ -28,14 +28,21 @@ internal abstract class ServiceFabricEventSource : EventSource
...
         /// <summary>
         /// Constructor which populates the events descriptor data for all events defined
         /// </summary>
         protected ServiceFabricEventSource() : this(DefaultPackageName)
         {
+            if(IsLinuxPlatform())
+            {
+#if !NETFRAMEWORK
+                var publisher = new UnstructuredTracePublisher();
+                publisher.EnableEvents(this, EventLevel.Informational);
```
olegsych, line 43: Could you change the EventLevel to Verbose? I realized that we need verbose events for the ActorFramework and ServiceFramework event sources. 

Ideally, I think we would want to use our custom configuration system to determine the EventLevel and EventKeywords we pass to the EnableEvents call. It also takes a dictionary, but I don't think it's used by the EventSource at all, in other words, I don't see an ability in EventSource to filter individual events like our custom config allows, but we could still use the EventLevel and EventKeywords from our custom config.

src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs
```diff
@@ -28,14 +28,21 @@ internal abstract class ServiceFabricEventSource : EventSource
...
-        protected static Func<OSPlatform, bool> IsOSPlatform = RuntimeInformation.IsOSPlatform;
+        protected static Func<OSPlatform, bool> isOSPlatform = RuntimeInformation.IsOSPlatform;
 #endif
         
         /// <summary>
         /// Constructor which populates the events descriptor data for all events defined
         /// </summary>
         protected ServiceFabricEventSource() : this(DefaultPackageName)
         {
+            if(IsLinuxPlatform())
```
olegsych, line 39: Shouldn't we have this code in the parameterized constructor so that both constructors support Linux tracing?

olegsych (approved):
I'm sorry about the noise, @ilijagrbic. Just thought of two more things. I don't think any of the comments I've left so far are blocking. Feel free to choose what to address and what to postpone.

src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs
```diff
@@ -292,19 +228,12 @@ unsafe void VariantWrite(int eventId, int argCount, Variant v0 = default, Varian
...
-#if DotNetCoreClr
-
-            if (IsLinuxPlatform())
-            {
-                this.VariantWriteViaNative(eventId, 9, v0, v1, v2, v3, v4, v5, v6, v7, v8);
-            }
-#endif
         }
 
         private bool IsLinuxPlatform()
```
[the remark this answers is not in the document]
ilijagrbic: Yeah, I kind of lost myself in trying to make the code pretty, I will use the simplest option for this.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorFrameworkEventSource.cs
```diff
@@ -8,6 +8,7 @@
 using System.Fabric;
 using System.Runtime.InteropServices;
 using Microsoft.ServiceFabric.Services;
+using Microsoft.ServiceFabric.Diagnostics.Tracing;
```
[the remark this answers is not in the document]
ilijagrbic: I fixed this and I have also set up Visual Studio to automatically do this on save - not sure if you are using that, but it is pretty cool.

test/unittests/Microsoft.ServiceFabric.Services.Tests/Diagnostics/ServiceFrameworkEventSourceTest.cs
```diff
@@ -6,8 +6,9 @@
 using Moq;
 using Xunit;
 using Xunit.Abstractions;
+using Microsoft.ServiceFabric.Diagnostics.Tracing;
 
-namespace Microsoft.ServiceFabric.Services.Runtime
+namespace Microsoft.ServiceFabric.Services.Diagnostics
 {
     public abstract class ServiceFrameworkEventSourceTest
```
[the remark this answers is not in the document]
ilijagrbic: I'm not sure if I understood correctly. But this test already does what you suggest - inherits from the `SericeFabricEventSource` and test the logic further. We could improve than naming of the tests, I basically took them from the `Actor/ServiceFrameworkEventSource` tests.

src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs
ilijagrbic: Good catch. I think we don't call parametrized constructor anywhere outside the class, so it is interesting it is still `protected`. I will move the logic to the parametrized constructor, since constructor without parameters calls it.

src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs
ilijagrbic: I will change to Verbose. I would like to understand a bit better how out configuration system works before making the other change you suggested, even though it makes sense to me.

olegsych: Sounds good. I'm not 100% sold on the idea of settings-based tracing configuration, it still sounds counterproductive and overcomplicated to me.

I've learned something new about proper way to enable publishing of events and unit tests them. Take a look at my #431. EnableEvents should specify EventLevel.LogAlways, but the unit test should verify EventLevel.Verbose and EventKeywords.All. I misunderstood the EnableEvents and IsEventEnabled APIs when I implemented the original version of this code and tests. The tests in particular were incorrect.

test/unittests/Microsoft.ServiceFabric.Services.Tests/Diagnostics/ServiceFrameworkEventSourceTest.cs
```diff
@@ -6,8 +6,9 @@
 using Moq;
 using Xunit;
 using Xunit.Abstractions;
+using Microsoft.ServiceFabric.Diagnostics.Tracing;
 
-namespace Microsoft.ServiceFabric.Services.Runtime
+namespace Microsoft.ServiceFabric.Services.Diagnostics
```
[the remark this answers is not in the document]
olegsych: Thanks @ilijagrbic. As to when and where to draw the line and introduce a new namespace, the good rule of thumb is to do it when you know the new namespace will have enough types in it to make the code [simpler](https://martinfowler.com/bliki/BeckDesignRules.html), specifically more intent-revealing with fewer code elements. For application code this is easy - keep all types in a single folder/namespace until you start noticing potential grouping patterns that could serve as sub-namespace candidates. Once you have a high-confidence candidate with 5+ types, move them to the new namespace. For public APIs this is more difficult because namespace changes aren't backward compatible. That's why I try to keep all APIs internal until we know they are stable enough to publish. Alternatively, if possible keep DLLs/packages at version 0.* until they are stable enough because semver states that there are no compatibility guarantees for version 0. We're probably not in a position to use the latter approach :)

ilijagrbic:
Thanks @olegsych, I have addressed the remaining comments and turned on auto-merge. (after you approve again) :)

test/unittests/Microsoft.ServiceFabric.Services.Tests/ServiceFrameworkEventSourceTest.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.IO;
+using Microsoft.ServiceFabric.Diagnostics.Tracing;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.ServiceFabric.Services.Diagnostics
+{
+    public abstract class ServiceFrameworkEventSourceTest
+    {
+#if NET // Remove #if once on net472+ where IsOSPlatform is available
```
olegsych, line 11: Is the #if still needed?

test/unittests/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/ActorFrameworkEventSourceTest.cs
```diff
@@ -15,58 +11,12 @@ public abstract class ActorFrameworkEventSourceTest
 #if NET // Remove #if once on net472+ where IsOSPlatform is available
```
olegsych, line 11: Is the #if still needed?

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/ServiceFabricEventSourceTest.cs
```diff
@@ -50,54 +53,65 @@ public void Dispose()
...
+                var traceEvent = eventDescriptors[eventId];
+                Assert.Equal(name, traceEvent.EventName);
+                Assert.Equal(eventLevel, traceEvent.Level);
+                Assert.Equal(message, traceEvent.Message);
             }
         }
 
-#else
-        public sealed class NetFramework : ServiceFabricEventSourceTest
+        public sealed class Manifest : ServiceFabricEventSourceTest
```
olegsych, line 96: Do we need the manifest test for the base class? This seems unnecessary because there are no events in it, so we wouldn't need to generate a manifest for PerfView, etc.

olegsych (approved):
Looks good @ilijagrbic. Just noticed a couple of places where code could be removed.

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/ServiceFabricEventSourceTest.cs
ilijagrbic: I would like to keep it, since we do expect every event that inherits from this class to have a manifest.

gobradovic (approved):

PR #434, opened 2025-06-24
Put ActorFrameworkEventSource under test

olegsych:
- Create unit tests for all internally accessible methods of the `ActorFrameworkEventSource`.
- Make projects build on Linux.
- Disable parallel execution of the `Microsoft.ServiceFabric.Diagnostics.Tests` to prevent failures caused by parallel modifiction of the global state.

### DIFF
--- a/test/unittests/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/ActorFrameworkEventSourceTest.cs
+++ b/test/unittests/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/ActorFrameworkEventSourceTest.cs
@@ -1,32 +1,417 @@
+using System;
 using System.Diagnostics.Tracing;
+using System.Fabric;
 using System.IO;
+using System.Runtime.InteropServices;
+using Fuzzy;
+using Inspector;
+using Microsoft.ServiceFabric.Actors.Tests;
 using Microsoft.ServiceFabric.Diagnostics.Tracing;
 using Xunit;
 using Xunit.Abstractions;
 
+using ActorFrameworkKeywords = Microsoft.ServiceFabric.Actors.Diagnostics.ActorFrameworkEventSource.Keywords;
+
 namespace Microsoft.ServiceFabric.Actors.Diagnostics
 {
-    public abstract class ActorFrameworkEventSourceTest
+    public abstract class ActorFrameworkEventSourceTest : IDisposable
     {
-#if NET // Remove #if once on net472+ where IsOSPlatform is available
+        readonly ActorFrameworkEventSource sut;
+
+        // Test fixture
+        static readonly IFuzz fuzzy = new RandomFuzz(Environment.TickCount);
+
+        public ActorFrameworkEventSourceTest()
+        {
+            // Allow event enablement to work on instances created by the tests
+            ActorFrameworkEventSource.Writer.Dispose();
+
+            // Disable Linux detection in sut to allow tests to run without libFabricCommon.so
+            typeof(ServiceFabricEventSource).Field<Func<OSPlatform, bool>>().Set(_ => false);
+
+            sut = new ActorFrameworkEventSource();
+        }
+
+        public virtual void Dispose()
+        {
+            sut.Dispose();
+
+            // Restore original static state
+            typeof(ServiceFabricEventSource).Field<Func<OSPlatform, bool>>().Set(RuntimeInformation.IsOSPlatform);
+            typeof(ActorFrameworkEventSource).Property<ActorFrameworkEventSource>().Set(new ActorFrameworkEventSource());
+        }
+
+        class TestEventListener : EventListener { }
+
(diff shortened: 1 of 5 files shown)


test/unittests/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/ActorFrameworkEventSourceTest.cs
```diff
@@ -1,32 +1,417 @@
...
+
+        // Test fixture
+        static readonly IFuzz fuzzy = new RandomFuzz(Environment.TickCount);
+
+        public ActorFrameworkEventSourceTest()
+        {
+            // Allow event enablement to work on instances created by the tests
+            ActorFrameworkEventSource.Writer.Dispose();
+
+            // Disable Linux detection in sut to allow tests to run without libFabricCommon.so
```
plave0, line 29: Where does this test suit look for `libFabricCommon.so`?
If `ServiceFabricEventSource` uses `libFabricCommon.so` or `FabricCommon.dll`, does this mean that we need SF runtime installed to run this test suit?

plave0 (approved):

test/unittests/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/ActorFrameworkEventSourceTest.cs
```diff
@@ -1,32 +1,417 @@
...
+            typeof(ServiceFabricEventSource).Field<Func<OSPlatform, bool>>().Set(_ => false);
```
ilijagrbic, line 30: As far as I understand, this mock will disable the Linux-specific code path for tracing. Was the goal to only test Windows logic for this PR, since the description says otherwise?

ilijagrbic: Now after reviewing entire PR, I see that we are testing using TestEventListener in the other test class. :)

ilijagrbic (approved):

test/unittests/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/ActorFrameworkEventSourceTest.cs
olegsych: I'm not sure what you mean by the other test class. I tried to explain why this had to be done in the comment above. It's not that we don't want to test Linux logic, rather that we don't want to test on Linux with the libFabricCommon.so dependency. The library is not available during unit test runs and was breaking them when I was still trying to rule out the possibility of managed side being responsible for the missing verbose traces.

test/unittests/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/ActorFrameworkEventSourceTest.cs
olegsych: It was using the default search logic, although I don't know what it is on Linux offhand. Probably starting with the location of the test runner. We don't have the runtime's native binaries available in unit tests and we don't want to turn them into integration tests, so instead, we're breaking off the dependencies to run unit tests in isolation.

plave0: I understand the necessity to decouple the SUT and the unit tests from their dependency on the runtime. 

What confuses me is why are we doing this only on Linux, and not on Windows. 

My assumption was that if we mock the Linux detection to always return false, then the SUT will always work as if it was running on Windows, and thus it will use `FabricRuntime.dll` instead of `libFabricRuntime.so`. But since the tests on this PR are all passing, and we don't have runtime libraries installed in our unit test environment, I conclude that my assumption is wrong. 

My question is: How do we decouple form SF runtime libraries on Windows (aka `FabricRuntime.dll`) with the current changes?

olegsych: Oh, I think I see where the confusion may be coming from. I should have said in the comment that disabling Linux detection allows the unit tests to run without the `FabricCommon` library, regardless of whether it's a `.dll` on Windows or `lib*.so` on Linux. The native library is not available for unit tests on either platform. They happened to run correctly on Windows before because the `UnstructuredTracePublisher` was enabled by the `ServiceFabricEventSource` base class only on Linux. But on Linux they were always failing before and this change allowed them to run isolated from the `UnstructuredTracePublisher`.

plave0: I just now realize I was writing `FabricRuntime` instead of `FabricCommon`, but you understood my question. Thank you for updating the comments.

PR #435, opened 2025-06-26
Remove binary serialization

gobradovic:
Remove deprecated binary serialization.

Switch WCF `FaultContract` to use `RemoteException2` instead of `RemoteException`. Rewrite unit tests accordingly.

Refactor WCF Remoting classes to improve testability.

Add unit tests for exception (de)serialization and propagation within WCF Remoting.

Bump version: `9.4.2` -> `9.5.0`


src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Client/WcfServiceRemotingClient.cs
```diff
@@ -3,30 +3,37 @@
 // Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
 // ------------------------------------------------------------
 
+using System;
+using System.Collections.Generic;
+using System.Fabric;
+using System.ServiceModel;
+using System.Threading.Tasks;
+
```
olegsych: Nit: I don't think micro-managing the using blocks this is useful. The "Remove and Sort Usings" refactoring does a good job, we just need to make sure the option to place System namespaces first is enabled. This is the old default that changed a few years ago, but once you enable it VS/Code should remember it in your cloud settings.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Client/WcfServiceRemotingClient.cs
```diff
@@ -3,30 +3,37 @@
...
+        private ExceptionDeserializer exceptionDeserializer;
 
         public WcfServiceRemotingClient(
             WcfCommunicationClient<IServiceRemotingContract> wcfClient,
             ServiceRemotingMessageSerializersManager serializersManager)
         {
             this.serializersManager = serializersManager;
             this.WcfClient = wcfClient;
+
+            IEnumerable<IExceptionConvertor> clientConvertors = new IExceptionConvertor[]
```
olegsych: I think we need to inject the IEnumerable<IExceptionConverter> as a constructor parameter in the WCF Client and Listener classes. I think we should also inject it in the constructor of the respective WCF Factory classes, like we do for FabricTransport Factory classes. This way the WCF API will become consistent with the FabricTransport API from the exception serialization standpoint.

olegsych: By the time we're done, we'll have 4 different places that build a list of well-known converters, fulfilling the rule of 3, so I think this would be a good time to extract into a reusable class or method in the Remoting.dll and verify that the factories using it call into the common logic in unit tests.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
```diff
@@ -300,13 +345,27 @@ private class WcfRemotingService : IServiceRemotingContract
...
             {
                 this.messageHandler = messageHandler;
                 this.serializersManager = serializersManager;
                 this.requestContext = new WcfServiceRemotingRequestContext(this.serializersManager);
+
+                var runtimeConvertors = new List<IExceptionConvertor>();
+                if (exceptionConvertors != null)
+                {
+                    runtimeConvertors.AddRange(exceptionConvertors);
+                }
```
ilijagrbic: nit: Maybe we can consider initializing `exceptionConvertors` to an empty list in the `WcfServiceRemotingListener` constructor. This way we don't have a possible null reference we need to take care in the rest of the code. Perhaps there is a reason against this, but anyways, if you think this is not needed, feel free to skip it. :)

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/RemoteExceptionTest.cs
```diff
@@ -22,34 +25,76 @@ public class RemoteExceptionTest
         /// SerializableExceptionStream Test.
         /// </summary>
         [Fact]
-        public static void SerializableExceptionStreamTest()
+        public static async Task SerializableExceptionStreamTest()
```
ilijagrbic: I think a lot of the value for Unit tests is in having clear names for test cases. At first glance I couldn't figure out what was the difference between this and the next test case (since the major difference is in the fact that the second exception type is not supported - and I had to find and check the list to confirm my assumption).

Perhaps we can rename these methods to something like: `GivenSupportedExceptionType_WhenDeserialized_IsRecognized()`
`GivenUnsupportedExceptionType_WhenDeserialized_FallbacksToServiceException()`

src/Microsoft.ServiceFabric.Services.Remoting/V2/Client/ExceptionDeserializer.cs
```diff
@@ -104,7 +104,33 @@ RemoteException2 DeserializeRemoteException2(byte[] buffer)
             }
         }
 
-        public async Task DeserializeRemoteExceptionAndThrowAsync(Stream stream)
+        internal Exception DeserializeRemoteException(RemoteException2 remoteException2)
```
ilijagrbic: nit: As far as I understand, this method doesn't deserialize the Exception, it rather converts it, so I find this naming a bit misleading. Perhaps we could rename it to `ConvertRemoteException`. Also, we could consider reusing it in `DeserializeRemoteExceptionAndThrowAsync()`, since these methods are identical apart from from deserialization first step in `DeserializeRemoteExceptionAndThrowAsync()`.

src/Microsoft.ServiceFabric.Actors/Remoting/V2/FabricTransport/Client/FabricTransportActorRemotingClientFactory.cs
```diff
@@ -93,19 +106,6 @@ private static IEnumerable<IExceptionConvertor> GetExceptionConvertors(
...
-            var handlers = new List<IExceptionHandler>();
-            if (exceptionHandlers != null)
-            {
-                handlers.AddRange(exceptionHandlers);
-            }
-
-            handlers.Add(new ActorRemotingExceptionHandler());
-            return handlers;
-        }
-
```
ilijagrbic, line 108: Is there a reason we moved this method up? Just curious, if this was perhaps done by editor or by you?

ilijagrbic (approved):
Great work @gobradovic , I left a couple of minor suggestions, but I don't see any major issues with this PR. :)

src/Microsoft.ServiceFabric.Actors/Remoting/V2/FabricTransport/Client/FabricTransportActorRemotingClientFactory.cs
gobradovic: It was done by me while I was attempting to make some additional changes which I reverted in the meantime. `GetExceptionHandlers` is used before `GetExceptionConvertors`, so I thought it makes _slightly_ more sense to be defined before it. No other reason besides that.

ilijagrbic: Got it, makes sense. Thanks for the explanation.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
gobradovic: Good suggestion for the name change (changed). As for the code reuse, I did notice the duplication, but as the different method call is in the middle of the method body, we'd have to pass the method reference. Additionally, duplicated code is both in try/catch blocks and outside it, which would make complicate reusing (not sure how readable it would be. I suggest we do a follow-up refactoring if you think we can do it elegantly.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Client/ExceptionDeserializer.cs
gobradovic: Good suggestion for the name change (changed). As for the code reuse, I did notice the duplication but I'm not sure if reuse would make the code significantly more unreadable. First, the different method call is in the middle of the method, so we'd probably have to pass in the method reference. Second, part of the duplicated code is in `try` and `catch` blocks, and part is below it, which might make reuse complicated. If you think we can do it elegantly let's follow up with a refactoring PR.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
gobradovic: This pattern is repeated in several other places. I wouldn't deviate from the pattern in only one place. I'd rather extract it to de-duplicate it and then change the logic there. I suggest doing this refactor ina  follow up, but if you think it makes more sense to do it here, I'm open to try.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/RemoteExceptionTest.cs
gobradovic: Great suggestion, changed.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Client/WcfServiceRemotingClient.cs
gobradovic: Injected IEnumerable to factory classes. For now, I haven't extracted the generation of list of well-known converters. We create two lists (one for runtime, one for client) in two different ways so although it is a similar procedure, it is not the same. Let me know if you think that we do want to extract it at this time.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
ilijagrbic: Yeah, I saw the pattern repeat multiple times when I went through the entire PR.
I don't think it is a big deal in this specific case, since I'm pretty sure we checked for null in all relevant places, and I agree on your approach how to tackle it (by extracting and in a follow up PR). 

Let's also wait to hear what Oleg thinks about this. :)

src/Microsoft.ServiceFabric.Services.Remoting/V2/Client/ExceptionDeserializer.cs
ilijagrbic: As we discussed, it doesn't seem obvious how to remove this duplication. Let's keep it as is for now, and if we think of something better, we can change it.

src/Microsoft.ServiceFabric.Actors/Remoting/V2/FabricTransport/Client/FabricTransportActorRemotingClientFactory.cs
```diff
@@ -79,6 +79,19 @@ public FabricTransportActorRemotingClientFactory(
         {
         }
 
+        private static IEnumerable<IExceptionHandler> GetExceptionHandlers(
+            IEnumerable<IExceptionHandler> exceptionHandlers)
+        {
+            var handlers = new List<IExceptionHandler>();
```
olegsych: consider using `new List<IExceptionHandler>(exceptionHandlers ?? Enumerable.Empty<IExceptionHandler>())`

src/Microsoft.ServiceFabric.Actors.Wcf/Remoting/V2/Wcf/Client/WcfActorRemotingClientFactory.cs
```diff
@@ -59,7 +61,66 @@ public WcfActorRemotingClientFactory(
...
+        [Obsolete]
+        public WcfActorRemotingClientFactory(
+            Binding clientBinding,
+            IServiceRemotingCallbackMessageHandler callbackClient,
+            IEnumerable<IExceptionHandler> exceptionHandlers,
+            IServicePartitionResolver servicePartitionResolver,
+            string traceId,
+            IServiceRemotingMessageSerializationProvider serializationProvider,
+            bool useWrappedMessage)
+            : base(
```
olegsych: can we call `this` instead of `base` to reduce duplication? 

Why are we passing `null` to the base class in this constructor rather than `GetExceptionConverters(null)`?

src/Microsoft.ServiceFabric.Actors.Wcf/Remoting/V2/Wcf/Client/WcfActorRemotingClientFactory.cs
```diff
@@ -78,11 +140,13 @@ public WcfActorRemotingClientFactory(
                 clientBinding,
                 callbackClient,
                 GetExceptionHandlers(exceptionHandlers),
+                GetExceptionConvertors(exceptionConvertors),
                 servicePartitionResolver,
                 traceId)
         {
         }
 
+
```
olegsych: Extra blank line?

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Client/WcfServiceRemotingClient.cs
```diff
@@ -3,30 +3,40 @@
...
     internal class WcfServiceRemotingClient : IServiceRemotingClient
     {
         private ServiceRemotingMessageSerializersManager serializersManager;
+        private ExceptionDeserializer exceptionDeserializer;
 
         public WcfServiceRemotingClient(
             WcfCommunicationClient<IServiceRemotingContract> wcfClient,
-            ServiceRemotingMessageSerializersManager serializersManager)
+            ServiceRemotingMessageSerializersManager serializersManager,
+            IEnumerable<IExceptionConvertor> exceptionConvertors)
```
olegsych: Since this class is internal and we don't have to maintain the constructor signature, it would be _simpler_ to inject an `ExceptionDeserializer` instance instead of injecting the exception converter and then creating it explicitly. I would also recommend encapsulating it in a new interface so that we can mock it for testing. Perhaps something like `IConvert<TSource, TTarget>` and making the `ExceptionDeserializer` implement `IConvert<RemoteException2, Exception>`

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Client/WcfServiceRemotingClient.cs
```diff
@@ -3,30 +3,40 @@
...
-    using Microsoft.ServiceFabric.Services.Communication;
-    using Microsoft.ServiceFabric.Services.Communication.Wcf.Client;
-    using Microsoft.ServiceFabric.Services.Remoting.V2;
-    using Microsoft.ServiceFabric.Services.Remoting.V2.Client;
-    using Microsoft.ServiceFabric.Services.Remoting.V2.Messaging;
-
     internal class WcfServiceRemotingClient : IServiceRemotingClient
     {
         private ServiceRemotingMessageSerializersManager serializersManager;
+        private ExceptionDeserializer exceptionDeserializer;
```
olegsych: Let's try making as many of the fields `readonly` as possible. `private` can be removed, it's an intuitive default every C# programmer should be familiar with.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Client/WcfServiceRemotingClientFactory.cs
```diff
@@ -31,6 +31,73 @@ public class WcfServiceRemotingClientFactory : IServiceRemotingClientFactory
...
+            Func<
+                Binding,
+                IEnumerable<IExceptionHandler>,
+                IServicePartitionResolver,
+                string,
+                IServiceRemotingCallbackContract,
+                WcfCommunicationClientFactory<IServiceRemotingContract>> createWcfClientFactory,
+            IServiceRemotingMessageSerializationProvider serializationProvider,
+            bool useWrappedMessage)
+        {
```
olegsych: Can we call the new/this constructor instead of duplicating this logic?

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
```diff
@@ -44,16 +45,57 @@ public class WcfServiceRemotingListener : IServiceRemotingListener
...
         /// <param name="useWrappedMessage">
         /// It indicates whether the remoting method parameters should be wrapped or not before sending it over the wire. When UseWrappedMessage is set to false, parameters  will not be wrapped. When this value is set to true, the parameters will be wrapped.Default value is false.</param>
+        [Obsolete]
+        public WcfServiceRemotingListener(
+            ServiceContext serviceContext,
+            IService serviceImplementation,
+            Binding listenerBinding,
+            IServiceRemotingMessageSerializationProvider serializationProvider,
+            string endpointResourceName,
+            bool useWrappedMessage) : this(serviceContext,
```
olegsych, line 62: Could you try making the indentation more readable? My general recommendation is to try placing all arguments on a single line by default and think about shortening the parameter names before resorting to placing every argument on a single line. This would allow maintaining the logical structure where this(...) is a function call. If we do need multiple lines, the indentation should maintain this logic.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
```diff
@@ -300,13 +345,27 @@ private class WcfRemotingService : IServiceRemotingContract
             // the actual callback channel is accessed from the current operation context.
             private readonly WcfServiceRemotingRequestContext requestContext;
 
+            private ExceptionSerializer exceptionSerializer;
+
             public WcfRemotingService(
                 IServiceRemotingMessageHandler messageHandler,
-                ServiceRemotingMessageSerializersManager serializersManager)
+                ServiceRemotingMessageSerializersManager serializersManager,
```
olegsych: Similar recommendations regarding encapsulation of the converters as for the client. On the server/listener side, the ExceptionSerializer could implement `IConvert<Exception, RemoteException2>`

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/Microsoft.ServiceFabric.Services.Wcf.Tests.csproj
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\..\properties\service_fabric_managed_netstandard.props" />
```
olegsych: This should be netframework.props, at least until #416 is merged, I think.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingClientTest.cs
```diff
@@ -0,0 +1,134 @@
...
+using Microsoft.ServiceFabric.Services.Remoting.V2.Runtime;
+using Microsoft.ServiceFabric.Services.Communication;
+using System.Threading.Tasks;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Wcf.Client
+{
+    public class WcfServiceRemotingClientTest
+    {
+        [Fact]
+        public async Task GivenSupportedExceptionType_WhenReceived_IsRecognized()
```
olegsych: Could you try using a hierarchical structure like [this](https://github.com/microsoft/service-fabric-services-and-actors-dotnet/blob/develop/test/unittests/Microsoft.ServiceFabric.Services.Tests/Runtime/ServiceFrameworkEventSourceTest.cs) based on public members instead of multi-part method names based on scenarios? For unit tests, we want to make sure that each type/member of the sut (class under test) is tested in isolation. I.e. constructor validates every argument, or generates a default value, etc. Unit tests should have clearly identifiable arrange, act and assert sections, either as single lines or separate paragraphs. Multi-paragraph sections need line comments to help reader understand what exactly is being tested. For example, does this test verify the constructor or the RequestResponseAsync method? Where does the setup/arrange logic ends and verification/begins?

Given/when/then test names work better for integration tests and scenarios verification. I find them problematic for long-term maintenance of the product code and tests. Without structure immediately obvious to everyone it's hard to reason about test coverage, i.e. we're changing members X, Y and Z, do tests 1 and 2 cover all changes? Later on, as the product code evolves, tests like this tend to be copy * pasted or become tightly-coupled.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingClientTest.cs
```diff
@@ -0,0 +1,134 @@
...
+                new CustomConvertorRuntime(),
+            };
+
+            ExceptionSerializer exceptionSerializer = new ExceptionSerializer(
+                runtimeExceptionConvertors,
+                null
+            );
+                
+            // Create RemoteException and FaultException
+            RemoteException2 customRemoteException = exceptionSerializer.BuildRemoteException(new CustomException("CustomEx", "CustomField1", "CustomField2"));
```
olegsych: Could you try using `fuzzy.String()` instead of hard-coded test values?

olegsych (changes requested):
Thanks @gobradovic. I've left a few suggestions about refining the product code and adding a bit more structure to the tests. Nothing major, feel free to push back. Let me know when you want me to make another pass.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
gobradovic: I like this idea, but we already digressed a bit with refactoring, and I'd prefer if we converged with this PR and do a follow up instead.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingClientTest.cs
```diff
@@ -0,0 +1,166 @@
...
+                // Assert that the unknown exception is converted to ServiceException during deserialization.
+                AggregateException exception = await Assert.ThrowsAsync<AggregateException>(() => sut.RequestResponseAsync(requestMessageMock));
+                Exception innerException = exception.Flatten().InnerException;
+                Assert.IsType<ServiceException>(innerException);
+                Assert.Equal(errorMessage, innerException.Message);
+            }
+        }
+    }
+
+    internal class CustomConvertorRuntime : ExceptionConvertorBase
```
olegsych: Why are these top-level classes defined in this file? Looks like they aren't needed in other test classes, so it would be better to make them nested inside of the main test/base class. They'll be accessible by default and won't need to be exposed to the rest of the project.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingClientTest.cs
```diff
@@ -0,0 +1,166 @@
...
+using System.Collections.Generic;
+using Microsoft.ServiceFabric.Services.Remoting.V2.Runtime;
+using Microsoft.ServiceFabric.Services.Communication;
+using System.Threading.Tasks;
+using Microsoft.ServiceFabric.Services.Remoting.V2.Client;
+using Fuzzy;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Wcf.Client
+{
+    public abstract class WcfServiceRemotingClientTestBase
```
olegsych: There is usually very little value in the term Base, unless you're implying https://youtu.be/rfMC2aVhYuo :). More importantly, the .cs file name should match the name of the class, so I would drop it.

olegsych (commented):
Thanks @gobradovic. I don't want to block the PR because, unfortunately, I can't spend much time on it now. My general feedback is to try unit testing every change. We expect one or more unit tests for each product change, so I would expect to see a lot more test changes in this PR.
For example, if we are adding a new constructor parameter, have a unit test that verifies that it worked as expected. Ideally, this means verifying that the constructor argument was used by a member of the class. For example, we could verify that the WcfServiceRemotingClientFactory.GetClientAsync returns an object created with the ExceptionSerializer that has the expected exception convertors. What's expected in this case? Default convertors when the constructor parameter was null or an empty collection. If the constructor parameter contained a custom convertor, the actual collection of convertors in the ExceptionSerializer should contain the custom and default converters in correct order.

It's fine to use Inspector (or Reflection APIs) to verify private fields, etc, in unit tests.

The test matrix grows very quickly, which is why I suggested treating ExceptionDeserializer.CreateSystemAndFabricExceptionDeserializer as an injected dependency, like we do in the event source classes where we also call static methods today. If you intercept this call, you can verify that an expected ExceptionDeserializer factory method is invoked, that the instance returned by this method is passed to the new WcfServiceRemotingClient, and then test the ExceptionDeserializer factory method separately.

If it helps we can try pair programming an example together.

src/Microsoft.ServiceFabric.Actors/Remoting/V2/FabricTransport/Client/FabricTransportActorRemotingClientFactory.cs
```diff
@@ -79,31 +80,26 @@ public FabricTransportActorRemotingClientFactory(
...
+            IEnumerable<IExceptionConvertor> exceptionConvertors)
         {
-            var handlers = new List<IExceptionHandler>();
-            if (exceptionHandlers != null)
+            var actorConvertors = new List<IExceptionConvertor>(exceptionConvertors ?? Enumerable.Empty<IExceptionConvertor>())
             {
-                handlers.AddRange(exceptionHandlers);
-            }
+                new FabricActorExceptionConvertor()
+            };
```
ilijagrbic, line 101: nit: I see that this (and similar) logic reoccurs multiple times. Perhaps we could extract this to avoid duplicated code?

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionDeserializerTest.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Collections.Generic;
+using Xunit;
+using Inspector;
+using Microsoft.ServiceFabric.Services.Remoting.V2.Client;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.Tests
+{
+    public abstract class ExceptionDeserializerTest
+    {
+        public class Constructor
```
ilijagrbic: I think Constructor and CreateSystemAndFabricExceptionDeserializer should inherit ExceptionDeserializerTest class for the proper hierarchical structure. Here is an example from [ServiceFabricEventSourceTest test](https://github.com/microsoft/service-fabric-services-and-actors-dotnet/blob/develop/test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/ServiceFabricEventSourceTest.cs).

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/RemoteExceptionTest.cs
```diff
@@ -7,49 +7,84 @@ namespace Microsoft.ServiceFabric.Services.Remoting.Tests.V2
...
-    /// </summary>
     public class RemoteExceptionTest
     {
-        /// <summary>
-        /// SerializableExceptionStream Test.
-        /// </summary>
         [Fact]
-        public static void SerializableExceptionStreamTest()
+        public static async Task GivenSupportedExceptionType_WhenDeserialized_IsRecognized()
         {
```
ilijagrbic: Since these tests basically test ExceptionDeserializer.DeserializeRemoteExceptionAndThrowAsync, perhaps we should move these tests to ExceptionDeserializerTest class into an appropriate hierarchical structure?

src/Microsoft.ServiceFabric.Services.Remoting/V2/Client/ExceptionDeserializer.cs
```diff
@@ -104,7 +117,33 @@ RemoteException2 DeserializeRemoteException2(byte[] buffer)
             }
         }
 
-        public async Task DeserializeRemoteExceptionAndThrowAsync(Stream stream)
+        internal Exception ConvertRemoteException(RemoteException2 remoteException2)
+        {
```
ilijagrbic, line 120: Since we have tests for DeserializeRemoteExceptionAndThrowAsync, I think we should include similar tests for this method as well? 

I understand that these methods have essentially the same logic, but since we are maintaining two different methods, we should maintain the tests for them as well.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingClientFactoryTest.cs
```diff
@@ -0,0 +1,97 @@
...
+using Xunit;
+using Microsoft.ServiceFabric.Services.Remoting.V2.Client;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Wcf.Client
+{
+    public abstract class WcfServiceRemotingClientFactoryTest
+    {
+        public class Constructor
+        {
+            public class WhenExceptionConvertorsArePassed
```
ilijagrbic: nit: I'm not sure if this extra level of hierarchy is really needed. Consider mering WhenExceptionConvertorsArePassed and WhenExceptionConvertorsAreNotPassed into 1 class while differentiating them in the TestCase name.

ilijagrbic: For example in WcfRemotingServiceTest you don't have extra hierarchy level, something similar to that could work. :)

src/Microsoft.ServiceFabric.Services.Remoting/V2/Runtime/ExceptionSerializer.cs
```diff
@@ -25,6 +26,19 @@ public ExceptionSerializer(IEnumerable<IExceptionConvertor> convertors, FabricTr
             this.settings = settings;
         }
 
+        public static ExceptionSerializer CreateSystemAndFabricExceptionSerializer(
+            IEnumerable<IExceptionConvertor> exceptionConvertors = null)
```
olegsych: I try to avoid default parameter values for internal APIs to reduce the risk of accidentally omitting them. Also, at the point of call `Enumerable.Empty<IExceptionConvertor>()` is more readable than `null` although it is significantly more verbose. So, if the increased verbosity is acceptable, expecting a non-null parameter value and throwing ArgumentNullException would be a faster way to fail. https://martinfowler.com/ieeeSoftware/failFast.pdf

src/Microsoft.ServiceFabric.Services.Remoting/V2/Runtime/ExceptionSerializer.cs
```diff
@@ -25,6 +26,19 @@ public ExceptionSerializer(IEnumerable<IExceptionConvertor> convertors, FabricTr
...
+        {
+            var convertors = new List<IExceptionConvertor>(exceptionConvertors ?? Enumerable.Empty<IExceptionConvertor>())
+            {
+                new SystemExceptionConvertor(),
+                new FabricExceptionConvertor(),
+                new DefaultExceptionConvertor()
+            };
+
+            return new ExceptionSerializer(convertors, null);
```
olegsych: It is suspicious that we're passing `null` as a `FabricTransportRemotingListenerSettings` here. Don't we need these settings in the `ToServiceException` method? It would be great to add ArgumentNullException logic to the constructor with respective unit tests.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Runtime/ExceptionSerializer.cs
```diff
@@ -25,6 +26,19 @@ public ExceptionSerializer(IEnumerable<IExceptionConvertor> convertors, FabricTr
             this.settings = settings;
         }
 
+        public static ExceptionSerializer CreateSystemAndFabricExceptionSerializer(
```
olegsych: Now that we're creating the `DefaultExceptionConvertor` as well, the method name became incomplete. Does it make sense to shorten it to something like `CreateDefault`?

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfRemotingService.cs
```diff
@@ -0,0 +1,102 @@
...
+using System.Threading.Tasks;
+using Microsoft.ServiceFabric.Services.Remoting.V2.Messaging;
+using Microsoft.ServiceFabric.Services.Remoting.V2.Runtime;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Wcf.Runtime
+{
+    [ServiceBehavior(ConcurrencyMode = ConcurrencyMode.Multiple)]
+    class WcfRemotingService : IServiceRemotingContract
+    {
+        private readonly IServiceRemotingMessageHandler messageHandler;
```
olegsych: Thanks for extracting this class into a separate file. Could you also clean up the formatting a bit
- private is redundant and could be removed.
- specify variable type instead of var unless it appears explicitly in the initialization statement
- avoid breaking lines unless they are difficult to read on our typical monitors, 120 characters still feels readable for me.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfRemotingService.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        private readonly ServiceRemotingMessageSerializersManager serializersManager;
+
+        // The request context need not be generated every time for WCF because for WCF,
+        // the actual callback channel is accessed from the current operation context.
+        private readonly WcfServiceRemotingRequestContext requestContext;
+
+        private ExceptionSerializer exceptionSerializer;
```
olegsych: Can we make the `exceptionSerializer` readonly?

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/Microsoft.ServiceFabric.Services.Wcf.Tests.csproj
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
```
olegsych: I thought VS project templates no longer created csproj files with the XML header. I don't think they are needed anymore, so I usually remove them.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
```diff
@@ -27,140 +26,302 @@ public class WcfServiceRemotingListener : IServiceRemotingListener
     {
         private IServiceRemotingMessageHandler messageHandler;
         private ICommunicationListener wcfListener;
+        private WcfRemotingService wcfRemotingService;
```
olegsych: Can `wcfRemotingService` be readonly? Would the field name make sense if we shorten it to `service`?

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
```diff
@@ -27,140 +26,302 @@ public class WcfServiceRemotingListener : IServiceRemotingListener
...
+        /// </param>
+        /// <param name="exceptionConvertors">
+        ///     Convertors to convert user exception to service exception.
+        /// </param>
         public WcfServiceRemotingListener(
             ServiceContext serviceContext,
             IService serviceImplementation,
             Binding listenerBinding = null,
             IServiceRemotingMessageSerializationProvider serializationProvider = null,
             string endpointResourceName = "ServiceEndpointV2",
```
olegsych: Does it make sense to make "ServiceEndpointV2" a constant?

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
```diff
@@ -27,140 +26,302 @@ public class WcfServiceRemotingListener : IServiceRemotingListener
...
-                address,
+                messageHandler,
                 serializersManager,
-                messageHandler);
+                address,
+                endpointResourceName);
         }
 
-        internal WcfServiceRemotingListener(
+        private void Initialize(
```
olegsych: It's weird that we have `Initialize` in addition to the internal constructor. It would be great to eliminate the `Initialize` method to reduce confusion. It might also allow us to make more fields readonly.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
```diff
@@ -229,51 +390,6 @@ private void DisposeIfNeeded()
             }
         }
 
-        private void Initialize(
```
olegsych, line 232: Thank you for simplifying and reducing duplication in this code.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionDeserializerTest.cs
olegsych: I think it's a good idea. This helps to make the test setup logic more granular and reusable without accidental sharing between tests, although it's not as relevant in this test class at the moment.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionDeserializerTest.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            [Fact]
+            public void UsesProvidedExceptionConvertors()
+            {
+                // Arrange
+                var customConvertors = new List<IExceptionConvertor> { new SystemExceptionConvertor() };
+                var deserializer = new ExceptionDeserializer(customConvertors);
+
+                // Assert
+                Assert.NotNull(deserializer);
```
olegsych: You could simplify the assertion block to `Assert.Same(customConverters, deserializer.Field<IEnumerable<IExceptionConvertor>>().Value)`.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionDeserializerTest.cs
```diff
@@ -0,0 +1,73 @@
...
+                IEnumerable<IExceptionConvertor> actualConvertors = deserializer.Field<IEnumerable<IExceptionConvertor>>().Value;
+                Assert.NotNull(actualConvertors);
+                var convertorList = new List<IExceptionConvertor>(actualConvertors);
+                Assert.Equal(2, convertorList.Count);
+                Assert.IsType<SystemExceptionConvertor>(convertorList[0]);
+                Assert.IsType<FabricExceptionConvertor>(convertorList[1]);
+            }
+
+            [Fact]
+            public void AppendsDefaultConvertorsToCustomList()
```
olegsych: I'm not sure if this test makes sense to keep. I don't think we could make it fail independently of the test above, which is a reliable sign of duplication. It also doesn't help to explain why the code works this way. If you think we should keep it, perhaps we could explain it in the test name or in the comments.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionDeserializerTest.cs
```diff
@@ -0,0 +1,73 @@
...
+            public void AddsDefaultConvertors()
+            {
+                // Act
+                var deserializer = ExceptionDeserializer.CreateSystemAndFabricExceptionDeserializer();
+
+                // Assert
+                Assert.NotNull(deserializer);
+                IEnumerable<IExceptionConvertor> actualConvertors = deserializer.Field<IEnumerable<IExceptionConvertor>>().Value;
+                Assert.NotNull(actualConvertors);
+                var convertorList = new List<IExceptionConvertor>(actualConvertors);
```
olegsych: Try using the `Enumerable` extension methods like `Count()` and `ElementAt()` to access `IEnumerable<>` without having to copy it in a new list.

Also, xUnit has good assertions for collections, which could be used in combination with LINQ APIs like this:

```C#
Type[] expected = new[] { typeof(SystemExceptionConvertor), typeof(FabricExceptionConvertor) };
Assert.Equal(expected, actualConvertors.Select(_ => _.GetType());
```

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionDeserializerTest.cs
```diff
@@ -0,0 +1,73 @@
...
+        public class CreateSystemAndFabricExceptionDeserializer
+        {
+            [Fact]
+            public void AddsDefaultConvertors()
+            {
+                // Act
+                var deserializer = ExceptionDeserializer.CreateSystemAndFabricExceptionDeserializer();
+
+                // Assert
+                Assert.NotNull(deserializer);
```
olegsych: I would usually avoid Assert.NotNull() when the test would throw a NullReferenceException anyway. With granular tests like this, it doesn't add much to the ability to understand the test or the failures.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionDeserializerTest.cs
olegsych: Oh, the test above passes null as the IEnumerable<IExceptionConverter> nevermind :)

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionSerializerTest.cs
```diff
@@ -0,0 +1,75 @@
...
+// ------------------------------------------------------------
+
+using System.Collections.Generic;
+using Xunit;
+using Inspector;
+using Microsoft.ServiceFabric.Services.Remoting.V2.Runtime;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.Tests
+{
+    public abstract class ExceptionSerializerTest
```
olegsych, line 20: We should also test the new `BuildRemoteException` method, particularly with the inner exceptions.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionDeserializerTest.cs
```diff
@@ -0,0 +1,73 @@
...
+// ------------------------------------------------------------
+
+using System.Collections.Generic;
+using Xunit;
+using Inspector;
+using Microsoft.ServiceFabric.Services.Remoting.V2.Client;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.Tests
+{
+    public abstract class ExceptionDeserializerTest
```
olegsych, line 20: We should also test the new `ConvertRemoteException` method

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/RemoteExceptionTest.cs
olegsych: I think it's a good idea. Besides we don't have a class called `RemoteException` anymore, correct?

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfRemotingServiceTest.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.Runtime.Remoting;
+using Microsoft.ServiceFabric.Services.Communication;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Wcf.Runtime.Tests
+{
+    public abstract class WcfRemotingServiceTest
+    {
+        // Text fixture
+        static readonly IFuzz fuzzy = new RandomFuzz(Environment.TickCount);
+        static readonly IServiceRemotingMessageHandler mockHandler = Mock.Of<IServiceRemotingMessageHandler>();
```
olegsych: fuzzy is stateless, and was specifically intended for static use to allowi initializing fields directly, without a constructor.

However, mocks are stateful; we should avoid any stateful instances in static, i.e. shared test fixtures, because they can make tests unreliable.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfRemotingServiceTest.cs
```diff
@@ -0,0 +1,127 @@
...
+
+        public class RequestResponseAsyncMethod
+        {
+            [Fact]
+            public async Task ReturnsActualExceptionTypeForKnownExceptions()
+            {
+                // Arrange
+                var errorMessage = fuzzy.String();
```
olegsych: Try extracting test fixture, like the `errorMessage`, that's common for all tests into fields and constructor of the class, `RequestResponseAsyncMethod` in this case. As long as it's straightforward initialization, without conditional logic, this can make the tests more compact and readable.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfRemotingServiceTest.cs
```diff
@@ -0,0 +1,127 @@
...
+                        new List<ArraySegment<byte>> { new ArraySegment<byte>(new byte[] { 4, 5, 6 }) }));
+
+                // Assert
+                IEnumerable<V2.Client.IExceptionConvertor> clientExceptionConvertors = new List<V2.Client.IExceptionConvertor>
+                {
+                    new V2.Client.SystemExceptionConvertor(),
+                };
+
+                // Create an ExceptionDeserializer to inspect the remote exception
+                var deserializer = new ExceptionDeserializer(clientExceptionConvertors);
```
olegsych, line 74: I try to avoid testing one class using implementation of another because this tends to tightly couple them and produce brittle tests. It's also often needs more test logic, which is I think evident here. Instead, I would change the test to verify that WcfRemotingService throws a FaultException<> where the RemoteException2 instance was created by the ExceptionSerializer.BuildRemoteException method. The method would need to be virtual, so that you could mock it and have it return a RemoteException2 instance. Then you could `Assert.Same(expected, faultException.Detail)` without having to duplicate the ExceptionSerializer tests.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingClientFactoryTest.cs
olegsych: +1. I try to read classes and methods collapsed to definitions to see if they explain what the tests do. 

In this case it reads like Constructor.WhenExceptionConvertorsArePassed.UsesPassedExceptionConvertors. Without the extra level, it could be something like Constructor.CreatesExceptionDeserializerWithGivenExceptionConvertors.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingClientTest.cs
```diff
@@ -0,0 +1,166 @@
...
+                clientExceptionConvertors);
+
+            sut = new WcfServiceRemotingClient(
+                    null, // wcfClient,
+                    null, // serializersManager
+                    exceptionDeserializer
+                );
+        }
+
+        public sealed class WellKnownException : WcfServiceRemotingClientTest
```
olegsych: Try reading collapsed class and method names to check if they explain how the system under test works and why. I.e. WcfServiceRemotingClientTest.WellKnownException.RequestResponseAsync.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingClientTest.cs
```diff
@@ -0,0 +1,166 @@
...
+                };
+
+                ExceptionSerializer exceptionSerializer = new ExceptionSerializer(
+                    runtimeExceptionConvertors,
+                    null
+                );
+
+                // Create RemoteException and FaultException
+                errorMessage = fuzzy.String();
+                RemoteException2 systemRemoteException = exceptionSerializer.BuildRemoteException(new NotImplementedException(errorMessage));
```
olegsych, line 66: If you make the BuildRemoteException virtual/mockable, you should be able to reduce these tests to a single test verifying that when WcfClient.Channel.RequestResponseAsync() throws a FaultException<>, the RequestResponseAsync throws the exception created by the ExceptionDeserializer.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingListenerTest.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Collections.ObjectModel;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Wcf.Runtime.Tests
+{
+    public abstract class WcfServiceRemotingListenerTest
+    {
+        public class Constructor
+        {
+            [Fact]
+            public void UsesPassedExceptionConvertors()
```
olegsych: It would be great to make the test name more specific - i.e. CreatesWcfRemotingServiceWithGivenExceptionConvertors.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingListenerTest.cs
```diff
@@ -0,0 +1,123 @@
...
+                var convertorList = new List<IExceptionConvertor>(actualConvertors);
+                Assert.Equal(4, convertorList.Count);
+                Assert.IsType<SystemExceptionConvertor>(convertorList[0]); // passed in
+                Assert.IsType<SystemExceptionConvertor>(convertorList[1]); // default
+                Assert.IsType<FabricExceptionConvertor>(convertorList[2]); // default
+                Assert.IsType<DefaultExceptionConvertor>(convertorList[3]); // default
+            }
+
+            [Fact]
+            public void UsesDefaultConvertorsWhenNotPassedAny()
```
olegsych: You could collapse these two tests into a single test verifying that ExceptionSerializer.CreateXxx is invoked with given convertors and that the instance it returns is passed to the WcfRemotingService instance. That way there'd be no duplication of the ExceptionSerializer tests here.

olegsych (approved):
Thanks @gobradovic. I've added a lot of suggestions on how the tests could be optimized. Feel free to pick what you want to implement; I don't expect us to actively maintain this code, so the main value of this exercise would be practicing different testing approaches. 

On thing I think we need to address is the null settings in the ExceptionSerializer, which could result in NullReferenceException for BuildRemoteException calls where exception has inner exceptions.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/RemoteExceptionTest.cs
ilijagrbic: Yeah, that's correct, this PR removed the class so no reason to keep the test class itself around anymore.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
ilijagrbic: I think this is a good idea, especially since this constant keeps reoccurring in multiple constructors. If it was a one occurrence I would prefer to keep it hardcoded.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
gobradovic: We don't initialize it in constructor but in Initialize method, that's why it can't be readonly. As for the name change, service is quite overloaded in this project: there's also `IService serviceImplementation`, but also `serviceContext` for which it would be ambiguous to which service it refers (actual service or remoting service.). We can rename it to `remotingService` perhaps if you'd prefer it that way.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
ilijagrbic: It's even more confusing since there are actually 2 Initialize methods.
I suspect the Initialize was extracted initially since the there was repeated logic between different constructors. Ideally we could return the initialization logic to one of the constructors and then call that constructor from the other.

So for example, WcfServiceRemotingListener():187 and WcfServiceRemotingListener():203 basically do the same thing, the first one only instantiates ServiceRemotingMessageSerializersManager(...) first. Similar case is also with WcfServiceRemotingListener():148 and WcfServiceRemotingListener():218

I'm not sure if there is an elegant way to use "this(...)", instantiate ServiceRemotingMessageSerializersManager and call the second constructor from the first.

ilijagrbic (approved):

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfRemotingService.cs
ilijagrbic: This might not be as related to this specific PR, but is there a way we can easily define and reuse same style conventions?  I'm not aware of something like that that we all currently use. But line length, sorting of dependencies/usings, indentation etc. could probably all be automatically applied on save action (or highlighted in the editor that you are straying away from the established convention). I know there is a "Code clean up" setting in Visual Studio, but that's something I have set up according to my preferences, not a team-wide standard.

I'm asking this because I also often have similar issues in my PRs and it would be awesome to prevent them in the future.
If we don't have something like this, perhaps we could consider creating a way to standardize it and use the same tools?

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
olegsych: Yeah, `remotingService` is better because wcf is already implied. I think we talked about options to eliminate the Initialize method today. I think it can be done if we replace it with static functions used to create constructor parameters instead, but don't know how much code restructuring would be needed to avoid duplication. This is definitely not a blocker for WCF which I hope we can drop soon just like the .NET team did.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
olegsych: Oh wow. Let's be careful tackling this. This is where accidental breaking changes are possible and it would be great to have tests for all constructors before the refactoring to verify we didn't break anything.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfRemotingService.cs
olegsych: I think .editorconfig can help with most conventions now https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/code-style-rule-options. At some point I maintained one for my personal projects: https://github.com/olegsych/csharp.common/blob/master/.editorconfig. Unfortunately, VS has been expanding .editorconfig very fast and I've fallen behind when I stopped actively working on them. Hopefully we can gradually adjust the shared settings https://github.com/microsoft/service-fabric-services-and-actors-dotnet/blob/develop/.editorconfig and cleanup the code.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionDeserializerTest.cs
gobradovic: Refactored and removed.

src/Microsoft.ServiceFabric.Actors/Remoting/V2/FabricTransport/Client/FabricTransportActorRemotingClientFactory.cs
gobradovic: There are two sets of classes with the same name (on client and runtime sides) that are actually different, so they can't be unified. This code actually occurs 2 instead of apparent 4 times. It would make sense to unify it, but that would require further refactoring, and this PR has already grown a bit too large.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
gobradovic: Yes, removed Initialize method as discussed. Renamed field as well.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
gobradovic: Removed Initialize, using internal constructor now.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/RemoteExceptionTest.cs
gobradovic: Added new unit tests, removed this test classes as obsolete.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingListenerTest.cs
gobradovic: Collapsed to single test. With current implementation it is hard to verify invocation of ExceptionSerializer.CreateDefault as it is called from constructor, so it is difficult to intercept it.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionDeserializerTest.cs
```diff
@@ -0,0 +1,181 @@
...
+
+                // Act & Assert
+                AggregateException exception = await Assert.ThrowsAsync<AggregateException>(async () =>
+                    {
+                        await exceptionDeserializer.DeserializeRemoteExceptionAndThrowAsync(stream);
+                    });
+
+                Exception innerException = exception.Flatten().InnerException;
+                Assert.IsType<FabricInsufficientMaxLoadCapacityException>(innerException);
+                Assert.Equal(originalException.Message, innerException.Message);
```
ilijagrbic, line 149: I think we can just use `Assert.Throws<FabricInsufficientMaxLoadCapacityException>` here.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionDeserializerTest.cs
```diff
@@ -0,0 +1,181 @@
...
+
+                // Act & Assert
+                AggregateException exception = await Assert.ThrowsAsync<AggregateException>(async () =>
+                    {
+                        await exceptionDeserializer.DeserializeRemoteExceptionAndThrowAsync(stream);
+                    });
+
+                Exception innerException = exception.Flatten().InnerException;
+                Assert.IsType<ServiceException>(innerException);
+                Assert.Equal(originalException.Message, innerException.Message);
```
ilijagrbic, line 172: I think we can just use `Assert.Throws<ServiceException>` here.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionDeserializerTest.cs
gobradovic: But FabricInsufficientMaxLoadCapacityException is not what is thrown here. AggregateException is thrown, and FabricInsufficientMaxLoadCapacityException is its InnerException. We also want to check the message of the inner exception. Maybe I didn't understand what you were getting at?

ilijagrbic: Ah, my bad, I totally missed this while glancing over, you are right, you can ignore the comment. :)

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
```diff
@@ -3,164 +3,334 @@
...
-
     /// <summary>
     /// An <see cref="IServiceRemotingListener"/> that uses
     /// Windows Communication Foundation to provide interface remoting for stateless and stateful services.
     /// </summary>
     public class WcfServiceRemotingListener : IServiceRemotingListener
     {
-        private IServiceRemotingMessageHandler messageHandler;
-        private ICommunicationListener wcfListener;
+        internal const string DefaultEndpointResourceName = "ServiceEndpointV2";
```
ilijagrbic: Why is this internal? We don't seem to use it anywhere outside of this class, so we should make it private.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
```diff
@@ -3,164 +3,334 @@
...
+                    GetDefaultSerializationProvider(serializationProvider, useWrappedMessage),
+                    new BasicDataContractHeaderSerializer(),
+                    useWrappedMessage),
                 listenerBinding,
-                endpointResourceName);
-        }
+                null,
+                endpointResourceName,
+                exceptionConvertors,
+                settings) { }
```
ilijagrbic: As we discussed @gobradovic , I think it is safer to revert the changes that removed `Initialize` methods since this doesn't appear to be a simple refactor. `ServiceRemotingMessageDispatcher` now doesn't use `MessageBodyFactory` from the original `ServiceRemotingMessageSerializersManager`, it rather constructs a new `ServiceRemotingMessageSerializersManager`, and if there are any issue with this, however unlikely, we won't be able to catch them with a unit test.

I would err on the safe side and avoid this change since the benefits here are not as significant.

ilijagrbic (approved):

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingClientTest.cs
gobradovic: Postponing for some future iteration.

ilijagrbic (approved):

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfRemotingServiceTest.cs
gobradovic: Leaving for a follow up refactor.

PR #438, opened 2025-06-28
Put ServiceFrameworkEventSource under test

olegsych:
- Extract `IFuzzExtensions` from the `Actors.Tests` to a new `TestFramework` project for reuse in `Services.Tests`
- Replace the unnecessary `TestEventListener` with partial mocks.
- Make GitHub copilot generate the `ServiceFrameworkEventSourceTest` based on the `ActorFrameworkEventSourceTest` with minor cleanup.
- Move the `ServiceFrameworkEventSourceTest` to the `Runtime` subnamespace/subfolder to match namespace/folder of the `ServiceFrameworkEventSource` and avoid unnecessary usings.


test/unittests/Microsoft.ServiceFabric.Services.Tests/Microsoft.ServiceFabric.Services.Tests.csproj
```diff
@@ -3,8 +3,7 @@
   <Import Project="..\..\..\properties\service_fabric_managed_netstandard.props" />
   <PropertyGroup>
     <ProjectGuid>{ECB3825D-B309-4C76-A20D-76544459F611}</ProjectGuid>
-    <AssemblyName>Microsoft.ServiceFabric.Services.Tests</AssemblyName>
-    <RootNamespace>$(AssemblyName)</RootNamespace>
+    <RootNamespace>Microsoft.ServiceFabric.Services</RootNamespace>
```
ilijagrbic, line 6: Why are we using `Microsoft.ServiceFabric.Services` instead of `Microsoft.ServiceFabric.Services.Tests` as root namespace for this project? Is this something we should change in all of the test projects?

olegsych: To reduce the number of using directives needed in the tests. Adding .Tests to every namespace in the test project breaks the namespace hierarchy. For example say you have a product namespace called Parent and a subnamespace called Parent.Child. If you're using .Tests suffix in your test namespaces, they will be like Parent.Tests and Parent.Tests.Child to be consistent with the hierarchy of the test project. So, the child tests would need to include a using Parent.Child directive.

You could call the test namespace Parent.Child.Tests to avoid this, but code analyzers have not been smart about recognizing or enforcing this, so I find it easier to place the tests in the same namespace as the product code they are testing.

ilijagrbic: Good to know, thanks for the explanation, Oleg! This seems like something we should adopt in other test projects at some point, for example when we start testing "Child" namespaces. :)

ilijagrbic (approved):

PR #441, opened 2025-07-16
Refactor usage of Performance Counters in ServicesRemoting and introduce abstractions needed for Telemetry

ilijagrbic:
**Motivation**
- We are not intending to change the behavior of the system in this PR.
- We want to refactor the way Performance Counters are emitted and to encapsulate it in an easily extendable abstraction, so that we can easily integrate telemetry in the future.
- In this PR we are trying to stray away from the existing pattern used for DiagnosticsManager in Actors project, but rather use Composite pattern.
- Use TDD for the development of this feature.

**Note**
- Before this PR is integrated in SF, we need to follow the process and publish new public NuGet packages, this is needed because we lacked InternalsVisibleTo for certain classes used in the ServicesRemoting projects. These classes were needed for testing, so update is necessary.


ilijagrbic:
Hey @olegsych, pinging you regarding this PR. We talked a bit during the scrum, and I have couple of questions you could help me with.

First of all, sorry if this is going to be a bit longer message, I will try to be as brief as I can.
We discussed the approach used in Actors repo for the DiagnosticsManager there, however we decided to use a different approach, relying on Composite pattern and trying to incorporate TDD in the development cycle.

My main questions are what you think about the direction of this PR from the design perspective? We mentioned the Composite pattern, I read about it, I think I have a good understanding of what I learned, but It will help me if you review the current design, just in case I have some blind spot or an obvious issue with what I want to do. Apart from the design, if you have any other suggestions more specific to the code, I'm always willing to listen, I just didn't want to bother you with low level details as I'm aware you are a bit overloaded right now. :)

There is a diagram illustrating what I'm trying to accomplish, but here is a short explanation, along with where we currently are in this PR:
- **[In Progress]** Currently I have `IDiagnosticsSource` interface, which defines methods currently needed to observe the `FabricTransportMessageHandler`. I'm planning to extend this interface with `Register(IDiagnosticsSource)` method, which would provide a way to register multiple sources - particularly in DiagnosticsManager (which implements this interface)
- **[In Progress]** `DiagnosticsManager` will manage multiple diagnostics sources (currently just PerfCounters, in future Telemetry as well), and expose this logic to `FabricTransportMessageHandler`. I have some of the test covering this integration.
- **[TODO]** I will implement `PerformanceCounterDiagnosticsSource`, which would have PerfCounter specific logic, I will basically move this logic from `FabricTransportMessageHandler`, where it is currently implemented. This class would essentially be a "Leaf" in the composite pattern, and `Register()` method would throw an exception.
- Continue working with TDD to make sure we have a high-quality bar for this.

src/Microsoft.ServiceFabric.Diagnostics/Util/SystemClock.cs
```diff
@@ -0,0 +1,15 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using Microsoft.ServiceFabric.Diagnostics.Util;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Diagnostic
```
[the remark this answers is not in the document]
olegsych: Good catch

src/Microsoft.ServiceFabric.Diagnostics/Util/IClock.cs
```diff
@@ -0,0 +1,14 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Util
```
olegsych: I think it's too early to create a new sub-namespace. A couple of reasons:
- The namespace structure should be optimized for API usability by people other than the original author. In practice, we usually do it the other way around, most of the time without considering trade-offs associated with namespace size and depth. The .NET Framework Design guidelines book has a good discussion of this: https://learning.oreilly.com/library/view/framework-design-guidelines/9780135896457/ch04.xhtml#sec4_1. 
- There aren't enough types to justify the added complexity of an extra namespace yet. We used to have an FxCop rule that generated a warning if a namespace had less than ~5 types. I don't think it was moved forward to the Roslyn Analyzers we use today, but I think it's still a good rule of thumb.
- The name `Util` doesn't convey additional information, which is a sign that we don't have a good way to classify the types in it. 

I would put these types in the parent namespace, Microsoft.ServiceFabric.Diagnostics. That's what I usually do, particularly with internal types, which can be moved at any time without breaking customers. So, keep accumulating them in a root namespace until the number of types grows large enough to spot emerging groupings that could have meaningful names.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/AgregatedDiagnosticEvents.cs
```diff
@@ -0,0 +1,107 @@
...
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Diagnostic
+{
+    internal class AgregatedDiagnosticEvents : IDiagnosticEvents, IDisposable
+    {
+        readonly static UniqueIDiagnosticEventsTypeComparer uniqueIDiagnosticEventsComparer = new UniqueIDiagnosticEventsTypeComparer();
+
+        readonly HashSet<IDiagnosticEvents> diagnosticsEventSet;
+
+        internal AgregatedDiagnosticEvents(IEnumerable<IDiagnosticEvents> diagnosticEvents)
```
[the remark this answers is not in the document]
olegsych: Good catch

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/PerformanceCounterDiagnosticEvents.cs
```diff
@@ -0,0 +1,84 @@
...
+        public void OnRequestResponseEnd(DateTime startTime)
+        {
+            if (performanceCounterProvider.ServiceOutstandingRequestsCounterWriter != null)
+            {
+                performanceCounterProvider.ServiceOutstandingRequestsCounterWriter.UpdateCounterValue(-1);
+            }
+            if (performanceCounterProvider.ServiceRequestProcessingTimeCounterWriter != null)
+            {
+                performanceCounterProvider.ServiceRequestProcessingTimeCounterWriter.UpdateCounterValue(
+                    CalculateMilisecondsSince(startTime));
```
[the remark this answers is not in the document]
olegsych: Good catch

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/PerformanceCounterDiagnosticEvents.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        public PerformanceCounterDiagnosticEvents(ServiceRemotingPerformanceCounterProvider performanceCounterProvider, IClock clock)
+        {
+            this.performanceCounterProvider = performanceCounterProvider ?? throw new ArgumentException(nameof(performanceCounterProvider));
+            this.clock = clock ?? throw new ArgumentException(nameof(clock));
+        }
+
+        public void OnRequestResponseBegin()
+        {
+            if(performanceCounterProvider.ServiceOutstandingRequestsCounterWriter != null)
```
olegsych: It looks like the ServiceRemotingPerformanceCounterProvider allows counter writers to be null. It looks like defensive programming, a well-known antipattern. We should fail-fast instead https://martinfowler.com/ieeeSoftware/failFast.pdf.

I'm not sure what to do about it here. Technically, we have to write two tests for each method - one for counter writer being null and the other for it being not null.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/PerformanceCounterDiagnosticEvents.cs
```diff
@@ -0,0 +1,84 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using Microsoft.ServiceFabric.Diagnostics.Util;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Diagnostic
+{
+    internal class PerformanceCounterDiagnosticEvents : IDiagnosticEvents, IDisposable
```
olegsych: I think it would be better not to implement IDisposable in classes that implement IDiagnosticEvents. The responsibility of disposing an object belongs to the object creating it. An object receiving a disposable shouldn't dispose it because it doesn't know what the lifetime of the disposable is supposed to be. There is a couple of unfortunate exceptions in the .NET Framework that weren't caught early enough and break this rule. For example StreamReader still disposes Stream object it doesn't own by default. However, we shouldn't repeat the same mistake. 

If there was something intrinsically disposable about diagnostic events, and implementing IDisposable was necessary, it would've been better to make IDiagnosticEvents inherit from IDisposable instead of implementing it as an optional internface.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/PerformanceCounterDiagnosticEvents.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        public void OnCreateTransportMessageBegin()
+        {
+            // Intentionally left blank, since we don't track transport message begin in performance counters.
+        }
+
+        public void OnCreateTransportMessageEnd(DateTime startTime)
+        {
+            if(performanceCounterProvider.ServiceRequestDeserializationTimeCounterWriter != null)
```
olegsych: I think our style expects a space between if and (

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/IDiagnosticEvents.cs
```diff
@@ -0,0 +1,19 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Diagnostic
+{
+    internal interface IDiagnosticEvents
```
olegsych: Outside of a class, all C# types are internal by default, so I usually omit the explicit access specifier.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/AgregatedDiagnosticEvents.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            if (diagnosticEvents == null || diagnosticEvents.Any(d => d == null)) 
+                throw new ArgumentException("Diagnostic events collection cannot be null or contain null elements.", nameof(diagnosticEvents)); 
+
+            this.diagnosticsEventSet = new HashSet<IDiagnosticEvents>(diagnosticEvents ?? throw new ArgumentNullException(nameof(diagnosticEvents)), uniqueIDiagnosticEventsComparer);
+
+            if (diagnosticEvents.Count() != this.diagnosticsEventSet.Count())
```
olegsych: I feel that checking for uniqueness of items is an overkill - additional allocation of memory for the HashSet, plus a custom comparer class. Ideally, we'd have tests verifying that an aggregate component was created with expected items, but even without it, the risk of duplication is usually low because the list of items is usually not dynamic. If you really want to keep this validation, the type of field we use to store the collection should be immutable, like IEnumerable<T> to reinforce that the class doesn't intend to change it, which is already implied by the readonly modifier.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/AgregatedDiagnosticEvents.cs
```diff
@@ -0,0 +1,107 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Diagnostic
+{
+    internal class AgregatedDiagnosticEvents : IDiagnosticEvents, IDisposable
+    {
+        readonly static UniqueIDiagnosticEventsTypeComparer uniqueIDiagnosticEventsComparer = new UniqueIDiagnosticEventsTypeComparer();
```
olegsych: Would `Comparer<T>.Default` work instead?

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/IDiagnosticEventsTest.cs
```diff
@@ -0,0 +1,47 @@
...
+using Xunit;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.Tests.V2.Diagnostic
+{
+    public abstract class IDiagnosticEventsTest
+    {
+        public class Signiture : IDiagnosticEventsTest
+        {
+            [Fact]
+            public void HasRequiredMethods()
```
olegsych: I usually avoid testing declarations explicitly because we already implicitly test them through the implementations, like the Aggregated* and PerformanceCounter* tests. How do you think this test would help someone understand and maintain the IDiagnosticEvents interface?

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/PerformanceCounterDiagnosticEventsTest.cs
```diff
@@ -0,0 +1,216 @@
...
+using Microsoft.ServiceFabric.Services.Remoting.V2.Diagnostic;
+using Moq;
+using Xunit;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.Tests.V2.Diagnostic
+{
+    public class PerformanceCounterDiagnosticEventsTest
+    {
+        readonly static ServiceRemotingPerformanceCounterProvider performanceCounterProvider = new ServiceRemotingPerformanceCounterProvider(Guid.NewGuid(), 0);
+        readonly static IClock mockClock = Mock.Of<IClock>();
```
olegsych: Try to avoid static fields in tests because they can easily lead to shared test fixtures which have significant downsides. https://learning.oreilly.com/library/view/xunit-test-patterns/9780131495050/ch18.html#ch18lev1sec8.

In this case, you could make them readonly instance fields and still initialize the sut field with a constructor. It takes a little more space, but not much more so.

```C#
readonly PerformanceCounterDiagnosticEvents sut;

protected PerformanceCounterDiagnosticEventsTest() =>
  sut = new PerformanceCounterDiagnosticEvents(performanceCounterProvider, mockClock);
```

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/FabricTransport/Runtime/FabricTransportMessageHandlerTest.cs
```diff
@@ -0,0 +1,186 @@
...
+using Moq;
+using Xunit;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.Tests.V2.FabricTransport.Runtime
+{
+    public abstract class FabricTransportMessageHandlerTest
+    {
+        readonly FabricTransportMessageHandler sut;
+
+        readonly IServiceRemotingMessageHandler mockRemotingMessageHandler = Mock.Of<IServiceRemotingMessageHandler>();
```
olegsych: I use mocks a lot and don't find the term `mock` useful in variable names. Most of the time you already know it's a mock, so it doesn't add much value to the variable name. Plus it creates duplication when you start using them in code like `Mock.Get(mockSomething)...`. Consider dropping it.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/FabricTransport/Runtime/FabricTransportMessageHandlerTest.cs
```diff
@@ -0,0 +1,186 @@
...
+
+            // Mock needed dependencies
+            Mock.Get(this.mockRemotingMessageHandler).Setup(m => m.HandleRequestResponseAsync(It.IsAny<IServiceRemotingRequestContext>(), It.IsAny<IServiceRemotingRequestMessage>()))
+                .Returns(Task.FromResult(Mock.Of<IServiceRemotingResponseMessage>()));
+
+            Mock.Get(this.mockSerializerManager).Setup(m => m.GetHeaderSerializer())
+                .Returns(mockHeaderSerializer);
+            Mock.Get(this.mockSerializerManager).Setup(m => m.GetRequestBodySerializer(It.IsAny<int>()))
+                .Returns(mockRequestMessageBodySerializer);
+            Mock.Get(this.mockSerializerManager).Setup(m => m.GetResponseBodySerializer(It.IsAny<int>()))
```
olegsych: Could you try creating new Mock<IServiceRemotingMessageSerializersManager >(){ DefaultValue = DefaultValue.Mock } to see if it recursively creates mocks for the results returned by the GetHeaderSerializer, GetRequestBodySerializer, etc. ?

This is excessive amount of code compared to what I would normally expect from NSubstitute for a similar interface. Moq API is more verbose but I hope it can able to handle this without too much overhead.

olegsych (changes requested):
Hey @ilijagrbic. I've completed the first pass, no real blockers, but a bunch of suggestions. I think design of the IDiagnosticEvents interface is the most important to nail down.

src/Microsoft.ServiceFabric.Diagnostics/Util/IClock.cs
ilijagrbic: I agree, we already had a discussion about this before. It just slipped my mind when I was moving these files to Diagnostics project. I will move them to the root namespace.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/PerformanceCounterDiagnosticEvents.cs
ilijagrbic: Yes, I noticed this and covered each method by two tests. It is not ideal, But I think we would require a more significant refactoring of the ServiceRemotingPerformanceCounterProvider class in order to evade this situation. Since currently we have a small number of methods, I feel this is not necessary.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/IDiagnosticEventsTest.cs
ilijagrbic: In the beginning I was following TDD way too strictly, so I avoided adding any code unless it would fix a test. I stopped doing that, but these tests were a symptom of it. I think you are right, especially since we do test the signature implicitly in other tests. Anyways I will remove this test. :)

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/FabricTransport/Runtime/FabricTransportMessageHandlerTest.cs
ilijagrbic: My reasoning was because the "Mock.Of(...)" API doesn't create Mock class types but rather the actual classes. This is normally not a problem if you only have mocks in the test, but in this specific test there was a mixture of mock and actual classes that were used to instantiate handler class. So having an easy way to differentiate it seemed useful to me.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/PerformanceCounterDiagnosticEventsTest.cs
ilijagrbic: This is a good point. I was trying to make the tests as short/simple as possible. But having them more maintainable and avoiding disadvantages associated with shared test fixture is definitely more important. I will fix this as you suggested.

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/FabricTransport/Runtime/FabricTransportMessageHandlerTest.cs
ilijagrbic: I will remove mock prefix in other test classes, where it is not needed, but I will keep it in the FabricTransportMessageHandlerTest.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/AgregatedDiagnosticEvents.cs
ilijagrbic: As far as I was aware, for a type to use `Comparer<T>`, it would need to implement `IComparer<T>` interface. Which is something I didn't want to do for `IDiagnosticEvents` interface because this logic was only needed in the `AgregatedDiagnosticEvents` class and I felt it would be better to have it as internal in this class and avoid exposing this.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/AgregatedDiagnosticEvents.cs
ilijagrbic: I agree with the take on `IEnumerable`, I thought that having `readonly` should be enough in that regard since I wanted to communicate that the collection is not supposed to change.

On the uniqueness part, I was honestly not sure if it would be an overkill. For me it seemed better to prevent creation of objects which behavior wouldn't be as intended by the design of the class. However, in practice, I do think likelihood of this happening is quite small. Anyways, to simplify everything, I will remove this validation. :)

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/PerformanceCounterDiagnosticEvents.cs
ilijagrbic: Thanks for this comment, I was also thinking about this and was unsure what to do. 

Initially I haven't thought `IDiagnosticEvents` needs to be disposable. However, as we discussed, we injected `ServiceRemotingPerformanceCounterProvider` (which needs to be disposed) in the `PerformanceCounterDiagnosticEvents`, which made that class depend on a disposable object. 

From what I understood, since `ServiceRemotingPerformanceCounterProvider` is instantiated in the `FabricTransportMessageHandler` that class should dispose it, regardless of the fact that the object is passed to `PerformanceCounterDiagnosticEvents` (and `AggregatedDiagnosticManager`)?
I case this is correct, I will keep a reference to `ServiceRemotingPerformanceCounterProvider` and dispose it in `FabricTransportMessageHandler`. In case you think this should be handled differently, we can discuss it further (I will keep the comment open). :)

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/FabricTransport/Runtime/FabricTransportMessageHandlerTest.cs
ilijagrbic: This is awesome, `DefaultValue.Mock` is working as intended, and now there is barely any code for initialization of mock fields. :)

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/PerformanceCounterDiagnosticEvents.cs
ilijagrbic: Hi @olegsych , just a quick ping regarding this. :)

olegsych: Yup, that's exactly what I meant.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/PerformanceCounterDiagnosticEvents.cs
```diff
@@ -0,0 +1,84 @@
...
+// ------------------------------------------------------------
+
+using System;
+using Microsoft.ServiceFabric.Diagnostics;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Diagnostic
+{
+    internal class PerformanceCounterDiagnosticEvents : IDiagnosticEvents, IDisposable
+    {
+        private ServiceRemotingPerformanceCounterProvider performanceCounterProvider;
```
olegsych: Let's try to make fields readonly as much as possible. They're all private by default, so I would not use this keyword at all. Readonly however is particularly important in multi-threaded and asynchronous code.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/PerformanceCounterDiagnosticEvents.cs
```diff
@@ -0,0 +1,84 @@
...
+        private IClock clock;
+
+        private long CalculateMillisecondsSince(DateTime startTime)
```
olegsych: Could you move private methods to the end of the class? This improves readability by keeping the more important details, like what constructors are available, at the top, where they are more easily discovered.

src/Microsoft.ServiceFabric.Services.Remoting/V2/FabricTransport/Runtime/FabricTransportMessageHandler.cs
```diff
@@ -19,16 +19,18 @@ sealed class FabricTransportMessageHandler : IFabricTransportMessageHandler
...
         private static readonly string TraceType = typeof(FabricTransportMessageHandler).Name;
         private readonly IServiceRemotingMessageHandler remotingMessageHandler;
-        private readonly ServiceRemotingMessageSerializersManager serializersManager;
+        private readonly IServiceRemotingMessageSerializersManager serializersManager;
         private readonly Guid partitionId;
         private readonly long replicaOrInstanceId;
-        private readonly ServiceRemotingPerformanceCounterProvider serviceRemotingPerformanceCounterProvider;
         private IServiceRemotingMessageHeaderSerializer headerSerializer;
         private ExceptionSerializer exceptionSerializer;
+        private IDiagnosticEvents diagnosticEvents;
```
olegsych: Can we make new fields readonly?

olegsych (approved):

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/PerformanceCounterDiagnosticEvents.cs
ilijagrbic: I agree, thanks for noticing and raising this.

src/Microsoft.ServiceFabric.Services.Remoting/V2/ServiceRemotingMessageSerializersManager.cs
```diff
@@ -1,94 +1,94 @@
...
-            if (!ServiceCodeBuilder.TryGetKnownTypes(interfaceId, out var interfaceDetails))
-            {
-                throw new ArgumentException("No interface found with this Id  " + interfaceId);
-            }
-
-            return interfaceDetails;
-        }
-    }
-}
+// ------------------------------------------------------------
```
olegsych: Could you double-check that we're not reverting from LF to CRLF line endings in the Git index?

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/AggregatedDiagnosticEvents.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            if (diagnosticEvents == null || diagnosticEvents.Any(d => d == null)) 
+                throw new ArgumentException("Diagnostic events collection cannot be null or contain null elements.", nameof(diagnosticEvents)); 
+
+            this.diagnosticEvents = diagnosticEvents;
+        }
+
+        public void OnRemotingRequestBegin()
+        {
+            foreach (var ds in diagnosticEvents)
```
olegsych: Use explicit type instead of var?

src/Microsoft.ServiceFabric.Services.Remoting/V2/FabricTransport/Runtime/FabricTransportMessageHandler.cs
```diff
@@ -1,204 +1,190 @@
...
-            {
-                this.serviceRemotingPerformanceCounterProvider.ServiceRequestDeserializationTimeCounterWriter.UpdateCounterValue(
-                    stopwatch.ElapsedMilliseconds);
-            }
-
-            return new ServiceRemotingRequestMessage(deSerializedHeader, deserializedMsg);
-        }
-    }
-}
+// ------------------------------------------------------------
```
olegsych: Could you double-check that we aren't reverting from LF to CRLF in Git index?

src/Microsoft.ServiceFabric.Services.Remoting/V2/FabricTransport/Runtime/FabricTransportServiceRemotingRequestContext.cs
```diff
@@ -1,45 +1,45 @@
...
-            if (this.callback == null)
-            {
-                throw new FabricTransportCallbackNotFoundException(string.Format(CultureInfo.CurrentCulture, SR.ErrorClientCallbackChannelNotFound, this.id));
-            }
-
-            return this.callback;
-        }
-    }
-}
+// ------------------------------------------------------------
```
olegsych: Could you double-check that we aren't reverting from LF to CRLF in Git index?

olegsych (approved):

src/Microsoft.ServiceFabric.Services.Remoting/V2/ServiceRemotingMessageSerializersManager.cs
ilijagrbic: Thank you for noticing these, I have renormalized the line endings again for these files.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/AggregatedDiagnosticEvents.cs
ilijagrbic: I thought the type was relatively obvious since the variable name is basically same as the type name (diagnosticEvents - `IDiagnosticEvents`), but there is no harm to use full type since these methods are already short.

olegsych (approved):

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/PerformanceCounterDiagnosticEvents.cs
olegsych: @ilijagrbic I don't think the plan we discussed is implemented yet.

ilijagrbic: Apologies @olegsych , a long time has passed since we initially talked about this, and I overlooked it. I will create a follow up PR for this. :)

PR #443, opened 2025-07-17
Fix Diagnostics.Tests to run for NetFramework

ilijagrbic:
During previous changes in Diagnostics project, we added several unit tests.
- Most of these tests focused on Linux/Windows logic relevant only on NET8 target
- UnstructuredTracePublisher only works on NET8 for Linux, so tests for it are also not relevant on net462 target
- This PR enabled some of the tests still relevant on net462 target such as
  - Manifest test for the ServiceFabricEventSourceTest
  - Constructor EventDescriptor test
- [MISC] Adding .nupkg files to gitgnore. We need to add WindowsFabric nugets into refs/ folder during local testing (with locally build WindowsFabric). These files would unnecessarily show up in difs


test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/ServiceFabricEventSourceTest.cs
```diff
@@ -75,6 +75,11 @@ public void DoesntEnableUnstructuredEventPublishingOnWindows()
                 Assert.False(sut.IsEnabled());
             }
 
+#else
+        public sealed class Constructor : ServiceFabricEventSourceTest
```
olegsych: I've never had to deal with NetFx/NetCore differences in tests before. I think this might be the first real-life example where my favorite hierarchical test class structure becomes less readable than the flat-list conventional alternative. 

It would be great to get rid of as much conditional compilation as possible. Unfortunately, we're limited by the unavailability of the RuntimeInformation.IsOSPlatform on net462. Hopefully, we can upgrade to net472 soon, but in the meantime, it would be great to avoid #ifs cutting across {} scopes. 

Can we do something like this?

```C#
#if NET
    partial public sealed class Constructor : ServiceFabricEventSourceTest
    {
         // NetCore tests
    }
#endif 

    partial public sealed class Constructor : ServiceFabricEventSourceTest
    {
         // Net-neutral tests
    }
```

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/ServiceFabricEventSourceTest.cs
```diff
@@ -15,9 +15,9 @@
 
 namespace Microsoft.ServiceFabric.Diagnostics.Tests
 {
-#if DotNetCoreClr
     public abstract class ServiceFabricEventSourceTest
     {
+#if DotNetCoreClr
```
olegsych: Let's use standard compiler symbols for all new code and opportunistically replace the legacy usage of our custom symbols. In this particular case, lets use `NET` instead of `DotNetCoreClr`. 

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/Microsoft.ServiceFabric.Diagnostics.Tests.csproj
```diff
@@ -7,11 +7,11 @@
   </PropertyGroup>
```
olegsych, line 6: Could you also remove the AllowUnsafeBlocks? I don't think we have any unsafe blocks in the test project.

olegsych (changes requested):
Thanks @ilijagrbic. I've added a couple of minor suggestions to improve readability. The main problem, however, is that the netfx tests still don't run in the PR validation build.

```
[xUnit.net 00:00:01.32] Skipping: Microsoft.ServiceFabric.Diagnostics.Tests (could not load dependent assembly 'Microsoft.ServiceFabric.Diagnostics.Tests, Version=6.0.0'): Could not load file or assembly 'Microsoft.ServiceFabric.Diagnostics.Tests, Version=6.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. Strong name signature could not be verified.  The assembly may have been tampered with, or it was delay signed but not fully signed with the correct private key. (Exception from HRESULT: 0x80131045)
```

I think we need to add the new assemblies to the `SkipStrongName.ps1`

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/ServiceFabricEventSourceTest.cs
ilijagrbic: Good point. I have a PR that would remove all of the DotNetCoreClr, which I would try to merge soon. In the meantime, I will use NET in this PR as you suggested.

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/Microsoft.ServiceFabric.Diagnostics.Tests.csproj
ilijagrbic: Good catch, I removed it.

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/ServiceFabricEventSourceTest.cs
ilijagrbic: The goal of why I was cutting across scopes was to avoid much more #ifs or to avoid adding a different class just for NetNeutral tests. However I see your point, and I was also puzzled how to make this readable. I implemented your suggestion.

ilijagrbic:
Thanks @olegsych for the review. Unfortunately, I reproduced the error in local and then only confirmed that the fix works in local, since at the time it seemed obvious what was the issue. In the end adding assemblies to `SkipStrongName.ps1` was indeed necessary and now the issue should be fixed. :)

olegsych (approved):

PR #445, opened 2025-07-18
Fix .NET8 warnings

ilijagrbic:
- Recently we moved from `netstandard2.0` -> `net8.0`
- After migration there are several new warnings appearing during the build
- Fixing the warnings except for `BinaryFormater` related ones, which will be fixed in [Remove binary serialization #435](https://github.com/microsoft/service-fabric-services-and-actors-dotnet/pull/435)

### DIFF
--- a/src/FabActUtil/Tool.cs
+++ b/src/FabActUtil/Tool.cs
@@ -307,12 +307,5 @@ namespace FabActUtil
 
             return false;
         }
-
-        private static string GetToolPath()
-        {
-            var codeBase = Assembly.GetEntryAssembly().CodeBase;
-            var uri = new UriBuilder(codeBase);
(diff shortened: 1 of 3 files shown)


src/FabActUtil/Tool.cs
```diff
@@ -307,12 +307,5 @@ private static bool RemoveFrom(IList<string> list, string item, StringComparison
 
             return false;
         }
-
-        private static string GetToolPath()
-        {
-            var codeBase = Assembly.GetEntryAssembly().CodeBase;
-            var uri = new UriBuilder(codeBase);
-            return Uri.UnescapeDataString(uri.Path);
-        }
```
ilijagrbic, line 316: I removed this since it is a private unused method.

olegsych (approved):

PR #456, opened 2025-08-05
Move and expand Telemetry API from WindowsFabric to GitHub

ilijagrbic:
We are moving current Telemetry API to Diagnostics project. Refactor existing tests to use hierarchical test structure.

**We will continue development of managed Telemetry API on GitHub. This PR moves reusable bits of the existing API:**
- `IMeter.cs`
- `NullMeter.cs`

**API will need to be expanded, hence adding additional changes to support further development:**
- Instead of existing `MeterProvider` class, introducing `IMeterProvider` interface, with plans to follow up with dedicated implementation for each metric type for example `Int64MeterProvider`, `DoubleMeterProvider`, `TimespanMeterProvider` (in contrast with current if-else based logic in singular `MeterProvider` class)
- Add `NullMeterProvider` class that returns `NullMeters`. Placeholder implementation before we align on P/Invoke contract which will enable us to implement other meter classes.

**Why no Meter.cs?**
 - Meter class is not moved since we lack the details necessary for its implementation, since alignment on P/Invoke interface is still ongoing. We will follow up with dedicated Meter implementations for each metric type, for example `Int64Meter`, `DoubleMeter`, `TimespanMeter`.

**Other:**
- Update `.editorconfig` with rules for type parameters. Fix naming violations.


test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/Telemetry/ConfigurableMeterProviderTest.cs
```diff
@@ -0,0 +1,139 @@
...
+        const string TestNodeVersion = "12.0.0";
+        const string TestServiceTypeName = "TestServiceType";
+        const string TestServiceUriString = "fabric:/TestApplication/TestService";
+        const string TestApplicationName = "TestApplicationName";
+        const string TestApplicationTypeName = "TestApplicationTypeName";
+        const long ReplicaId = 1L;
+
+        readonly Guid testPartitionId = new Guid();
+        readonly ServiceContext serviceContext;
+        readonly IConfigStore2 configStore = Mock.Of<IConfigStore2>();
```
ilijagrbic: Currently the build is failing because IConfigStore2 is an internal interface in System.Fabric assembly which Diagnostics project doesn't have access to. This will get fixed once we have the ability to release new WindowsFabric nugets.

src/Microsoft.ServiceFabric.Diagnostics/Telemetry/IMeterProvider.cs
```diff
@@ -0,0 +1,18 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Microsoft.ServiceFabric.Diagnostics.Telemetry
+{
+    internal interface IMeterProvider<ValueType>
+    {
+        IMeter<ValueType> CreateMeter(string metricNamespace, string name);
```
ilijagrbic: Unlike previous `MeterProvider`, I have changed the method signatures to not have templates. Rather the interface itself is generic.

src/Microsoft.ServiceFabric.Diagnostics/Telemetry/ConfigurableMeterProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Fabric.Common;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Telemetry
+{
+    internal abstract class ConfigurableMeterProvider<ValueType> : IMeterProvider<ValueType>
+    {
+        readonly protected IDictionary<string, string> systemDimensions = new ConcurrentDictionary<string, string> ();
+        readonly protected bool metricsEnabled;
+
+        protected ConfigurableMeterProvider(IConfigStore2 configStore, ServiceContext serviceContext = null)
```
ilijagrbic: I think allowing service context to be null is better since then we can use this logic outside of regular service context. While configStore can always be used for general info such us node name and runtime version.

src/Microsoft.ServiceFabric.Diagnostics/Telemetry/NullMeter.cs
```diff
@@ -0,0 +1,47 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Microsoft.ServiceFabric.Diagnostics.Telemetry
+{
+    internal class NullMeter<ValueType> : IMeter<ValueType>
+    {
+        public NullMeter() { }
```
olegsych: C# automatically supplies a public parameterless constructor. Let's remove these.

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/Telemetry/NullMeterProviderTest.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+    public abstract class NullMeterProviderTest
+    {
+        public class CreateMeter : NullMeterProviderTest
+        {
+            [Fact]
+            public void ReturnsNullMeterForNoDimension()
+            {
+                NullMeterProvider<long> meterProvider = new NullMeterProvider<long>();
+                Assert.True(meterProvider.CreateMeter("namespace", "meterName") is NullMeter<long>);
```
olegsych: Xunit has a rich Assert API that provides descriptive failure messages for different kinds of checks. Let's use `Assert.IsType<T>`. 

https://api.xunit.net/v3/3.0.0/Xunit.Assert.html

src/Microsoft.ServiceFabric.Diagnostics/Telemetry/ConfigurableMeterProvider.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace Microsoft.ServiceFabric.Diagnostics.Telemetry
+{
+    internal abstract class ConfigurableMeterProvider<TValueType> : IMeterProvider<TValueType>
+    {
+        readonly protected IDictionary<string, string> systemDimensions = new ConcurrentDictionary<string, string>();
+        readonly protected bool metricsEnabled;
+
+        protected ConfigurableMeterProvider(IConfigStore2 configStore, ServiceContext serviceContext = null)
+        {
+            if (configStore == null)
```
ilijagrbic: Remove the IConfigStore2 logic, since these dimensions will already be available on the native side.

src/Microsoft.ServiceFabric.Diagnostics/Telemetry/ConfigurableMeterProvider.cs
```diff
@@ -0,0 +1,46 @@
...
+            {
+                throw new ArgumentNullException(nameof(configStore), "Config store cannot be null.");
+            }
+
+            this.systemDimensions.Add("NodeName", configStore.ReadUnencryptedString("FabricNode", "InstanceName") ?? string.Empty);
+            this.systemDimensions.Add("RuntimeVersion", configStore.ReadUnencryptedString("FabricNode", "NodeVersion")?.Split(':')[0] ?? string.Empty);
+            this.metricsEnabled = (configStore.ReadUnencryptedString("Telemetry/Metrics", "IsEnabled") ?? "false").ToLowerInvariant() == "true";
```
ilijagrbic: Switch for enabling metrics will also be on the native side. Remove.

src/Microsoft.ServiceFabric.Diagnostics/Telemetry/ConfigurableMeterProvider.cs
```diff
@@ -0,0 +1,46 @@
...
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Fabric;
+using System.Fabric.Common;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Telemetry
+{
+    internal abstract class ConfigurableMeterProvider<TValueType> : IMeterProvider<TValueType>
```
olegsych: I would keep the actual implementation out of this PR. Our immediate goal is to give people the instrumentation APIs, i.e. the interfaces. NullMeter and NullMeterProvider should be sufficient for now. I think we would need to implement this class along with the concrete meters to flush out the design.

src/Microsoft.ServiceFabric.Diagnostics/Telemetry/ConfigurableMeterProvider.cs
olegsych: From API standpoint, the most specific definition would be one constructor that takes non-null serviceContext and another constructor without the serviceContext paramter. This could also be a sign that two classes would better represent the same concepts, one that takes the serviceContext and one that doesn't. Conditional logic in object-oriented code is a common smell of architectural problems.

src/Microsoft.ServiceFabric.Diagnostics/Telemetry/IMeter.cs
```diff
@@ -0,0 +1,27 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Microsoft.ServiceFabric.Diagnostics.Telemetry
+{
+    internal interface IMeter<ValueType>
```
olegsych: I would prefer to keep interface definitions in separate files to make the code navigation easier. I guess these are so simple that an exception would be reasonable.

olegsych: Let's add doc comments to all APIs meant to be used by our customers. Normally that means public APIs, but in this case internal types as well.

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/Telemetry/ConfigurableMeterProviderTest.cs
olegsych: This should be irrelevant with today's agreement that the native implementation will provide the NodeName and RuntimeVersion dimensions. But let's keep this test out of this PR regardless, see my comment for the SUT.

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/Microsoft.ServiceFabric.Diagnostics.Tests.csproj
```diff
@@ -9,6 +9,7 @@
     <PackageReference Include="Inspector" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" />
     <PackageReference Include="Moq" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" />
```
olegsych: Oh, this is where the System.Diagnostics.Metrics namespace lives. This looks very odd and I had to check why this is now needed. I wonder if they overspecified the assembly name back when they implemented the DiagnosticSource class.

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/Telemetry/NullMeterProviderTest.cs
```diff
@@ -0,0 +1,52 @@
...
+            }
+
+            [Fact]
+            public void ReturnsNullMeterForOneDimension()
+            {
+                NullMeterProvider<long> longNullMeterProvider = new NullMeterProvider<long>();
+                Assert.True(longNullMeterProvider.CreateMeter("namespace", "meterName", "stringDimension1") is NullMeter1D<long>);
+
+                NullMeterProvider<double> doubleNullMeterProvider = new NullMeterProvider<double>();
```
olegsych: Let's remove test logic for double meters because we aren't planning to implement them. To verify that the meter provider can create meters of expected types, I would create a nested struct, like TestValue and change the logic from creating NullMeterProvider<long> to NullMeterProvider<TestValue>.

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/Telemetry/NullMeterProviderTest.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace Microsoft.ServiceFabric.Diagnostics.Telemetry
+{
+    public abstract class NullMeterProviderTest
+    {
+        public class CreateMeter : NullMeterProviderTest
+        {
+            [Fact]
+            public void ReturnsNullMeterForNoDimension()
+            {
+                NullMeterProvider<long> meterProvider = new NullMeterProvider<long>();
```
olegsych: Extract meterProvider variables to a readonly field in the base test class?

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/Telemetry/NullMeterProviderTest.cs
```diff
@@ -0,0 +1,52 @@
...
+                Assert.True(meterProvider.CreateMeter("namespace", "meterName") is NullMeter<long>);
+            }
+
+            [Fact]
+            public void ReturnsNullMeterForOneDimension()
+            {
+                NullMeterProvider<long> longNullMeterProvider = new NullMeterProvider<long>();
+                Assert.True(longNullMeterProvider.CreateMeter("namespace", "meterName", "stringDimension1") is NullMeter1D<long>);
```
olegsych: I would normally define method parameter values as readonly fields generated by fuzzy, but I guess in this case we'd only save on removing the "s, so it might not be worth the extra lines of code.

olegsych (approved):
Thanks @ilijagrbic. I think we should exclude the ConfigurableMeterProvider from this PR. All other comments are for cosmetics.

olegsych (approved):
Forgot one more thing. Let's move the new types to the Metrics subnamespace. Telemetry is too generic. Plus, Metrics is consistent with the System.Diagnostics.Metrics namespace name.

src/Microsoft.ServiceFabric.Diagnostics/Telemetry/IMeter.cs
olegsych: Forgot one more thing. Let's move the new types to the Metrics subnamespace. Telemetry is too generic. Plus, Metrics is consistent with the System.Diagnostics.Metrics namespace name.

src/Microsoft.ServiceFabric.Diagnostics/Telemetry/ConfigurableMeterProvider.cs
ilijagrbic: This is a good point. And now it raised the question of naming for these classes. Perhaps `ServiceMeterProvider` and `StandaloneMeterProvider` would be better, since it would communicate better to the user when to use which one, rather than exposing internal logic in the name (ie. current name mentions Configuration, which is an implementation detail). Anyways, I will remove this classes for now, and follow up with this implementation in a follow up PR.

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/Telemetry/ConfigurableMeterProviderTest.cs
ilijagrbic: Agreed, this is something that I will follow up in subsequent PR.

src/Microsoft.ServiceFabric.Diagnostics/Telemetry/IMeter.cs
ilijagrbic: Do you think we should split implementations (ie for NullMeters) in seperate files? I feel that for NullMeters specifically it might be an overkill, but worth considering for future int/timespan implementations.

test/unittests/Microsoft.ServiceFabric.Diagnostics.Tests/Microsoft.ServiceFabric.Diagnostics.Tests.csproj
ilijagrbic: Turns out I didn't need this reference. It was needed in one of the tests implemented in WindowsFabric's version of the API, which I in the end decided not to move in this PR, but he reference stayed.

src/Microsoft.ServiceFabric.Diagnostics/Telemetry/IMeter.cs
olegsych: I don't think we should block this PR to do it, but generally speaking yes, I think that interfaces, classes and structs should be in separate files. This helps to make the code easier to navigate and duplication more noticeable. When multiple types are in the same file, it's easier to copy&paste them and move on without careful consideration and testing. This problem is evident in FabricEventSource, where we kept adding more and more event method overloads with more and more parameters. I suspect that we already have this problem with the IMeter interfaces because three custom dimensions is not enough if you don't have system dimensions and too much if you do.

olegsych (approved):

src/Microsoft.ServiceFabric.Diagnostics/Telemetry/IMeter.cs
ilijagrbic: Thanks for the answer, I will make sure to follow up on this in the subsequent PRs.

PR #457, opened 2025-08-06
Add general and test specific GitHub Copilot instructions

ilijagrbic:
- Add initial copilot instructions to the repo, to facilitate better code generation with GitHubCopilot
- Coding standards are according to existing .editorconfig, aim of this PR is not to change or revisit them, rather just help Copilot better use them
- Inspector library which we use for white-box testing is not commonly known and Copilot strugled to generate code for it. Adding examples in a dedicated `.instructions.md` file.


.github/copilot-instructions.md
```diff
@@ -0,0 +1,44 @@
...
+- Contains APIs for Reliable Services and Reliable Actors
+- Supports both .NET Framework and .NET targets
+- Uses MSBuild and NuGet for package management
+
+## Documentation
+- Update XML documentation for all public APIs
+- Only generate documentation for public APIs
+
+## Testing
+- Use MSTest framework following existing patterns
```
olegsych: This should be xUnit

.github/copilot-instructions.md
```diff
@@ -0,0 +1,44 @@
...
+- When writing tests, use a hierarchical pattern similar to existing tests. For example:
+  public abstract class SystemUnderTestClassTest
+```csharp
+
+    namespace Microsoft.ServiceFabric.TestedProject
+    {
+        public abstract class SystemUnderTestClassTest
+        {
+            // Setup common test context
+        }
```
olegsych: We're nesting member test classes inside the class test class

.github/instructions/inspector.instructions.md
```diff
@@ -0,0 +1,662 @@
+---
+applyTo: "**"
+---
+# Whitebox Testing
+- Where possible, use inspector for reflection to simplify test setup and make tests more maintainable.
```
olegsych, line 5: Can we link to the github repo instead of duplicating the instructions here?

.github/copilot-instructions.md
ilijagrbic: This slipped by me, you are right, I will fix it.

.github/instructions/inspector.instructions.md
ilijagrbic: That would be ideal, and we would avoid duplicated docs. However, I haven't seen an option to do this, at least not according to the [Copilot customization documentation.](https://code.visualstudio.com/docs/copilot/copilot-customization)

I don't think this is a big deal, since Interceptor library probably won't change rapidly, so we won't have the need to update these instructions soon.

olegsych (approved):

PR #458, opened 2025-08-11
Implement integration test for reminder registration and reminder callback

plave0:
❗**Important:** I have both renormalized and edited _code.sln_ in this PR. To see the actual changes made to this file, please exclude commit "Renormalize code.sln" `f4a2e1ea959ccb76ec2fd02b1327ba12f9558004` when viewing changed files.

The following has been implemented:
- Base class `ActorIntegrationTest` which contains `GetActorService` which creates an `ActorService` objects and mocks its dependencies. It uses `NullActorStateProvider` as a state provider which stores data in-memory and does not have any dependencies to the SF runtime. This makes it suitable for implementing simple integration tests. 
- Test `ReminderParametersPersistBetweenRegisteringAndReceivingReminder` which creates an `ActorService` which hosts actors of type `TestableActor`. The test registers a reminder on one actor, checks the return value, waits for the reminder to fire and finally checks whether the callback was executed successfully.

Changes to GitHub actions:
- The "Build Test" action has been modified by adding "Uninstall SF SDK assemblies from GAC". This step removes `Microsoft.ServiceFabric.Data.Interfaces` and `Microsoft.ServiceFabric.Data.Interfaces.V2` from the GAC on the runner.  `Microsoft.ServiceFabric.Actors.ItegrationTests` targeting .NET Framework 4.6 references `Microsoft.ServiceFabric.Data` version 9 package which contains `Microsoft.ServiceFabric.Data.Interfaces.dll` version 5.0. The runner has `Microsoft.ServiceFabric.Data.Interfaces.dll` version 5.0 installed in the GAC, but this assembly corresponds to `Microsoft.ServiceFabric.Data` version 8. In this case, the .NET runtime chooses to load the assembly found in the GAC, which causes a `System.MissingMethodException` at runtime because the wrong version of the assembly is being used. These changes to the GitHub action are a temporary solution until assembly versions are changed to match the version of the SDK they belong to.


properties/service_fabric_managed_netframework.targets
```diff
@@ -15,7 +15,8 @@
                                    $(RepoRoot)test\unittests\Microsoft.ServiceFabric.Services.Remoting.Tests\Microsoft.ServiceFabric.Services.Remoting.Tests.csproj;
                                    $(RepoRoot)test\unittests\Microsoft.ServiceFabric.Services.Tests\Microsoft.ServiceFabric.Services.Tests.csproj;
                                    $(RepoRoot)test\unittests\Microsoft.ServiceFabric.Actors.StateMigration.Tests\Microsoft.ServiceFabric.Actors.StateMigration.Tests.csproj;
-                                   $(RepoRoot)test\unittests\Microsoft.ServiceFabric.Diagnostics.Tests\Microsoft.ServiceFabric.Diagnostics.Tests.csproj" />
+                                   $(RepoRoot)test\unittests\Microsoft.ServiceFabric.Diagnostics.Tests\Microsoft.ServiceFabric.Diagnostics.Tests.csproj;
+                                   $(RepoRoot)test\Microsoft.ServiceFabric.Actors.IntegrationTests\Microsoft.ServiceFabric.Actors.IntegrationTests.csproj" />
```
ilijagrbic, line 20: Just a remainder that we probably need to add the assembly for this new project in the `SkipStrongName.ps1`. Currently the tests are not being run in the pipeline.

plave0: Thank you Ilija for pointing this out. I have added the new test project to the list.

code.sln
```diff
@@ -1,202 +1,210 @@
...
-		{B91993CC-B3CC-4B6F-9474-E582B8505C12} = {BB93122C-21C5-4BB8-B385-74FC423D6027}
-		{B8F12A3E-4C5D-4F6A-8B7C-9D1E2F3A4B5C} = {BB93122C-21C5-4BB8-B385-74FC423D6027}
-		{1BFA8614-0A48-4D17-97E5-A34E630A4C2B} = {BB93122C-21C5-4BB8-B385-74FC423D6027}
-		{A1B2C3D4-E5F6-7890-1234-56789ABCDEF0} = {BB93122C-21C5-4BB8-B385-74FC423D6027}
-	EndGlobalSection
-	GlobalSection(ExtensibilityGlobals) = postSolution
-		SolutionGuid = {8CC7D7F7-CBD9-428B-A190-7A4D56C920FE}
-	EndGlobalSection
-EndGlobal
+Microsoft Visual Studio Solution File, Format Version 12.00
```
olegsych, line 1: This file has completely changed because the line endings in the index were changed from CRLF to LF. This will create merge conflicts with other PRs modifying this file. I think it's a good idea to re-normalize the line endings to LF and created https://github.com/microsoft/service-fabric-services-and-actors-dotnet/pull/462 to do it in the entire repository at once. If we can fix this problem everywhere at once, I think it would be best to avoid mixing line ending changes with the functional changes.

src/Microsoft.ServiceFabric.Actors/Friend.cs
```diff
@@ -9,6 +9,7 @@
...
+[assembly: InternalsVisibleTo("Microsoft.ServiceFabric.Actors.IntegrationTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")][assembly: InternalsVisibleTo("Microsoft.ServiceFabric.Actors.Tests, PublicKey=00240000048000009400000006020000002400005253413100040000010001003f8c902c8fe7ac83af7401b14c1bd103973b26dfafb2b77eda478a2539b979b56ce47f36336741b4ec52bbc51fecd51ba23810cec47070f3e29a2261a2d1d08e4b2b4b457beaa91460055f78cc89f21cd028377af0cc5e6c04699b6856a1e49d5fad3ef16d3c3d6010f40df0a7d6cc2ee11744b5cfb42e0f19a52b8a29dc31b0")]
```
olegsych: Could you put the two attributes on separate lines?

olegsych: Do we need both? There is only one for the Services assembly.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Microsoft.ServiceFabric.Actors.IntegrationTests.csproj
```diff
@@ -0,0 +1,18 @@
...
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" />
+    <PackageReference Include="xunit" />
+    <PackageReference Include="xunit.runner.visualstudio" />
+    <PackageReference Include="Moq" />
+    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="$(RepoRoot)src\Microsoft.ServiceFabric.Actors\Microsoft.ServiceFabric.Actors.csproj" />
+    <ProjectReference Include="$(RepoRoot)test\unittests\Microsoft.ServiceFabric.TestFramework\Microsoft.ServiceFabric.TestFramework.csproj" />
```
olegsych: I don't think we're using anything from the TestFramework at the moment. I'm adding suggestions on using the IFuzzExtensions in the tests, but if we don't need the reference, let's remove it from the project to keep the dependency graph as simple as possible.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Microsoft.ServiceFabric.Actors.IntegrationTests.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <ProjectGuid>788378DE-D58E-3914-CE90-3498F7365C4F</ProjectGuid>
+    <TargetFrameworks>net9.0;net462</TargetFrameworks>
+  </PropertyGroup>
+  <Import Project="..\..\properties\service_fabric_managed_common.props" />
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" />
```
olegsych, line 10: Could you sort the package references? I usually use the Sort Selected Lines command in VSCode; it's not the most logical order, but better than random, which leads to merge errors and conflicts.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Reminders/ReminderRegistrationTest.cs
```diff
@@ -0,0 +1,145 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.ServiceFabric.Actors.Runtime;
+using Microsoft.ServiceFabric.Services.Communication;
+using Xunit;
+
+namespace Microsoft.ServiceFabric.Actors.IntegrationTests
+{
+public class ReminderRegistrationTest : ActorIntegrationTest
```
olegsych: Could you make the indentation consistent?

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Reminders/ReminderRegistrationTest.cs
```diff
@@ -0,0 +1,145 @@
...
+using System.Reflection.Emit;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.ServiceFabric.Actors.Runtime;
+using Microsoft.ServiceFabric.Services.Communication;
+using Xunit;
+
+namespace Microsoft.ServiceFabric.Actors.IntegrationTests
```
olegsych: Could you remove the IntegrationTests suffix from the namespace? We want the test namespace hierarchy matching the product namespace as closely as possible and test suffixes only get in the way.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorIntegrationTest.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Numerics;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.ServiceFabric.Actors.Runtime;
+using Microsoft.ServiceFabric.Services.Runtime;
+using Moq;
+
+namespace Microsoft.ServiceFabric.Actors.IntegrationTests
+{
+    public class ActorIntegrationTest
```
olegsych: Let's make this class abstract to communicate that it's meant to be used as a base class for the actual integration tests.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Reminders/ReminderRegistrationTest.cs
```diff
@@ -0,0 +1,145 @@
...
+                }
+
+                await this.StateManager.AddOrUpdateStateAsync(this.fireCountKey, 1, (stateName, oldValue) => { return oldValue + 1; });
+            }
+        }
+
+        [Fact]
+        public async Task ReminderSuccessfullyRegisteredAndFired()
+        {
+            ActorService actorService = await this.GetActorService<RemiderTestActor>();
```
olegsych: I think the term `this` is redundant and sometimes misleading, such as in this case. The GetActorService method is defined in the base class and yet the code doesn't say base.GetActorService, which sends the reader looking for a method that doesn't exist. I try to avoid it completely, except where it allows distinguishing between parameters and fields with the same name, and instead focus on making the methods as granular as possible so that there is no need to qualify method calls.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorIntegrationTest.cs
```diff
@@ -0,0 +1,93 @@
...
+            await statefulServiceReplica.RunAsync(CancellationToken.None);
+
+            return actorService;
+        }
+
+        /// <summary>
+        /// Gets a mock StatefulServiceContext.
+        /// </summary>
+        /// <returns>A mock StatefulServiceContext.</returns>
+        protected StatefulServiceContext GetMockStatefulServiceContext()
```
olegsych: Can we use IFuzzExtensions from the TestFramework instead of defining new helpers here? We can setup additional properties there and, as long as they are generated/fuzzy, it shouldn't result in tight coupling between different test projects. To make the fuzzy strings more understandable, you can specify a prefix for fuzzy.String() calls.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Reminders/ReminderRegistrationTest.cs
```diff
@@ -0,0 +1,145 @@
...
+{
+public class ReminderRegistrationTest : ActorIntegrationTest
+    {
+        interface IReminderTestActor : IActor
+        {
+            Task<IActorReminder> InitStateAndCreateReminder(string reminderName, byte[] state, TimeSpan dueTime, TimeSpan period);
+            Task<bool> ValidateActorState();
+        }
+
+        class RemiderTestActor : Actor, IReminderTestActor, IRemindable
```
olegsych: If I understand correctly, we need this class to access the protected APIs defined by ActorBase and Actor. I would keep as much testing logic in the tests themselves and out of this actor class as possible. It would help making the test code easier to read and avoid having to switch back and force to see what's going on. I usually create a subclass that inherits from a class I need to use in tests, make it define public APIs that call the inherited protected ones, and call it TestableSomething. I think it falls somewhere between the Stub and Spy patterns described here https://learning.oreilly.com/library/view/xunit-test-patterns/9780131495050/ch23.html.

olegsych: For reminders specifically, I would define a public field of delegate type matching the signature of the the IRemindable.ReceiveReminderAsync method and make the method implementation call it like so

```
public Func<string, byte[], TimeSpan, TimeSpan, Task> ReceiveReminderCallback = (name, state, dueTime, period) => Task.CompletedTask; 

Task ReceiveReminderAsync(string name, byte[] state, TimeSpan dueTime, TimeSpan period)
{
    return ReceiveReminderCallback(name, state, dueTuime, period);
} 
```

This way the tests could supply the ReceiveReminderCallback delegate appropriate for the test and we don't have to hard-code it in the TestableActor class.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Reminders/ReminderRegistrationTest.cs
```diff
@@ -0,0 +1,145 @@
...
+                    {
+                        return ((RemiderTestActor)actorBase).ValidateActorState();
+                    },
+                    callContext: "TestCallContext",
+                    timerCall: false,
+                    cancellationToken: new CancellationToken());
+            });
+
+            Assert.Null(exception);
+        }
```
olegsych, line 141: Looking at the assertion blocks and the ValidateActorState, I think the primary purpose of this test is to verify that the reminder was invoked with expected parameters. The use of actor state appears to be accidental at this point. Try reworking it using the approach I outlined above (the TestableActor suggestions).

If that approach works, I think we should try structuring the integration tests similar to unit tests - by the types and members they are testing, in this case Actor and its RegisterReminderAsync. I've never worked on a project where scenario-based test taxonomies were successful in the long run, but I know that class/member based taxonomies can be maintained effectively.

olegsych (changes requested):
Thanks @plave0. I was pleasantly surprised by how simple the setup for integration testing is. That's awesome!

I've left a bunch of suggestions as I was reviewing the PR, but at the high level, I would like you to try:
- Changing ReminderTestActor into a simple TestableActor that allows tests to invoke the protected APIs inherited from the Actor and ActorBase classes. This way it should be possible to express the arrange, act and assert logic completely in the tests.
- Arranging tests by product class/method rather than by scenario as much as possible, such as in this case, where we're testing the ActorBase.RegisterReminderAsync method.

src/Microsoft.ServiceFabric.Actors/Friend.cs
plave0: We don't need them, this was duplicated accidentally.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Reminders/ReminderRegistrationTest.cs
plave0: I think that's a good idea, but I don't see us using this approach in other test projects. Should we remove the "Text" suffix form other test projects?

olegsych: I think so. This would be a good one to ask the Copilot to do.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Reminders/ReminderRegistrationTest.cs
```diff
@@ -0,0 +1,144 @@
...
+                    actualState = state;
+                    actualDueTime = dueTime;
+                    actualPeriod = period;
+
+                    reminderCallbackInvocationCounter += 1;
+                };
+
+                Func<ActorService, ActorId, ActorBase> actorFactory = (actorService, actorId) => new TestableActor(actorService, actorId, receiveReminderCallback);
+
+                ActorService actorService = await GetActorService<TestableActor>(actorFactory);
```
gobradovic, line 113: nit: First argument is "expected", second is "actual" as in the section above.

gobradovic (approved):

test/Microsoft.ServiceFabric.Actors.IntegrationTests/App.config
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+    <runtime>
+        <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+            <dependentAssembly>
+                <assemblyIdentity name="System.Fabric" publicKeyToken="31bf3856ad364e35" culture="neutral" />
+                <bindingRedirect oldVersion="11.0.0.0" newVersion="12.0.0.0" />
```
olegsych: Why do we need this explicit binding redirect?

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Reminders/ReminderRegistrationTest.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            IStatefulUserServiceReplica statefulServiceReplica = actorService;
+            await statefulServiceReplica.OnOpenAsync(ReplicaOpenMode.New, CancellationToken.None);
+            await statefulServiceReplica.OnChangeRoleAsync(ReplicaRole.Primary, CancellationToken.None);
+            await statefulServiceReplica.RunAsync(CancellationToken.None);
+
+            return actorService;
+        }
+
+        interface ITestableActor : IActor
```
olegsych, line 38: Do we have to have the ITestableActor interface? Can we cast directly to TestableActor instead?

olegsych (approved):
:shipit:

test/Microsoft.ServiceFabric.Actors.IntegrationTests/App.config
plave0: Thank you for noticing this. We do not need the binding redirect anymore since we are deleting `Data.Interfaces` assemblies from the GAC

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Reminders/ReminderRegistrationTest.cs
plave0: We cannot. If actor class does not inherit from a and interface that inherits from `IActor`, we get a runtime error: 
```
Q:\source\service-fabric-services-and-actors-dotnet\src\Microsoft.ServiceFabric.Actors\Runtime\ActorTypeInformation.cs(150): error TESTERROR:
      Microsoft.ServiceFabric.Actors.ActorIntegrationTest+RegisterReminderAsync.ReminderParametersPersistBetweenRegisteringAndReceivingReminder (1s 732ms): Error Message: System.ArgumentException : The actor t
      ype 'Microsoft.ServiceFabric.Actors.ActorIntegrationTest+TestableActor' does not implement any actor interfaces or one of the interfaces implemented is not an actor interface. All interfaces(including it
      s parent interface) implemented by actor type must be actor interface. An actor interface is the one that ultimately derives from 'Microsoft.ServiceFabric.Actors.IActor' type.
      Parameter name: actorType
      Stack Trace: ...
```
Even when inheriting from `IActor` directly we get the same error.

This check is implemented in `ActorTypeInformation.Get` method.

gobradovic (approved):

PR #459, opened 2025-08-12
Implement Metrics API Interop and Int64 meters

ilijagrbic:
**Interop:**
- In order to enable Metrics API to call native implementation adding: `ComImport` interfaces and `DllImport` methods
- Using new Marshaling for `string[]`, new approach is tested by creating a dummy dll

**New meters:**
- In order to support `long` metrics, implementing `Int64Meter` and `Int64MeterProvider`


src/Microsoft.ServiceFabric.Diagnostics/Metrics/Interop/NativeRuntimeMethods.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            Marshal.ThrowExceptionForHR(PInvoke.FabricCreateMeterProvider(out IFabricMeterProvider meterProvider));
+            return meterProvider;
+        }
+
+        static class PInvoke
+        {
+            private const string FabricTelemetryLib = "FabricTelemetry";
+
+            [DllImport(FabricTelemetryLib)]
```
ilijagrbic: TODO: Use new interop

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64MeterProviderTest.cs
```diff
@@ -0,0 +1,118 @@
...
+            [Fact]
+            public void ShouldCreateMeterWithCorrectSystemDimensions()
+            {
+                var meter = sut.CreateMeter(testNamespace, testMetric);
+
+                Assert.NotNull(meter);
+                Assert.IsType<Int64Meter>(meter);
+
+                var expectedSystemDimensions = sut.Protected().Field<IList<string>>().Value;
+                Assert.Equal(expectedSystemDimensions, ((Int64Meter)meter).Field<IList<string>>().Value);
```
ilijagrbic: TODO: Verify that ComInterface is being called.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Interop/IFabricMeter.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Interop
+{
+    [ComImport]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    [Guid("A0D80970-C062-4C59-A3CA-7AEEB901B49C")]
+    internal interface IFabricMeter
+    {
+        [MethodImpl(MethodImplOptions.InternalCall, MethodCodeType = MethodCodeType.Runtime)]
+        void Record([In] long value, [In][MarshalAs(UnmanagedType.LPArray, ArraySubType = UnmanagedType.LPWStr)] string[] dimensionValues, [In] uint count);
```
ilijagrbic: Think about removing `[In]` if this goes into WindowsFabric after interop changes

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Interop/IFabricMeterProvider.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Interop
+{
+    [ComImport]
+    [Guid("15AD37D2-F641-4188-824B-0D68CB4F6C17")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    internal interface IFabricMeterProvider
+    {
+        [MethodImpl(MethodImplOptions.InternalCall, MethodCodeType = MethodCodeType.Runtime)]
+        [return: MarshalAs(UnmanagedType.Interface)]
+        IFabricMeter CreateMeter([In][MarshalAs(UnmanagedType.LPWStr)] string metricNamespace, [In][MarshalAs(UnmanagedType.LPWStr)] string name, [In][MarshalAs(UnmanagedType.LPArray, ArraySubType = UnmanagedType.LPWStr)] string[] dimensionNames, [In] uint count);
```
ilijagrbic: Think about removing [In] if this goes into WindowsFabric after interop changes

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64Meter.cs
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.ServiceFabric.Diagnostics.Metrics.Interop;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics
+{
+    internal class Int64Meter : MeterBase, IMeter<long>
```
olegsych: Let's make the new classes `sealed` unless there is a specific reason to allow them to serve as base classes. Keeping public APIs unsealed was a decent advice in the 90s back when the first version of the .NET Framework design guidelines was published, but I don't think it still holds true today.

olegsych: Let's avoid specifying internal visibility for top-level types. The default visibility in C# is internal, so there is no need to specify it explicitly. Knowing how meticulous Anders Hejlsberg is, I don't think these two choices (the default and the default visibility) were accidental. I think we should make our C# code as idiomatic as possible. Unfortunately, we have a lot of bad habits that were introduced by StyleCop which was developed outside of the C# team by a very opinionated engineer who must have come from the C++ background because some of the default style rules were plain wrong. I think that specifying redundant accessibility modifiers is one of the more counterproductive ones.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64Meter.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        public Int64Meter(IFabricMeter fabricMeter, IList<string> systemDimensionValues) : base(fabricMeter, systemDimensionValues)
```
olegsych: Let's use `internal` instead of `public` to specify accessibility of members of internal types. It's not a problem here because the class is small; in larger classes, where you can't see both the class and the member definitions on the same page, the inconsistent visibility becomes confusing. In this case the effective visibility of the members is internal, despite them being defined as public.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/MeterBase.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using System.Collections.Generic;
+using Microsoft.ServiceFabric.Diagnostics.Metrics.Interop;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics
+{
+    internal class MeterBase
+    {
+        readonly protected IList<string> systemDimensionValues;
```
olegsych: Let's try to use immutable and the simplest types as much as possible to help reduce the risk of mistakes in multi-threaded async services. For collections, the simplest is IEnumerable<T>, if we need to access number of items without re-calculating it every time, we should use IReadOnlyCollection<T> and if we need to access items by index rather than by enumerating them, we should use IReadOnlyList<T>.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/NullMeter1D.cs
```diff
@@ -0,0 +1,15 @@
...
+// ------------------------------------------------------------
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics
+{
+    internal class NullMeter1D<TValueType> : IMeter1D<TValueType>
+    {
+        public void Record(TValueType value, string dimension1)
+        {
+
+        }
```
olegsych: The blank line in the middle of the method looks accidental. I would define methods like this on a single line because the NullMeter name should be sufficient to explain why it's empty.
```
public void Record(TValueType value, string dimension1) {}
```

src/Microsoft.ServiceFabric.Diagnostics/Metrics/ServiceMeterProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Fabric;
+using System.Linq;
+using Microsoft.ServiceFabric.Diagnostics.Metrics.Interop;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics
+{
+    internal abstract class ServiceMeterProvider<TValueType> : IMeterProvider<TValueType>
```
olegsych: Let's add a brief doc comment to each type we expect our internal customers. Just a `<summary>` should be enough.

olegsych: Usually, I don't add doc comments for internal types because customers would be using the public types only, however, in this case, all of our APIs are internal. What do you think about renaming the `Interop` subnamespace to `Implementation` and moving the implementation types that customers shouldn't be using into it so that the `Metrics` namespace only has the APIs meant for customer usage.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/ServiceMeterProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+        readonly private IList<string> systemDimensionNames = new List<string>
+        {
+            "ReplicaOrInstanceId",
+            "PartitionId",
+            "ServiceTypeName",
+            "ServiceName",
+            "ApplicationName",
+            "ApplicationTypeName"
+        };
+        readonly protected IList<string> systemDimensionValues = new List<string>();
```
olegsych: I usually place accessibility keyword first, i.e. `protected readonly` because accessibility is the API aspect whereas readonly is the implementation aspect. People reading the code would be more interested in the API.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/ServiceMeterProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+        readonly protected IFabricMeterProvider fabricMeterProvider;
+
+        private static Func<IFabricMeterProvider> createFabricMeterProvider = () => NativeRuntimeMethods.FabricCreateMeterProvider();
```
olegsych: I would define it like this:

```
readonly Func<IFabricMeterProvider> createFabricMeterProvider = NativeRuntimeMethods.FabricCreateMeterProvider;
```

I wouldn't use the lambda because by assigning a method directly to the delegate we are enabling a unit test to verify that the field was initialized with the expected method. You could do this with lambdas, but it's more complicated. 

I wouldn't use a static field because readonly static fields can't be changed through reflection. I would use a readonly instance field because I think immutability is very important and worth the negligible overhead of an instance field.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics
+{
+    public abstract class Int64Meter1DTest
+    {
+        readonly internal IFabricMeter fabricMeter = Mock.Of<IFabricMeter>();
+
+        public class Class : Int64Meter1DTest
+        {
+            [Fact]
+            public void InheritsFromInt64MeterBase()
```
olegsych: I'm not sure what this test is meant to verify. I would replace it with a constructor test to verify that the constructor initializes the base with given arguments.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
```diff
@@ -0,0 +1,34 @@
...
+                Assert.IsAssignableFrom<MeterBase>(meter);
+            }
+        }
+
+        public class Record : Int64Meter1DTest
+        {
+            [Fact]
+            public void CallsFabricMeterRecord()
+            {
+                var systemDimenisions = new List<string> { "systemDimension1", "systemDimension2" };
```
olegsych: I try to avoid hard-coded literals and use fuzzy instead. Here I would use `fuzzy.List(fuzzy.String)`. In the typical TDD flow you're supposed to start with a simplest test, like a test with one system dimension, add another test with second dimension, etc. I prefer fuzzy arguments because they help to reduce the typical number of tests. 

I also like sabotage testing, when you try to change the product code to produce false test results, such as false success. It's too easy to sabotage the product code when the tests use hard-coded literals.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
```diff
@@ -0,0 +1,34 @@
+using System.Collections.Generic;
+using Microsoft.ServiceFabric.Diagnostics.Metrics.Interop;
+using Moq;
+using Xunit;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics
+{
+    public abstract class Int64Meter1DTest
+    {
+        readonly internal IFabricMeter fabricMeter = Mock.Of<IFabricMeter>();
```
olegsych: Let's use the most restrictive accessibility possible - protected is more restrictive than internal

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public class Record : Int64Meter1DTest
+        {
+            [Fact]
+            public void CallsFabricMeterRecord()
+            {
+                var systemDimenisions = new List<string> { "systemDimension1", "systemDimension2" };
+                var meter = new Int64Meter1D(fabricMeter, systemDimenisions);
+                meter.Record(42, "customDimension1");
+                Mock.Get(fabricMeter).Verify(m => m.Record(42, It.Is<string[]>(arr => arr.Length == 3 && arr[0] == "systemDimension1" && arr[1] == "systemDimension2" && arr[2] == "customDimension1"), 3), Times.Once);
```
olegsych: Ugh... This is very verbose. A couple of thoughts:
- Defining fabricMeter as Mock<T> instead of T would shorten this to fabricMeter.Verify(...)
- Extracting the lambda passed to It.Is to a local variable on a separate line would shorten the Verify call.
- Another technique I like is using the Mock as a Spy, i.e. add code above, in the arrange section, to save the actual value that needs to be verified in a local variable and then the assert section could verify the local variable instead of verifying the mocked call. It gets uglier for function calls with multiple arguments, so it wouldn't be my first choice here.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/TestServiceContext.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+using System.Fabric;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics
+{
+    internal class TestServiceContext : ServiceContext
```
olegsych: Why do we need TestServiceContext?

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/MeterBaseTest.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics
+{
+    public abstract class MeterBaseTest
+    {
+        readonly internal IFabricMeter fabricMeter = Mock.Of<IFabricMeter>();
+
+        public class Constructor : MeterBaseTest
+        {
+            [Fact]
+            public void SholdThrowArgumentNullExceptionWhenMeterNameIsNull()
```
olegsych: I would drop the term Should, so that it reads simply ThrowsArgument...

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Interop/NativeRuntimeMethods.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        internal static IFabricMeterProvider FabricCreateMeterProvider()
+        {
+            Marshal.ThrowExceptionForHR(PInvoke.FabricCreateMeterProvider(out IFabricMeterProvider meterProvider));
+            return meterProvider;
+        }
+
+        static class PInvoke
+        {
+            private const string FabricTelemetryLib = "FabricTelemetry";
```
olegsych: I would remove private - class and struct members are private by default.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Interop/NativeRuntimeMethods.cs
```diff
@@ -0,0 +1,29 @@
...
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System.Runtime.InteropServices;
+using HRESULT = System.Int32;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Interop
+{
+    internal static class NativeRuntimeMethods
```
olegsych: If we follow the WindowsFabric convention, this class should be called `NativeTelemetry`. However, I'm not sure if I like this pattern much. We kind of gotten into trouble by following it and creating gigantic classes that serve completely unrelated areas of the code. The most encapsulated solution would be to put this code in the ServiceMeterProvider, which is the only class that needs to invoke the native function right now. 

By the way, the nested PInvoke class was meant to keep the existing function definitions the same and avoid changing the internal API of System.Fabric which would break the Microsoft.ServiceFabric.Data.Impl in NetCore services. With a single call to the FabricCreateMeterProvider, it might be reasonable to do without it completely and test that the ServiceMeterProvider invokes it correctly, i.e. converts HRESULT errors to COMExceptions....

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64Meter1D.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+        }
+
+        public void Record(long value, string dimension1)
+        {
+            var allDimensionsList = new List<string>(systemDimensionValues)
+            {
+                dimension1
+            };
+            string[] allDimensionArray = allDimensionsList.ToArray();
```
olegsych: Record implementations are performance critical, so we should avoid heap allocations if at all possible. Here there are potentially 3 - one to create the List, one to convert the list to an array and possibly one to add dimensions to the list. Lists preallocate memory to hold more items than needed, but when the buffer is full, a new item can result in re-allocating the buffer and copying its contents to the new location.

olegsych: Try stackalloc with arrays https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/stackalloc
I was also thinking about Span<T>, which should be superior to arrays in terms of implementation APIs but don't know how well it works with PInvoke and have specific suggestions. 
This guy seems to have researched it a lot: https://ericsink.com/entries/utf8z.html

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Interop/IFabricMeter.cs
olegsych: Based on the description of https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.unmanagedtype, I think the count parameter could be removed if we make the arrays null-terminated like the strings themselves.

olegsych: Good idea. It shouldn't be needed in most cases.

olegsych (changes requested):
Thanks @ilijagrbic I've left a bunch of thoughts in the comments. The one thing I think would be important to think about is how we communicate which APIs are meant for customers and which ones are implementation details. The other thing is the implementation efficiency, but we should be able to iterate on it and don't need to make it perfect from the start.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/ServiceMeterProvider.cs
ilijagrbic: This is a good suggestion, and I was thinking about similar solutions. I will rename the namespace and move the implementation types.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64Meter.cs
ilijagrbic: Agreed, I will fix this.

ilijagrbic: Agreed, I will fix this as well. 
One reason I think this keeps happening is that Visual Studio adds internal by default when creating a new class. And it aligned with my thought process since I wanted to have these classes as internal, so I overlooked that this is not needed. Not sure if there is a way to specify to Visual Studio to not generate it, I will look into it.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64Meter.cs
ilijagrbic: Yeah, this makes sense. However, sine interface methods need to be public, we will still have a mix - internal constructor and public methods. From my perspective, it comes down to preference regarding these nuances. I will switch the constructors to internal to be consistent as possible.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/NullMeter1D.cs
ilijagrbic: Yeah, this happened because of the copy-paste from WindowsFabric. I will fix it.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/ServiceMeterProvider.cs
ilijagrbic: I agree on point regarding lambdas.

However, I'm not sure about the static field issue, since we need to call this `Func<>` in the constructor. And the only way we can mock it before creating the instance is through a static field.
This is why we used a non-readonly static field in the `ServiceFabricEventSource` as well. In that case we also had constructor logic that was needed to be mockable before invoking the constructor.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
ilijagrbic: If we use protected instead, we get an error since the field (which would be protected in a public class) would be more accessible than the `IFabricMeter` type (which is internal). Changing the class from public to internal would solve this issue, but the test classes need to be public to my knowledge, in order for them to be executed. So that's why I used internal here.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
ilijagrbic: We verify that the Int64Meter inherits from the `MeterBase` class. This class has the constructor with initialization logic for the fields.
I started with the approach you suggested, but then I had a bunch of copy-pasted tests for different Meter classes (1D, 2D, ...)
Currently we have separate tests for `MeterBase`, while here I just verify that other classes inherit from it.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/TestServiceContext.cs
ilijagrbic: `ServiceContext` is an abstract class one option was to use a specific implementation - such as `StatefulServiceContext`/`StetlessServiceContext`. But I wanted to have a general test - for any possible `ServiceContext`. We can easily just use one of the existing implementations though, if you think this is an overkill.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
ilijagrbic: Turns out this can all be fixed by just using private (instead of internal). Nested test classes can still access private members of the root class. :)

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
ilijagrbic: I agree with all of this. I have worked on improving these tests, so hopefully they are better now.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
ilijagrbic: These are all good points, as I mentioned in the other comment, I worked on improving these tests, so hopefully the new version is better.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/MeterBaseTest.cs
ilijagrbic: Good observation, I guess my old habit with naming tests kicked in here. I fixed this. :)

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Interop/NativeRuntimeMethods.cs
ilijagrbic: Unfortunately, `DllImport` attribute can't be applied within a generic type - which `ServiceMeterProvider` is.

Regarding the `PInvoke` class, I wanted to have it for consistency within the interop code, and I also like having a nested class to group the `PInvoke` functions - which might not be needed when we have just 1 function.

I will remove the `PInvoke` class and rename the interop class into `NativeTelemetry` as you suggested.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Interop/IFabricMeter.cs
ilijagrbic: It probably could be removed, however, I just see no major reason to tackle this. Passing the size parameter is not a significant overhead in terms of performance - since value types are passed through stack. And in terms of readability and cleanliness, I would argue it is better since it is obvious how and what are we doing, instead of relying on a functionality that is not as obvious to the reader.

olegsych: Sounds good, could we move it in front of the array parameter though? The count plus pointer/data pattern goes way back almost as far as null-terminated strings; it's very odd to see them in reverse order.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/TestServiceContext.cs
olegsych: Ah, makes sense. I agree that it makes sense to use a ServiceContext descendant other than the Stateful/Stateless classes. This helps to avoid taking accidental dependencies on the unrelated product classes. However, I would prefer to use Mock instead of hard-coding a new class. The ServiceContext constructor is protected, but System.Fabric already has InternalsVisibleTo("DynamicProxyGenAssembly2") so it should be possible. I think we already did something similar in EventSource tests: `var listener = new Mock<EventListener>() { CallBase = true }.Object;`. Unfortunately, Mock syntax is noisy, this would have been `var listener = Substitute.ForPartsOf<EventListener>;` with NSubstitute, but it's not far off. I usually end up creating helper methods that return dynamic mocks. In fact, we already have one that returns ServiceContext.

https://github.com/microsoft/service-fabric-services-and-actors-dotnet/blob/048e83fd6f5bf057732036ab64354ab2ea03158b/test/Microsoft.ServiceFabric.TestFramework/IFuzzExtensions.cs#L35

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
olegsych: Oh, cool!

src/Microsoft.ServiceFabric.Diagnostics/Metrics/ServiceMeterProvider.cs
ilijagrbic: By the way, GitHub Copilot seems really great for this. It generates pretty much flawless docs - I just tweak minor things. :)

src/Microsoft.ServiceFabric.Diagnostics/Metrics/MeterBase.cs
ilijagrbic: I will switch to `IEnumerable` for now. This might be changes in the future if we try to optimize Record() method.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/TestServiceContext.cs
ilijagrbic: I completely forgot we have extensions for ServiceContext for IFuzz. I reworked the implementation to use Mock now, sadly I can't just use fuzzy.ServiceContext(), since I still need to to Asserts based on the contents of ServiceContext, but new implementation should be much better. :)

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Interop/IFabricMeter.cs
ilijagrbic: Will do, I will also update the FabricTelemetry.idl in our draft PR, since there were some changes needed for it as well.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64Meter1D.cs
ilijagrbic: I lowered the number of allocations in this method to 1 - when we create an array that is passed to the interop method.

As we agreed we will iterate on performance in the future, but for now resolving this conversation to not block the PR.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/ServiceMeterProvider.cs
olegsych: Interesting, I was about to respond saying that we should be careful about using Copilot because it can summarize current usage but has to infer the intent, which is what we want. But then realized that if a Copilot-generated comment accurately infersthe intent of a component it means that its design reveals the intent well. This could be an interesting way to test the design. 

I'm still not sure if we should keep Copilot-generated comments though. I noticed you've moved this class to the Implementation subnamespace, so a doc comment is not needed, particularly if the design is self-evident.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
olegsych: Here's a test I would use instead:

```C#
        public class Constructor : Int64Meter1DTest
        {
            [Fact]
            public void InvokesBaseWithGivenArguments()
            {
                IFabricMeter fabricMeter = Mock.Of<IFabricMeter>();
                IList<string> systemDimensions = fuzzy.List(fuzzy.String);

                MeterBase sut = new Int64Meter1D(fabricMeter, systemDimensions);

                Assert.Same(fabricMeter, sut.Field<IFabricMeter>().Value);
                Assert.Same(systemDimensions, sut.Field<IEnumerable<string>>().Value);
            }
        }
```

It is comparable in size to the inheritance test, however it also verifies that the constructor invokes the base constructor and that it passes the arguments to it. The inheritance verification is still performed, albeit implicitly. Notice that it helped me to flush out the difference between the Int64Meter1D constructor which takes IList<string> and the MeterBase constructor which takes IEnumerable<string>.

And because a sut instance is needed by other tests, I typically reduce duplication by structuring the code like this:

```C#
    public abstract class Int64Meter1DTest
    {
        readonly IMeter1D<long> sut;

        // Constructor parameters
        readonly IFabricMeter fabricMeter = Mock.Of<IFabricMeter>();
        readonly IList<string> systemDimensions = fuzzy.List(fuzzy.String);

        // Test fixture
        static readonly IFuzz fuzzy = new RandomFuzz(Environment.TickCount);

        public Int64Meter1DTest() => sut = new Int64Meter1D(fabricMeter, systemDimensions);

        public class Constructor : Int64Meter1DTest
        {
            [Fact]
            public void InvokesBaseWithGivenArguments()
            {
                var meter = (MeterBase)sut;
                Assert.Same(fabricMeter, meter.Field<IFabricMeter>().Value);
                Assert.Same(systemDimensions, meter.Field<IEnumerable<string>>().Value);
            }
        }
```

The constructor test shrinks to just 3 lines, and could shrink to just 2 if we forego the base class typecast. And the sut field along with the fields storing the constructor parameters are now available to for use in the Record tests, helping to reduce the amount of setup/arrange logic there. 

Also notice the choice of the sut field type. This is a form of an inheritance test that verifies that Int64Meter1D implements IMeter1D<long>. If you TDD and write code like `IMeter1D<long> sut = new Int64Meter1D(foo, bar)` before the Int64Meter1D class exists, VS will offer you an option to generate it, which works pretty well.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64Meter1D.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        /// Records a long integer telemetry value with one additional dimension.
+        /// The metric will be recorded with system dimensions plus the specified additional dimension.
+        /// </summary>
+        /// <param name="value">The integer value to record.</param>
+        /// <param name="dimension1">The additional dimension value to record with the metric.</param>
+        public void Record(long value, string dimension1)
+        {
+            var systemDimensionsList = (List<string>)systemDimensionValues;
```
ilijagrbic: I've cast this `IEnumerable` to list in order to use `CopyTo` method into an array. I know we are assuming the underlying implementation of this structure now, so this might be a problem. I just don't think it is a huge deal now when we will probably rework and iterate on this code. If you think this is a problem, I can keep it outside of the PR and use a simpler implementation.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64Meter.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// Records integer telemetry values with system dimensions.
+    /// </summary>
+    sealed class Int64Meter : MeterBase, IMeter<long>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Int64Meter"/> class.
+        /// </summary>
+        /// <param name="fabricMeter">The native fabric meter implementation.</param>
+        /// <param name="systemDimensionValues">System dimension values that will be included with all recorded metrics.</param>
+        internal Int64Meter(IFabricMeter fabricMeter, IList<string> systemDimensionValues) : base(fabricMeter, systemDimensionValues) { }
```
olegsych: Don't forget to update constructors of classes inheriting from MeterBase to take IEnumerable<string>

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/MeterBase.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
+{
+    class MeterBase
```
olegsych: Let's make this class abstract and drop the term Base. https://en.wikipedia.org/wiki/All_your_base_are_belong_to_us :)

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/NativeTelemetry.cs
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        [LibraryImport("FabricTelemetry")]
+        internal static
+#if NET
+        partial
+#else
+        extern
+#endif
+            HRESULT FabricCreateMeterProvider([MarshalAs(UnmanagedType.Interface)] out IFabricMeterProvider meterProvider);
```
olegsych: Could you make the indentation consistent?

Let's remove the `[MarshalAs(UnmanagedType.Interface)]`. They are redundant for interfaces. The only time when `UnmanagedType.Interface` needs to be specified explicitly is when marshalling delegates as com objects instead of function pointers.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/IFabricMeter.cs
```diff
@@ -0,0 +1,27 @@
...
+#endif
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
+{
+    [GeneratedComInterface]
+    [Guid("A0D80970-C062-4C59-A3CA-7AEEB901B49C")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    partial interface IFabricMeter
+    {
+        [MethodImpl(MethodImplOptions.InternalCall, MethodCodeType = MethodCodeType.Runtime)]
```
olegsych: Could you remove `[MethodImpl]` attributes? They shouldn't be needed.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
+{
+    [GeneratedComInterface]
+    [Guid("15AD37D2-F641-4188-824B-0D68CB4F6C17")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    partial interface IFabricMeterProvider
+    {
+        [MethodImpl(MethodImplOptions.InternalCall, MethodCodeType = MethodCodeType.Runtime)]
+        [return: MarshalAs(UnmanagedType.Interface)]
```
olegsych: Could you remove `[MarshalAs(UnmanagedType.Interface)] and `[MethodImpl]` attributes? See my comments in other interop files for details.

olegsych (changes requested):
A couple more thoughts.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64MeterProviderTest.cs
```diff
@@ -0,0 +1,143 @@
...
+            public CreateMeter()
+            {
+                sut = new Int64MeterProvider(serviceContext);
+                systemDimensionsNames = sut.Private().Field<IList<string>>().Value;
+                systemDimensionsValues = sut.Protected().Field<IList<string>>().Value;
+                sut.Field<IFabricMeterProvider>().Set(fabricMeterProvider);
+            }
+
+            [Fact]
+            public void ShouldCreateMeterWithCorrectSystemDimensions()
```
ilijagrbic: TODO: Drop "Should"

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/IFabricMeter.cs
ilijagrbic: This is what was left when I copied the generated code from the FabricTelemetry.tlb, I will remove it.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
ilijagrbic: Thanks for a detailed example and explanation, this was all insightful. I have improved these tests according to your suggestions.

.github/workflows/build.yml
```diff
@@ -24,6 +24,12 @@ jobs:
       - name: Disable strong name verification for testing
         shell: powershell
         run: .\SkipStrongName.ps1
+      - name: Uninstall SF SDK assemblies from GAC
+        shell: powershell
+        # Run command in VS Dev shell
+        run: |
+          & "${env:ProgramFiles}\Microsoft Visual Studio\2022\Enterprise\Common7\Tools\Launch-VsDevShell.ps1" -SkipAutomaticLocation 
+          .\UninstallSfSdkFromGac.ps1
```
ilijagrbic: Hey @olegsych , just a fyi on why these changes are needed. This is basically the same issue Pavle had in his PR. And it manifested in this PR because I enabled `Microsoft.ServiceFabric.Actors.Wcf.Tests` by fixing the typo (missing ,) in `SkipStrongName.ps1`

ilijagrbic: I talked to Pavle, and we decided to keep `Actors.Wcf` tests disabled in this PR and that he will enable them in his PR. 
We this it is better to keep these changes in a related PR, since they don't have much to do with Telemetry. Apologies for multiple comments. :)

SkipStrongName.ps1
```diff
@@ -13,9 +13,9 @@ $assemblies = "FabActUtil",
               "Microsoft.ServiceFabric.Services.Remoting.Tests",
               "Microsoft.ServiceFabric.Diagnostics",
               "Microsoft.ServiceFabric.Diagnostics.Tests",
-              "Microsoft.ServiceFabric.Services.Wcf.Tests"
-              "Microsoft.ServiceFabric.Actors.Wcf.Tests"
-
+              "Microsoft.ServiceFabric.Services.Wcf.Tests",
+              "Microsoft.ServiceFabric.TestFramework"
+
```
olegsych: Unnecessary white space?

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/IFabricMeter.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
+{
+    [GeneratedComInterface]
+    [Guid("A0D80970-C062-4C59-A3CA-7AEEB901B49C")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    partial interface IFabricMeter
+    {
+        void Record(long value, uint count, [MarshalAs(UnmanagedType.LPArray, ArraySubType = UnmanagedType.LPWStr)] string[] dimensionValues);
+    }
+
```
olegsych: Unnecessary blank line?

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
```diff
@@ -0,0 +1,24 @@
...
+#endif
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
+{
+    [GeneratedComInterface]
+    [Guid("15AD37D2-F641-4188-824B-0D68CB4F6C17")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    partial interface IFabricMeterProvider
+    {
+        IFabricMeter CreateMeter([MarshalAs(UnmanagedType.LPWStr)] string metricNamespace, [MarshalAs(UnmanagedType.LPWStr)] string name, uint count, [MarshalAs(UnmanagedType.LPArray, ArraySubType = UnmanagedType.LPWStr)] string[] dimensionNames);
```
olegsych, line 22: Does this build without the `[MarshalAs(UnmanagedType.LPWStr)]`? I know `MarshalAsAttribute` is required for arrays but was hoping we could do without it for strings.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    abstract class Meter
+    {
+        protected readonly IEnumerable<string> systemDimensionValues;
+        protected readonly IFabricMeter fabricMeter;
+
+        internal Meter(IFabricMeter fabricMeter, IEnumerable<string> systemDimensionValues)
+        {
+            this.fabricMeter = fabricMeter ?? throw new ArgumentNullException(nameof(fabricMeter));
+            this.systemDimensionValues = systemDimensionValues ?? new List<string>();
```
olegsych: Accepting null systemDimensionValues looks odd. I don't see any Meter descendants with constructors that allow null systemDimensionValues. We should fail fast and throw ArgumentNullException unless there is a specific reason for system dimensions to be empty. Even then I think it's better to require a non-null instance that the caller can provide as Enumerable.Empty<String>().

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/NativeTelemetry.cs
```diff
@@ -0,0 +1,34 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System.Runtime.InteropServices;
+using HRESULT = System.Int32;
+#if NET
+using System.Runtime.InteropServices.Marshalling;
```
olegsych: I don't think we need the System.Runtime.InteropServices.Marshalling namespace because the LibraryImportAttribute is in the System.Runtime.InteropServices.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/ServiceMeterProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+            systemDimensionValues.Add(serviceContext.PartitionId.ToString());
+            systemDimensionValues.Add(serviceContext.ServiceTypeName);
+            systemDimensionValues.Add(serviceContext.ServiceName.ToString());
+            systemDimensionValues.Add(serviceContext.CodePackageActivationContext.ApplicationName);
+            systemDimensionValues.Add(serviceContext.CodePackageActivationContext.ApplicationTypeName);
+        }
+
+        protected IFabricMeter CreateNativeMeter(string metricNamespace, string metricName, IList<string> additionalDimensions)
+        {
+            string[] allDimensionNamesArray = new string[systemDimensionNames.Count + additionalDimensions.Count];
```
olegsych: Let's try using immutable string collection interfaces as method parameters and fields. This is a design issue in my mind and I would like to tackle it upfront. So let's make additionalDimensions and systemDimensionNames an IEnumerable<string>.  I suspect you may have chosen an IList<string> because it has a CopyTo method. But you can use List<T>.AddRange(IEnumerable<T>) instead. Allocating List<T> will be as efficient as an array if you construct it with the initial size parameter.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64Meter.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Linq;
+using Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics
+{
+    /// <summary>
+    /// Implementation of <see cref="IMeter{TValueType}"/> for long integer metrics.
+    /// Records integer telemetry values with system dimensions.
+    /// </summary>
+    sealed class Int64Meter : Meter, IMeter<long>
```
olegsych: I would move the IMeter implementations to the Implementation namespace. They aren't meant to be instantiated by the user code directly, only the IMeterProvider implementations are. We can also lose the doc comments in that case.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64Meter1D.cs
olegsych: That's fine for now. By the way, you can also use https://learn.microsoft.com/en-us/dotnet/api/system.array.copyto

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/ServiceMeterProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Collections.Generic;
+using System.Fabric;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
+{
+    abstract class ServiceMeterProvider<TValueType> : IMeterProvider<TValueType>
+    {
+        readonly IList<string> systemDimensionNames = new List<string>
+        {
+            "ReplicaOrInstanceId",
```
olegsych: This may be a nit, but in case there are spelling errors I can't spot now or in the future, I would use `nameof(ServiceContext.ReplicaOrInstanceId)` instead of hardcoding a string literal.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64MeterProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// <summary>
+    /// Implementation of <see cref="ServiceMeterProvider{TValueType}"/> for long integer metrics.
+    /// Creates meters for recording long integer telemetry values with various dimension configurations.
+    /// </summary>
+    sealed class Int64MeterProvider : ServiceMeterProvider<long>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Int64MeterProvider"/> class.
+        /// </summary>
+        /// <param name="serviceContext">The Service Fabric service context that provides runtime information about the service instance.</param>
```
olegsych: This looks like something Copilot would generate based on the class information. The important thing to call out here would be that contextual dimensions are extracted from the serviceContext.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64MeterProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// </summary>
+        /// <param name="metricNamespace">The namespace for the metric.</param>
+        /// <param name="name">The name of the metric.</param>
+        /// <param name="dimension1Name">The name of the first dimension.</param>
+        /// <param name="dimension2Name">The name of the second dimension.</param>
+        /// <param name="dimension3Name">The name of the third dimension.</param>
+        /// <returns>A meter instance for recording long integer values with three additional dimensions.</returns>
+        public override IMeter3D<long> CreateMeter(string metricNamespace, string name, string dimension1Name, string dimension2Name, string dimension3Name)
+        {
+            return new Int64Meter3D(CreateNativeMeter(metricNamespace, name, new List<string>() { dimension1Name, dimension2Name, dimension3Name }), systemDimensionValues);
```
olegsych: One nice thing about taking IEnumerable<T> as a parameter is that you can shorten initializers to something like `new[] { dimension1Name, dimension2Name, dimension3Name }`

src/Microsoft.ServiceFabric.Diagnostics/Metrics/NullMeter3D.cs
```diff
@@ -0,0 +1,12 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics
+{
+    sealed class NullMeter3D<TValueType> : IMeter3D<TValueType>
```
olegsych: Thanks for placing classes in separate files!
I would move the NullMeter* classes to the Implementation namespace as well - only NullMeterProvider should be used by our internal customers.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/NullMeterProvider.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.ServiceFabric.Diagnostics.Metrics
 {
-    internal class NullMeterProvider<TValueType> : IMeterProvider<TValueType>
+    sealed class NullMeterProvider<TValueType> : IMeterProvider<TValueType>
```
olegsych, line 10: Thanks for removing the redundant internal visibility and sealing these classes!

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Implementation/MeterTest.cs
```diff
@@ -0,0 +1,54 @@
...
+                Assert.NotNull(meter.Field<IEnumerable<string>>().Value);
+                Assert.Empty(meter.Field<IEnumerable<string>>().Value);
+            }
+
+            [Fact]
+            public void SetsSystemDimensionAndMeterValuesWhenProvided()
+            {
+                var expectedValues = fuzzy.List(() => fuzzy.String());
+
+                Meter meter = new MeterImplementation(fabricMeter, expectedValues);
```
olegsych: I have a feeling we'll need to have a Record method in the Meter base class because there is too much low-level, hard-to-write code in each concrete class, particularly once we add TimeSpan meters. At that point, it would be better to test the initialization logic through the dimension values passed to the IFabricMeter.Record call.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Implementation/ServiceMeterProviderTest.cs
```diff
@@ -0,0 +1,111 @@
...
+        readonly string testApplicationName = fuzzy.String();
+        readonly string testApplicationTypeName = fuzzy.String();
+        readonly long replicaId = fuzzy.Int64();
+        readonly Guid testPartitionId = Guid.NewGuid();
+
+        readonly ServiceContext serviceContext;
+
+        public ServiceMeterProviderTest()
+        {
+            var codePackageActivationContext = Mock.Of<ICodePackageActivationContext>();
```
olegsych: Why can't we use `fuzzy.ServiceContext()` extension method from the TestFramework project here? The only reason I can think of is that we use `testApplicationName`, etc. in the assertions below. However. we should also be able to use `serviceContext.CodePackageActivationContext.ApplicationName` with the mocked objects.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Implementation/ServiceMeterProviderTest.cs
```diff
@@ -0,0 +1,111 @@
...
+            [Fact]
+            public void ShouldHaveIFabricMeterNativeInterop()
+            {
+                Func<IFabricMeterProvider> expected = typeof(NativeTelemetry).Method<Func<IFabricMeterProvider>>(nameof(NativeTelemetry.FabricCreateMeterProvider));
+                Func<IFabricMeterProvider> actual = typeof(ServiceMeterProvider<int>).Field<Func<IFabricMeterProvider>>();
+                Assert.Equal(expected, actual);
+            }
+        }
+
+        class TestMeterProvider<TValueType> : ServiceMeterProvider<TValueType>
```
olegsych, line 67: I like using dynamic mocking to test abstract classes, particularly when their constructors have strongly typed arguments that would be difficult to mix up, just like in this case. Creation of the sut becomes a little more obscure `ServiceMeterProvider<int> sut = new Mock<ServiceMeterProvider<int>>(serviceContext) { CallBase = true }.Object`, but it can usually be limited to just one or two lines initializing a protected field or return from a helper creation method. Not as obvious as the explicit code, but in many cases, a lot more compact.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        public class Record : Int64Meter1DTest
+        {
+            readonly string customDimension1 = fuzzy.String();
+            readonly long value = fuzzy.Int64();
+
+            [Fact]
+            public void CallsFabricMeterRecordWithMultipleSystemDimensions()
+            {
+                // Arrange
```
olegsych: I think we can omit the a/a/a comments when the code paragraphs are small and separated.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
```diff
@@ -0,0 +1,74 @@
...
+
+                // Act
+                sut.Record(value, customDimension1);
+
+                // Assert
+                Mock.Get(fabricMeter).Verify(m => m.Record(value, (uint)expectedArray.Length, It.Is<string[]>(arr => arr.SequenceEqual(expectedArray))), Times.Once);
+            }
+
+            [Fact]
+            public void CallsFabricMeterRecordWithNoSystemDimensions()
```
olegsych: Can this test ever fail independently of the other? Ideally, each test should fail independently and pinpoint the problem with the information in its name. So, if it doesn't fail independently, it is likely redundant and would be best to remove. This goes against the common test-later practices when we aren't sure which tests are needed and which aren't. In this case, however, the system under test is small and well defined, so we should be able to verify if this test is needed by trying to sabotage the product code. If we can't modify the product code in a way that makes this test fail without the other, it's redundant.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter3DTest.cs
```diff
@@ -0,0 +1,76 @@
...
+                Assert.Same(systemDimensions, meter.Field<IEnumerable<string>>().Value);
+            }
+        }
+
+        public class Record : Int64Meter3DTest
+        {
+            readonly string customDimension1 = fuzzy.String();
+            readonly string customDimension2 = fuzzy.String();
+            readonly string customDimension3 = fuzzy.String();
+            readonly long value = fuzzy.Int64();
```
olegsych: I like fields like this because they help reader understand the method. It would be great to place them in the order of the method parameters.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64MeterProviderTest.cs
```diff
@@ -0,0 +1,143 @@
...
+
+                Assert.Equal(systemDimensionsValues, ((Int64Meter3D)meter3D).Field<IEnumerable<string>>().Value);
+            }
+
+            [Fact]
+            public void CreatesNativeMeterWithCorrectDimensions()
+            {
+                var meter = sut.CreateMeter(testNamespace, testMetric);
+
+                var expectedArray = systemDimensionsNames.ToArray();
```
olegsych: The `expectedArray` variable is redundant here and could be inlined. Where something like it is needed below let's use more specific terms, like  `combinedDimensions` instead of `array`. The type of the variable should make it obvious that it's an array.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64MeterProviderTest.cs
```diff
@@ -0,0 +1,143 @@
...
+                Assert.NotNull(meter2D);
+                Assert.IsType<Int64Meter2D>(meter2D);
+
+                Assert.Equal(systemDimensionsValues, ((Int64Meter2D)meter2D).Field<IEnumerable<string>>().Value);
+            }
+
+            [Fact]
+            public void CreatesMeter3DWithCorrectSystemDimensions()
+            {
+                var meter3D = sut.CreateMeter(testNamespace, testMetric, testDimension1, testDimension2, testDimension3);
```
olegsych: Could you specify variable type instead of var when it doesn't appear explicitly in the initializer?

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64MeterProviderTest.cs
```diff
@@ -0,0 +1,143 @@
...
+                Assert.Equal(systemDimensionsValues, ((Int64Meter3D)meter3D).Field<IEnumerable<string>>().Value);
+            }
+
+            [Fact]
+            public void CreatesNativeMeterWithCorrectDimensions()
+            {
+                var meter = sut.CreateMeter(testNamespace, testMetric);
+
+                var expectedArray = systemDimensionsNames.ToArray();
+                Mock.Get(fabricMeterProvider).Verify(x => x.CreateMeter(testNamespace, testMetric, (uint)expectedArray.Length, It.Is<string[]>(arr => arr.SequenceEqual(expectedArray))), Times.Once);
```
olegsych: These test is not quite complete because it doesn't verify that the meter instance was created with the correct fabric meter and with correct system dimension values. 

I would create an IFabricMeter mock instance and setup the fabricMeterProvider mock to return this instance when it receives an expected CreateMeter call. Then you could assert that the Meter instance returned by the sut.CreateMeter was created with the IFabricMeter mock instance created earlier. 

The system dimension values would need to be verified in additional assert.

olegsych (changes requested):
Thanks @ilijagrbic I have a couple of minor concerns about low-level design and test implementation, but think the PR is close. We should be able to merge it tomorrow.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/Meter.cs
ilijagrbic: I was thinking about a possible future case where we emit metrics outside of a service context. However, you are right, even in that case we can handle this on MeterProvider level and still supply an empty collection.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/ServiceMeterProvider.cs
ilijagrbic: I will change the parameter type. I wanted to avoid allocating a new List in this method at all, since the entire purpose of such list would be to convert it to an Array. So, I decided to allocate an array right away and copy the contents into it so minimize number of heap allocations. However now when I think about this again - this path doesn't seem performance critical (as opposed to Record() methods), So I will refactor this logic to be straight forward.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/ServiceMeterProvider.cs
ilijagrbic: Good idea, I will fix this.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64MeterProvider.cs
ilijagrbic: Thanks for spotting this. I will rewrite this comment.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64MeterProvider.cs
ilijagrbic: This is quite cool. I was not familiar with these details regarding Arrays/IEnumerables, so it's nice to learn about it. :)

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Implementation/ServiceMeterProviderTest.cs
ilijagrbic: Yeah, for some reason I thought these were inaccessible, I will simplify this.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Implementation/ServiceMeterProviderTest.cs
ilijagrbic: Yeah, these are good points. I guess it comes down to preference. I kind of prefer having explicit code in cases where the abstract class is short, since the additional length doesn't impact the tests themselves, but rather this implementation class, that can be placed at the end of the file, and all of this makes the test just a little bit more obvious. I will keep it as is in this case, but good tip. :)

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
ilijagrbic: Yup, in the beginning these tests were much longer, and the comments remained as a relic of the initial implementation. I will remove them.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64Meter1DTest.cs
ilijagrbic: One way this test could fail is if we for example throw an exception in constructor for Int64Meter when we accept an empty list. I wanted to explicitly document we support empty system dimensions by this test. 

Another reason is that if I insatiate system dimensions with `fuzzy.List(() => fuzzy.String())` (instead of  `fuzzy.List(() => fuzzy.String(), Count.Min(1))`), some of the time we would test the case with an empty list, some of the time we won't. This would be an issue for example if someone adds a validation to not support empty system dimensions. The test would then start failing randomly, and it would not be clear what original author intended. 

Now that I'm writing this and rethinking it - a constructor test is better for this use-case. I will replace these tests with a constructor test in the base class.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64MeterProviderTest.cs
ilijagrbic: I thought this is better since inlining would arguable be less readable in this case - we would have two `systemDimensionsNames.ToArray()` and it would make the entire line longer. I agree on the naming though - I will rename these variables.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64MeterProviderTest.cs
ilijagrbic: Agreed, will fix.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
ilijagrbic: Good question. We do get build errors if we omit this attribute. I think it is due to string being a bit more complicated for marshaling, since there are multiple ways to handle them.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Int64MeterProviderTest.cs
ilijagrbic: Good suggestion, I will add this.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Implementation/MeterTest.cs
ilijagrbic: Yup, I agree with this. I'm planning to extract the implementation of these methods into the base class in a follow up PR (probably for the TimeSpan meters or when we work on optimizing the interop call). 
Also, I agree on the test logic, that's pretty much how we test `Record()` method for the current `Int64Meters`, by verifying calls to the `IFabricMeter.Record` methods.

We still have multiple heap allocation in the `Record()` methods and I'm not sure how to tackle this just by using `IEnumerable` for `systemDimensionValues`. The only way to convert the `IEnumerable` to an array is by using `ToArray()` method. While using `CopyTo` (either from `List` or from `Array`) is not possible without casting an IEnumerable into the respective types. The only other way I see is to manually iterate over the `IEnumerable` and copy elements into an array, instead of using `CopyTo`. This way we will only have one heap allocation for the String array we want to pass into the interop call. Unless there is a different way of achieving this with `IEnumerable` (or `IReadOnlyList`/`IReadOnlyCollection`) - but I don't see any method similar to `CopyTo` in the docs.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/ServiceMeterProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+using System;
+using System.Collections.Generic;
+using System.Fabric;
+using System.Linq;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
+{
+    abstract class ServiceMeterProvider<TValueType> : IMeterProvider<TValueType>
+    {
+        readonly IEnumerable<string> systemDimensionNames = new[]
```
gobradovic: This might be static. But if this class is not instantiated too much, it is not that important to change it.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/Int64Meter2D.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    sealed class Int64Meter2D : Meter, IMeter2D<long>
+    {
+        internal Int64Meter2D(IFabricMeter fabricMeter, IEnumerable<string> systemDimensionValues) : base(fabricMeter, systemDimensionValues) { }
+
+        public void Record(long value, string dimension1, string dimension2)
+        {
+            var allDimensionArray = new string[systemDimensionValues.Count() + 2];
+
+            systemDimensionValues.ToArray().CopyTo(allDimensionArray, 0);
```
gobradovic: Both `ToArray()` and `CopyTo()` copy the collection, so we effectively copy it twice here. I would consider converting this collection to array in constructor (in Meter superclass) and keeping it as an array, so that we don't have to copy it twice on each invocation.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Int64MeterProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// <summary>
+    /// Implementation of <see cref="ServiceMeterProvider{TValueType}"/> for long integer metrics.
+    /// Creates meters for recording long integer telemetry values with various dimension configurations.
+    /// </summary>
+    sealed class Int64MeterProvider : ServiceMeterProvider<long>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Int64MeterProvider"/> class.
+        /// </summary>
+        /// <param name="serviceContext">The Service Fabric ServiceContext provides system dimensions used by IMeter implementations to emitt default metrics.</param>
```
gobradovic: nit: typo (emit).

gobradovic (approved):

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/ServiceMeterProvider.cs
ilijagrbic: Nice catch, there is no reason for this not to be static. I will fix it.

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/Int64Meter2D.cs
ilijagrbic: Good suggestion, I will do that. :)

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Implementation/MeterTest.cs
ilijagrbic: I have talked to Goran when he was reviewing this PR, and regarding the `Record()` method and array copy, I decided to keep `IEnumerable` in the constructor API for the `Meter` superclass. However, I store it as an array in the class by converting it from `IEnumerable` to an array, this was we don't do this conversion during `Record()` method calls.

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Implementation/NullMeterProviderTest.cs
```diff
@@ -5,7 +5,7 @@
 
 using Xunit;
 
-namespace Microsoft.ServiceFabric.Diagnostics.Metrics
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
```
olegsych: Was this intentional? I think NullMeterProvider is still in the Metrics namespace.

olegsych (approved):
Love it! :shipit:

test/Microsoft.ServiceFabric.Diagnostics.Tests/Metrics/Implementation/NullMeterProviderTest.cs
ilijagrbic: Yeah, this was not intentional, perhaps it happened while I was moving files around. I will revert this. :)

gobradovic (approved):

PR #462, opened 2025-08-18
Renormalize line endings

olegsych:
- Create .gitattributes with instructions for Git to assume all file types are text, store LF line endings in the repo and check them out as LF on Linux and CRLF on Windows. This helps to prevent storing files with mixed or CRLF line endings in the Git index. Linux tools expect LF line endings and may have difficulty with CRLFs. On Windows, mixed line endings are reported by Visual Studio, which by default prompts you to normalize line endings and leading to unnecessary churn in Git history and unnecessary merge conflicts.
- Re-normalize line endings for all files in the repository.


ilijagrbic (approved):

plave0 (approved):
No text files have been found in the index which have mixed or CRLF line endings. Only binary files are marked as `non-text`.

Image

PR #465, opened 2025-08-27
Add test for checking whether reminder callbacks are fired consecutively

plave0:
`ReminderIsRegisteredAndFiredMultipleTimes` has been added to `Microsoft.ServiceFabric.Actors.IntegrationTests`. 

This functionality was originally tested internally, and now we are moving it to this repo.


plave0:
@olegsych @gobradovic let's think about whether a test implemented this way is useful to us, or whether we can implement it in such a way that it is useful. 

I think that for now it is okay to keep it, but I think it will eventually be covered by more detailed unit tests.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Reminders/ReminderRegistrationTest.cs
```diff
@@ -86,95 +93,221 @@ public Task ReceiveReminderAsync(string reminderName, byte[] state, TimeSpan due
...
-                await Task.Delay(TimeSpan.FromSeconds(2));
-
-                Assert.Equal(expectedCallbackInvocationCounter, reminderCallbackInvocationCounter);
-                Assert.Equal(expectedReminderName, actualReminderName);
-                Assert.Equal(expectedActorMethodName, actualActorMethodName);
-                Assert.Equal(expectedActorCallType, actualActorCallType);
-                Assert.Equal(expectedState, actualState);
-                Assert.Equal(expectedDueTime, actualDueTime);
-                Assert.Equal(expectedPeriod, actualPeriod);
+                    await Task.Delay(reminderDueTime + allowedTimeVariation);
```
olegsych, line 218: Here's the technique we discussed in scrum last week that could be used to make testing of delays deterministic.

```C#
var reminderReceived = new ManualResetEventSlim();
Action<ReminderCallbackInfo> receiveReminderCallback = 
  (reminderCallbackInfo) =>
  {
      reminderReceived.Set();
  };

// Register reminder ....

Assert.True(reminderReceived.Wait(timeout));

// Assert....
```

This way you can shorten the time spans and eliminate the time variation. It does assume that the underlying implementation uses a mockable timer abstraction and doesn't really verify that the delay is of the expected duration.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Reminders/ReminderRegistrationTest.cs
```diff
@@ -56,13 +56,20 @@ public ReminderCallbackInfo(string reminderName, byte[] state, TimeSpan dueTime,
...
 
         class TestableActor : Actor, ITestableActor, IRemindable
         {
             ActorMethodContext currentMethodContext;
             Action<ReminderCallbackInfo> receiveReminderCallback;
 
+            public TestableActor(ActorService actorService, ActorId actorId)
+                : base(actorService, actorId)
+            {
+                this.receiveReminderCallback = (reminderCallbackInfo) => { return; };
```
olegsych, line 70: I wonder if we could make the `ReceiveReminderAsync` method virtual and use the TestableActor as a dynamic mock.

olegsych (approved):
Looks good. I feel that the test setup logic is getting to more and more complex and at some point we should try making another pass at simplification. I left some thoughts in the comments. 

@ilijagrbic incorporated a lot of dynamic mocking in telemetry tests. Perhaps he has more suggestions. 

:shipit:

gobradovic (approved):

PR #467, opened 2025-09-04
Implement integration tests for ActorService.GetReminderAsync

plave0:
- Adding integration tests for ActorSerivce.GetReminderAsync
- Tests implemented in this PR reflect the testing done internally in the `ReminderEnumeration` family of tests ([workload implementation](https://dev.azure.com/msazure/One/_git/WindowsFabric?path=%2Fsrc%2Ftest%2Fcurrent%2Fsource%2Fsdk%2FClientWorkload%2FActorReminderEnumerationClientWorkload.cs&_a=contents&version=GBdevelop) which test `ActorService.GetRemindersAsync()` method
- The `GetRemindersAsync` test class is structured as follows:
  - WhenNoReminderIsRegistered
  - WhenRemindersAreRegistered
    - WithCancellationToken
    - WhenNoChangesBetweenResults
    -  WhenChangesBetweenResults
       - WhenChangingCurrentPage
       - WhenChangingUpcomingPage
    -  WhenActorDeleted
- Adopting new way to organize integration tests: Separation by tested class is done using namespaces. Ex. `Microsoft.ServiceFabric.Actors.ActorIntegrationTests` will contain integration tests for `Actor` class. Classes in this namespace will separate tests by the method they test. Ex. class `GetRemindersAsync` in `Microsoft.ServiceFabric.Actors.ActorServiceIntegrationTests` will contain tests for `ActorService.GetReminderAsync()`
- Tests from `ReminderRegistrationTests.cs` have been reorganized to follow the organization scheme explain above.


test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorServiceTests.cs
```diff
@@ -0,0 +1,471 @@
...
+
+                protected void ReadReminderPage(
+                    ReminderPagedResult<KeyValuePair<ActorId, List<ActorReminderState>>> resultPage,
+                    Dictionary<ActorId, List<IActorReminder>> output
+                )
+                {
+                    foreach (var kvp in resultPage.Items)
+                    {
+                       if (!output.ContainsKey(kvp.Key))
+                            output[kvp.Key] = new List<IActorReminder>();
```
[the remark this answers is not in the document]
plave0: These changes cause a build error

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Reminders/ReminderRegistrationTest.cs
```diff
@@ -10,31 +10,8 @@
...
-                actorServiceSettings);
-
-            IStatefulUserServiceReplica statefulServiceReplica = actorService;
-            await statefulServiceReplica.OnOpenAsync(ReplicaOpenMode.New, CancellationToken.None);
-            await statefulServiceReplica.OnChangeRoleAsync(ReplicaRole.Primary, CancellationToken.None);
-            await statefulServiceReplica.RunAsync(CancellationToken.None);
-
-            return actorService;
-        }
-
```
ilijagrbic, line 37: This is out of the scope for this PR, but I will leave some comments regarding this file.

I think it would be good to extract `ReminderCallbackInfo`, `ITestableActor` and `TestableActor` into a separate file, since this file is already quite long and it will get harder to navigate as the number of tests grows.

This is really a nit, and not needed to fix in this PR, but worth mentioning, I think. For example, let's take this test name:
`RegisterReminderAsync.WithNewReminderName.ReminderParametersPersistBetweenRegisteringAndReceivingReminder`
When I name tests, I avoid adding unnecessary references to the same concept whenever I can, so that we have shorter and more readable names. So for example, in this case it is clear we are talking about a remainder and we can shorten the name of the test case to: `PersistsParametersBetweenRegisteringAndReceiving`, and when read together with the class name (`RegisterReminderAsync.WithNewReminderName`) it is quite clear what we are testing. :)

test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorServiceTests.cs
```diff
@@ -0,0 +1,480 @@
...
+using Microsoft.ServiceFabric.Actors.Query;
+using Microsoft.ServiceFabric.Actors.Runtime;
+using Microsoft.ServiceFabric.Services.Runtime;
+using Moq;
+using Xunit;
+using Xunit.Internal;
+
+namespace Microsoft.ServiceFabric.Actors
+{
+    public class ActorServiceIntegrationTest
```
ilijagrbic: I would prefer the name of the class to be the same as the name of the file. Also, right now we have `ActorService` file, `ActorServiceIntegrationTest` class and `ActorIntegrationTest` class which makes it hard to understand what is responsible for what.

As far as I can understand the purpose of the `ActorServiceIntegrationTest` is to share the `GetActorService` logic between different actor tests. If this is the case, I would extract `ActorServiceIntegrationTest` into a separate file and then rename this file to `GetRemindersAsync` (to match the class name). I would also separate the `RegisterReminderAsync` and `UnregisterReminderAsync` into different files in order to have only 1 test class per test process.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorServiceTests.cs
```diff
@@ -0,0 +1,480 @@
...
+            {
+                public TestActor(ActorService actorService, ActorId actorId) : base(actorService, actorId)
+                {
+                }
+            }
+
+            public class WhenNoReminderIsRegistered : GetRemindersAsync
+            {
+                [Fact]
+                public async Task ReturnEmptyResultIfNoRemindersAreRegistered()
```
ilijagrbic: Consider shortening the test case name from `ReturnEmptyResultIfNoRemindersAreRegistered` to `ReturnEmptyResult` since it is clear that remainder is not registered from the class name (`WhenNoReminderIsRegistered`). Another alternative is to keep the test case name, but drop the `WhenNoReminderIsRegistered` as it might not be needed in the test hierarchy (there is only one test case in it).

test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorServiceTests.cs
```diff
@@ -0,0 +1,480 @@
...
+                {
+                    this.serviceStateFixture = serviceStateFixture;
+                }
+
+                public class CancellationTokenIsNotNull : WhenRemindersAreRegistered
+                {
+                    public CancellationTokenIsNotNull(ServiceStateFixture serviceStateFixture) : base(serviceStateFixture) { }
+
+                    [Fact]
+                    public async Task ThrowsWhenCancellationTokenIsCanceled()
```
ilijagrbic: nit: Consider improving the naming of this class/test case, for example:
`WhenRemindersAreRegistered.`**`WithCancelationToken.ThrowsWhenCanceled `**

test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorServiceTests.cs
```diff
@@ -0,0 +1,480 @@
...
+
+                        // Act
+                        cts.Cancel();
+
+                        //Assert
+                        await Assert.ThrowsAsync<OperationCanceledException>(() => actorService.GetRemindersAsync(null, null, cts.Token));
+                    }
+                }
+
+                public class WhenNoChangesAreMadeToTheRemindersBetweenResults : WhenRemindersAreRegistered
```
ilijagrbic: nit: Consider improving the naming of this class/test case, for example:
`WhenRemindersAreRegistered.`**`WithNoChangesBetweenResults...`**

test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorServiceTests.cs
```diff
@@ -0,0 +1,480 @@
...
+                            IActorService actorService = await GetActorService<TestActor>(actorStateProvider: actorStateProviderWithReminders);
+                            var expectedQueryResult = registeredRemindersPerActor;
+
+                            // Act
+                            Dictionary<ActorId, List<IActorReminder>> actualQueryResult = await QueryReminders(actorService, null);
+
+                            // Assert
+                            Assert.True(ReminderDictionariesAreEqual(expectedQueryResult, actualQueryResult));
+                        }
+                    }
```
ilijagrbic: nit: Consider collapsing these two classes into a single class, since this hierarchy level is probably not needed.
For example:
`WhenRemindersAreRegistered.`**`WithNoChangesBetweenResults.SameRemaindersReturned`**
`WhenRemindersAreRegistered.`**`WithNoChangesBetweenResults.SameRemaindersPerActorReturned`**

test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorServiceTests.cs
```diff
@@ -0,0 +1,480 @@
...
+                }
+
+                public class WhenChangesAreMadeToTheRemindersBetweenResults : WhenRemindersAreRegistered
```
ilijagrbic: nit: Consider improving the naming of this class/test case, for example:
`WhenRemindersAreRegistered.`**`WithRemaindersChanged...`**

test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorServiceTests.cs
```diff
@@ -0,0 +1,480 @@
...
+                {
+                    public WhenChangesAreMadeToTheRemindersBetweenResults(ServiceStateFixture serviceStateFixture) : base(serviceStateFixture) { }
+
+                    public class WhenActorIdIsGiven : WhenChangesAreMadeToTheRemindersBetweenResults
+                    {
+                        public WhenActorIdIsGiven(ServiceStateFixture serviceStateFixture) : base(serviceStateFixture) { }
+
+                        [Fact]
+                        public async Task ChangesToReminderAreNotReflectedInPageThatHasBeenRead()
```
ilijagrbic: I think we don't need class hierarchy per `ActorIdIsGiven`. But can probably leverage test hierarchy for the "ReflectedInPage" part. So instead of the example from previous comment:
`WhenRemindersAreRegistered.`**`WithRemaindersChanged...`**

Perhaps something like this could work:
`WhenRemindersAreRegistered.`**`WithRemaindersChangedUnrelatedPage.SameRemaindersPerActorReturned`**
`WhenRemindersAreRegistered.`**`WithRemaindersChangedUnrelatedPage.SameRemaindersReturned`**
And:
`WhenRemindersAreRegistered.`**`WithRemaindersChangedReadPage.UnchangedRemaindersPerActorReturned`**
`WhenRemindersAreRegistered.`**`WithRemaindersChangedReadPage.UnchangedRemaindersReturned`**

test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorServiceTests.cs
```diff
@@ -0,0 +1,480 @@
...
+                            // Act
+                            Dictionary<ActorId, List<IActorReminder>> actualQueryResult = await QueryRemindersWithChangesOnFirstPage(actorService, null, editStateWhenFirstPageIsRead);
+
+                            // Assert
+                            Assert.True(ReminderDictionariesAreEqual(expectedQueryResult, actualQueryResult));
+                        }
+                    }
+                }
+
+                public class WhenChangesAreMadeToActorsBetweenResults : WhenRemindersAreRegistered
```
ilijagrbic: nit: Consider improving the naming of this class/test case, for example:
`WhenRemindersAreRegistered.`**`WithActorsChanged...`**

ilijagrbic: Now when I looked again, we only delete actors here, so perhaps something like this:
`WhenRemindersAreRegistered.`**`WithActorDeleted...`**

test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorServiceTests.cs
```diff
@@ -0,0 +1,480 @@
...
+                            firstPage = false;
+                        }
+                        
+                        ReadReminderPage(page, output);
+                    }
+                    while (continuationToken != null);
+
+                    return output;
+                }
+            }
```
ilijagrbic: Do we think these methods would be useful outside of this specific test?

ilijagrbic (approved):
Hey @plave0 , great work on migrating these tests. I think we covered everything we already had in the [script tests workload](https://dev.azure.com/msazure/One/_git/WindowsFabric?path=%2Fsrc%2Ftest%2Fcurrent%2Fsource%2Fsdk%2FClientWorkload%2FActorReminderEnumerationClientWorkload.cs&_a=contents&version=GBdevelop) and then some. I like how test cases are short and easily understandable, especially compared to previous script tests where I needed to spend some time just to find what the test does. :)

I left a bunch of comments, mostly relating to code organization and naming. I think this is important especially when we have many (and complex) test cases.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorServiceTests.cs
plave0: @ilijagrbic this helper methods are specific for querying reminders. They might come in handy if we have some more tests that do reminder querying, but I am not aware of that. 

Do you have suggestions on where to place this helper methods?

ilijagrbic: Perhaps we could put it in the `ActorServiceIntegrationTest` class for now (which we can move to a separate file). That way each test that inherits it will have common Actor operations (`GetActorService`, `QueryReminders` etc.). If this grows too large, we can always consider extracting Remainder methods from it.

test/Microsoft.ServiceFabric.Actors.IntegrationTests/Reminders/ReminderRegistrationTest.cs
plave0: Thank you for the suggestions @ilijagrbic

Regarding your first comment, I agree that these three classes can and should be extracted in some way. We also have `ITestableActor` and `TestableActor` in `ActorServiceIntegrationTest` which do not implement any API but are necessary for creating an `ActorService` since we cannot just use the base `Actor` class. Both of this use cases can most likely be generalized or mocked in some way so that they can be used in any test class. But for now, I will leave the implementation as is until we migrate a few more test scenarios and see how different test cases use this `TestableActor`. 

And I agree with you second comment as well. The naming in this test class got a bit out of hand and needs to be shortened :)

ilijagrbic (approved):

PR #468, opened 2025-09-09
Improve resiliency and diagnostics of ICommunicationListener instantiation

olegsych:
- Make `FabricTransportServiceRemotingListener` immutable to eliminate the possibility of `NullReferenceException`s after disposal
- Implement `TracingCommunicationListener`, an `ICommunicationListener` that adds robust tracing, including instance hash code, to help us understand lifecycle of the listeners created by the user code.
- Implement `Trace` encapsulating tracing in the context of a `ServiceContext`. It implements `ITrace`, a high-level abstraction meant for instrumenting runtime code.
- Make `ServiceEventSource` implement `IServiceEventSource`, a low-level abstraction used to test `Trace` implementation.
- Strengthen argument checking in `ServiceReplicaListener` and `ServiceInstanceListener`, make them immutable and add Instantiate method to encapsulate creation of a `TracingCommunicationListener` wrapping an `CommunicationListener` created by the user code.
- Make `StatefulServiceReplicaAdapter` and `StatelessServiceInstanceAdapter` use `TracingCommunicationListener` instead of tracing listener operations explicitly
- Increment `SF.ActorsServices.Internal` package version to 9.7.0

https://dev.azure.com/msazure/One/_workitems/edit/34587764


src/Microsoft.ServiceFabric.Services/ServiceEventSource.cs
```diff
@@ -13,12 +13,12 @@ namespace Microsoft.ServiceFabric.Services
     /// Reliable Services event source collected by Service Fabric runtime diagnostics system.
     /// </summary>
     [EventSource(Guid = "27b7a543-7280-5c2a-b053-f2f798e2cbb7", Name = "ServiceFramework")]
-    internal sealed class ServiceEventSource : ServiceFabricEventSource
+    sealed class ServiceEventSource : ServiceFabricEventSource, IServiceEventSource
     {
         /// <summary>
         /// Gets instance of <see cref="ServiceEventSource"/> class.
         /// </summary>
-        internal static readonly ServiceEventSource Instance = new ServiceEventSource();
```
plave0, line 21: What is the benefit of using a private setter instead of `readonly` in this case? I don't see we are setting the value of `Instance` anywhere apart from it's declaration?

src/Microsoft.ServiceFabric.Services/Runtime/StatefulServiceReplicaAdapter.cs
```diff
@@ -523,45 +515,26 @@ private async Task<ServiceEndpointCollection> OpenCommunicationListenersAsync(
...
             foreach (var entry in this.replicaListeners)
             {
-                string traceMsg;
-
                 if (entry is null)
                 {
-                    traceMsg = "Skipped (<null>) replica listener.";
-                    ServiceTrace.Source.WriteInfoWithId(TraceType, this.traceId, traceMsg);
-
+                    ServiceTrace.Source.WriteInfoWithId(TraceType, traceId, "Skipped (<null>) replica listener.");
```
plave0, line 520: Can we use `ServiceEventSource.Instance` instead of `ServiceTrace.Source`? Both can be used to get an instace of `ServiceEventSource` but are used interchanging throughout the codebase (although `ServiceTrace.Source` is used much more often). We should consider whether we really need to keep both.

src/Microsoft.ServiceFabric.Services/Trace.cs
```diff
@@ -0,0 +1,39 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using System.Fabric;
+
+namespace Microsoft.ServiceFabric.Services
+{
+    sealed class Trace : ITrace, IEquatable<Trace>
```
plave0, line 11: Currently, we use the `Trace` class only for `TracingCommunicationListener`? What are other use cases where we should start using this high-level abstraction instead of `ServiceEvensSource`?

src/Microsoft.ServiceFabric.Services/Friend.cs
```diff
@@ -23,3 +23,4 @@
...
 [assembly: InternalsVisibleTo("Microsoft.ServiceFabric.Actors.StateMigration.Tests, PublicKey=00240000048000009400000006020000002400005253413100040000010001003f8c902c8fe7ac83af7401b14c1bd103973b26dfafb2b77eda478a2539b979b56ce47f36336741b4ec52bbc51fecd51ba23810cec47070f3e29a2261a2d1d08e4b2b4b457beaa91460055f78cc89f21cd028377af0cc5e6c04699b6856a1e49d5fad3ef16d3c3d6010f40df0a7d6cc2ee11744b5cfb42e0f19a52b8a29dc31b0")]
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
```
plave0, line 26: What is this assembly?

plave0 (approved):
Leaving a few questions. Other than that, the changes look good, very good improvements.

src/Microsoft.ServiceFabric.Services/ServiceEventSource.cs
olegsych: This is a test seam. Static readonly fields cannot be changed by Reflection APIs in .NET Core, so a property with a private getter is the next best thing I could find. We've already hit this with other tests, so I thought this was understood well enough. I'll add a comment to explain in a separate PR.

src/Microsoft.ServiceFabric.Services/Runtime/StatefulServiceReplicaAdapter.cs
olegsych: Yeah, and now we also have Trace, which I realized overlaps with ServiceTrace. Let's tackle it in a separate PR.

src/Microsoft.ServiceFabric.Services/Trace.cs
olegsych: I'm not sure yet. Generally speaking, we should use the highest-level abstractions appropriate for a specific context. For tracing, ITrace is a more of a mid-level abstraction because it uses string messages, so we should use it more than the low-level ServiceEventSource. However, monitoring abstractions could be further elevated to the level of the business logic although we're probably not there yet and won't be for a while.

src/Microsoft.ServiceFabric.Services/Friend.cs
olegsych: It's the CastleProxy assembly used by Moq to generate mocks dynamically. This one should be familiar to people using dynamic mocking frameworks, so I would prefer not to explain it every time.

PR #470, opened 2025-09-17
Introduce IExceptionSerializerSettings

gobradovic:
WCF remoting listener should not depend on Fabric Transport, so 
it should not use `FabricTransportRemotingListenerSettings`.

We introduce common interface, `IExceptionSerializerSettings` that
exposes settings needed for `ExceptionSerializer` (currently `RemotingExceptionDepth`)
and which is implemented by `FabricTransportRemotingListenerSettings`
and `WcfRemotingListenerSettings`.

Bump minor version (`9.7.0` -> `9.8.0`).


src/Microsoft.ServiceFabric.Services.Remoting/V2/Runtime/IExceptionSerializerSettings.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// <summary>
+    /// Defines the interface that provides settings for exception serialization.
+    /// </summary>
+    public interface IExceptionSerializerSettings
+    {
+        /// <summary>
+        /// Gets maximal exception depth to be serialized for remoting transfer.
+        /// </summary>
+        /// <returns>Maximum exception depth for remoting transfer.</returns>
+        int GetRemotingExceptionDepth();
```
olegsych: Why not make this a get-only property matching the one in the FabricTransportRemotingListenerSettings?

src/Microsoft.ServiceFabric.Services.Remoting/V2/Runtime/IExceptionSerializerSettings.cs
```diff
@@ -0,0 +1,19 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Runtime
+{
+    /// <summary>
+    /// Defines the interface that provides settings for exception serialization.
+    /// </summary>
+    public interface IExceptionSerializerSettings
```
olegsych: Can this interface be internal? I don't think we have any public APIs exposing it. The public classes can implement internal interfaces, like the IStatefulUserServiceReplica implemented by the StatefulServiceBase

src/Microsoft.ServiceFabric.Services.Remoting/V2/Runtime/ExceptionSerializer.cs
```diff
@@ -10,23 +10,24 @@
...
     {
+        public static readonly int DefaultRemotingExceptionDepth = 2;
+
         static readonly string TraceEventType = "ExceptionSerializer";
         readonly IEnumerable<IExceptionConvertor> convertors;
-        readonly FabricTransportRemotingListenerSettings settings;
+        readonly IExceptionSerializerSettings settings;
 
-        public ExceptionSerializer(IEnumerable<IExceptionConvertor> convertors, FabricTransportRemotingListenerSettings settings)
+        public ExceptionSerializer(IEnumerable<IExceptionConvertor> convertors, IExceptionSerializerSettings settings)
```
olegsych: Since we're already touching these declarations, it would be great to make the visibility more accurate. They are definitely not public, but internal because the class is. Should the constructor be private?

src/Microsoft.ServiceFabric.Services.Remoting/V2/Runtime/ExceptionSerializer.cs
```diff
@@ -35,7 +36,7 @@ public static ExceptionSerializer CreateDefault(IEnumerable<IExceptionConvertor>
                 new DefaultExceptionConvertor()
             };
 
-            return new ExceptionSerializer(convertors, settings ?? FabricTransportRemotingListenerSettings.GetDefault());
```
olegsych, line 38: Since now it could be null, I would add null checking for settings here or in the constructor to make sure we don't have NullReferenceException later.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Runtime/IExceptionSerializerSettings.cs
olegsych: If internal, this interface feels self-explanatory to me, so I would remove the doc comments. As a side note, I find many <returns> comments redundant; usually <summary> already says the same thing and the return type of the method or property adds the rest.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfRemotingListenerSettings.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// <summary>
+    /// Settings that configures the Wcf Remoting Listener.
+    /// </summary>
+    public class WcfRemotingListenerSettings : IExceptionSerializerSettings
+    {
+        private int remotingExceptionDepth;
+
+        internal static WcfRemotingListenerSettings GetDefault()
+        {
+            return new WcfRemotingListenerSettings() { remotingExceptionDepth = ExceptionSerializer.DefaultRemotingExceptionDepth };
```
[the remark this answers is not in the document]
olegsych: I would remove GetDefault; it seems to do the same thing as the constructor. For types with parameterless constructor, a new instance serving as default is a common C# idiom.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfRemotingListenerSettings.cs
```diff
@@ -0,0 +1,62 @@
...
+        public int RemotingExceptionDepth
+        {
+            get
+            {
+                return this.remotingExceptionDepth;
+            }
+
+            set
+            {
+                if (value <= 0)
```
olegsych, line 38: Ugh... I really would prefer ArgumentOutOfRangeException. But since this is consistent with the other settings, let's add unit tests for this because it's a public API.

olegsych (changes requested):

src/Microsoft.ServiceFabric.Services.Remoting/V2/Runtime/IExceptionSerializerSettings.cs
```diff
@@ -0,0 +1,12 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Runtime
+{
+    internal interface IExceptionSerializerSettings
```
olegsych: Just a nit for the future PRs: types have internal visibility by default, so I think it's better to not specify it explicitly. It's similar to how the interface members have default visibility that cannot be specified at all.

test/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionDeserializerTest.cs
```diff
@@ -85,7 +86,8 @@ public class ConvertRemoteException : ExceptionDeserializerTest
             public void ReturnsOriginalExceptionIfItIsKnownExceptionType()
             {
                 // Arrange
-                var exceptionSerializer = Remoting.V2.Runtime.ExceptionSerializer.CreateDefault(Enumerable.Empty<Remoting.V2.Runtime.IExceptionConvertor>(), null);
+                var exceptionSerializer = Remoting.V2.Runtime.ExceptionSerializer
```
olegsych: This may have been one of those places you've mentioned during standup where we had a ripple effect from the null checking change. I think this was a good trigger to re-evaluate whether the duplication we have in the test "arrange" logic is still acceptable. With 4 identical changes, I would say the rule of three is broken/satisfied and extract the `exceptionSerializer` variables with their initialization to a field reused by all tests.

test/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionDeserializerTest.cs
```diff
@@ -85,7 +86,8 @@ public class ConvertRemoteException : ExceptionDeserializerTest
...
+                        .CreateDefault(Enumerable.Empty<Remoting.V2.Runtime.IExceptionConvertor>(), new FabricTransportRemotingListenerSettings());
```
olegsych: I think it's OK to use the FabricTransportRemotingListenerSettings here, but since we have an interface, it would be best to use Mock<IExceptionSerializerSettings> and test the ExceptionDeserializer without the dependendcy.

test/Microsoft.ServiceFabric.Services.Tests/Remoting/FabricTransport/Runtime/FabricTransportRemotingListenerSettingsTest.cs
```diff
@@ -0,0 +1,98 @@
...
+
+using Microsoft.ServiceFabric.Services.Remoting.FabricTransport.Runtime;
+using Microsoft.ServiceFabric.Services.Remoting.V2.Runtime;
+using Xunit;
+
+namespace Microsoft.ServiceFabric.Services
+{
+    public abstract class FabricTransportRemotingListenerSettingsTest
+    {
+        public FabricTransportRemotingListenerSettingsTest()
```
olegsych: this is odd. An abstract class shouldn't need a public constructor, particularly an empty one.

test/Microsoft.ServiceFabric.Services.Tests/Remoting/FabricTransport/Runtime/FabricTransportRemotingListenerSettingsTest.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+        }
+
+        public sealed class Constructor : FabricTransportRemotingListenerSettingsTest
+        {
+            [Fact]
+            public void ShouldInitializeRemotingExceptionDepthToDefaultValue()
+            {
+                // Act
```
olegsych: Just a nit: As long as there is a single small paragraph per section, I think we can omit the arrange/act/assert comments and reduce the amount of code.

test/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfRemotingListenerSettingsTest.cs
```diff
@@ -0,0 +1,98 @@
...
+
+using Microsoft.ServiceFabric.Services.Remoting.V2.Runtime;
+using Microsoft.ServiceFabric.Services.Remoting.V2.Wcf.Runtime;
+using Xunit;
+
+namespace Microsoft.ServiceFabric.Services.Wcf.Remoting.V2.Wcf.Runtime
+{
+    public abstract class WcfRemotingListenerSettingsTest
+    {
+        public WcfRemotingListenerSettingsTest()
```
olegsych: Empty public constructors shouldn't be needed, particularly in abstract classes.

test/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfRemotingListenerSettingsTest.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+        }
+
+        public sealed class Constructor : WcfRemotingListenerSettingsTest
+        {
+            [Fact]
+            public void ShouldInitializeRemotingExceptionDepthToDefaultValue()
+            {
+                // Act
+                var settings = new WcfRemotingListenerSettings();
```
olegsych: I would make this a readonly field in the `WcfRemotingListenerSettingsTest` class, called `sut` for System Under Test and initialize it inline or, if it needs parameters that cannot be constructed inline, in the base test class constructor. That way you don't have to repeat the same statement in every test.

olegsych (approved):
:shipit: Just a couple of comments with musings about test organization and ongoing refactoring. Feel free to merge as is; it would be nice to integrate it by Monday. We can always polish it in the next iteration.

olegsych (approved):
:shipit:

PR #471, opened 2025-09-19
Make UnstructuredTracePublisher emit SF trace id and type on Linux

olegsych:
- Generalize tracing abstractions and test logic originally defined in `Microsoft.ServiceFabric.Services` to make them reusable from `Microsoft.ServiceFabric.Diagnostics` and `Microsoft.ServiceFabric.TestFramework` respectively.
  - Move `IServiceEventSource` to `Microsoft.ServiceFabric.Diagnostics.Tracing` and rename it to `ITextEventSource`.
  - Introduce `EventSourceTest<TEventSource>` in `Microsoft.ServiceFabric.TestFramework` for unit testing classes derived from `ServiceFabricEventSource` and those implementing the `ITextEventSource`.
  - Make `ServiceFabricStringEventSource` and `ActorEventSource` implement the `ITextEventSource` in addition to the `ServiceEventSource` which implemented the old `IServiceEventSource`.
  - Add `NoiseText` to `ITextEventSource`, it is already implemented by all event source classes now implementing the `ITextEventSource` interface.
  - Move `ITrace` and `Trace` for reuse to `Microsoft.ServiceFabric.Diagnostics.Tracing` as well.
- Make `UnstructuredTracePublisher` emit Service Fabric trace `id` and `type` properties correctly in unstructured traces for Linux.


test/Microsoft.ServiceFabric.Diagnostics.Tests/Tracing/UnstructuredTracePublisherTest.cs
```diff
@@ -0,0 +1,157 @@
...
+                @event.Property<EventLevel>().Set(level);
+                @event.Property<ReadOnlyCollection<object>>().Set(new ReadOnlyCollection<object>(payload));
+                @event.Property<ReadOnlyCollection<string>>().Set(new ReadOnlyCollection<string>(payloadNames));
+                @event.Property<string>(nameof(EventWrittenEventArgs.Message)).Set(CreateEventMessage());
+                return @event;
+            }
+
+            string CreateEventMessage()
+            {
+                var message = new StringBuilder("Message " + fuzzy.String().Replace("{", string.Empty).Replace("}", string.Empty));
```
gobradovic: I see that `FuzzyString` can accept set of characters. Would it be more readable to create it directly with set of characters we want? Something like `new FuzzyString(fuzzy, length, PrintableAsciiCharacters().Except({ '{', '}' }));`

test/Microsoft.ServiceFabric.Diagnostics.Tests/Tracing/UnstructuredTracePublisherTest.cs
```diff
@@ -0,0 +1,157 @@
...
+            {
+                int id = fuzzy.Int32().Maximum(0); // To prevent EventWrittenEventArgs from fetching EventLevel from uninitialized EventMetadata
+                string name = "Event" + fuzzy.String();
+                var level = fuzzy.Enum<EventLevel>();
+                return CreateEvent(eventSource, id, name, level);
+            }
+
+            EventWrittenEventArgs CreateEvent(EventSource eventSource, int id, string name, EventLevel level)
+            {
+                var @event = Type<EventWrittenEventArgs>.New(eventSource, id);
```
gobradovic, line 104: Can we rename it to avoid escape? `writtenEvent` or something similar?

test/Microsoft.ServiceFabric.Diagnostics.Tests/Tracing/UnstructuredTracePublisherTest.cs
olegsych: Good point. It's been so long since I've actively coded, I forgot my own API :)

test/Microsoft.ServiceFabric.Diagnostics.Tests/Tracing/UnstructuredTracePublisherTest.cs
olegsych: Hmm... add a whole word to eliminate one symbol?

test/Microsoft.ServiceFabric.Diagnostics.Tests/Tracing/UnstructuredTracePublisherTest.cs
olegsych: Switched to the existing LettersOrDigits().

test/Microsoft.ServiceFabric.Diagnostics.Tests/Tracing/UnstructuredTracePublisherTest.cs
olegsych: Thanks for researching and sharing your thoughts on this, @gobradovic. You've raised good points, on both sides. 

I only started escaping C# keywords after years of TDD and aggresive application of the SOLID principles. Once you boil down a method to a few lines and a class to a page of code, longer variable names start to standout as inappropriately long for the small context where they are used. So, when you're looking at just a few lines of code in a method, a parameter or a variable called @event is no longer confusing and arguably more readable than eventWritten if it makes other lines longer than comfortable to read. The opposite extreme of this is when people name variables with the full name of their class, which would be something like eventWrittenEventArgs in this case. I actually commented on this practice in one of the recent PR where someone had an even longer variable name in a very small method. 
 
The more important aspect to me is whether the variable name conveys it's intent. Ideally, they should be closely related, but especially when using someone else's APIs the intent can be different from the type. In our case, the CreateEvent method returns an EventWrittenEventArgs instance that will be passed to the OnEventWritten method under test. So, in my mental model for this method, it's an "event" that hasn't been "written" yet. 
 
Event is the only keyword I can think of escaping in application code. I think I've escaped other keywords in reflection-heavy code like Inspector, but they are rarely, if ever used in the application terminology. Also, because of the PascalCase convention for public .NET APIs, escaping is only applicable to small scopes, mostly a single method, a class at the most. As the scope increases, the variable names should to and escaping becomes unnecessary.

test/Microsoft.ServiceFabric.TestFramework/Friend.cs
```diff
@@ -0,0 +1,19 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System.Runtime.CompilerServices;
+using Microsoft.ServiceFabric;
+
+[assembly: InternalsVisibleTo("Microsoft.ServiceFabric.Actors.Tests" + Signature.PublicKey)]
+[assembly: InternalsVisibleTo("Microsoft.ServiceFabric.Diagnostics.Tests" + Signature.PublicKey)]
+[assembly: InternalsVisibleTo("Microsoft.ServiceFabric.Services.Tests" + Signature.PublicKey)]
```
ilijagrbic, line 11: I like how more readable this is. We could do something similar in other `Friend.cs` files, especially in the ones where we use two different keys.

src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs
```diff
@@ -27,6 +27,13 @@ internal abstract class ServiceFabricEventSource : EventSource
         private readonly TraceConfig configMgr;
         private readonly string eventSourceName;
 
+        // Expected event properties for ITextEventSource implementations
+        internal const int InfoTextEventId = 1;
+        internal const int WarningTextEventId = 2;
+        internal const int ErrorTextEventId = 3;
+        internal const int NoiseTextEventId = 4;
+        internal const string TextEventFormat = "{2}";
```
ilijagrbic, line 35: Consider moving these constants to the `ITextEventSource` interface. The constants are not necessarily relating to every descendent of `ServiceFabricEventSource` but are relating to every descendent of `ITextEventSource`.

ilijagrbic: Ah, I just double checked, and unfortunately static interface fields are not supported in .NETFramework (they are only suppored in .NETCore), so having that in mind it makes sense to keep them in this class.

ilijagrbic (approved):
Thanks @olegsych for these changes, it all looks good.

I especially like the test improvements, just going through them was quite insightful.

gobradovic (approved):

src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs
olegsych: Good suggestion. I've tried it and as you said they aren't supported on .NET Framework. I didn't dig deeply enough to understand which are C# features and which are the .NET feature. Hopefully we can make more modern language features available by bumping the LangVersion.

test/Microsoft.ServiceFabric.TestFramework/Friend.cs
olegsych: Thanks! I wonder if Copilot can help with this.

src/Microsoft.ServiceFabric.Diagnostics/Tracing/ServiceFabricEventSource.cs
ilijagrbic: Weird part is that we already have LangVersion set to 11.0 (and this feature is supported from 9.0+ in .NET). Perhaps it could be that it is just not supported in NETFramework 4.6.2.
Another way I thought we could resolve this is by introducing an `enum` for the text trace type and then just cast `enum` values as `int` and use them as ids. But that might be overdoing it a little bit. :)

olegsych: I like the idea of using an enum. What do you think about reusing EventLevel enum items for event IDs? I don't think we care about IDs of these events because they aren't "public" but I'm not 100% sure.

PR #472, opened 2025-09-24
Extract ITextEventSourceExtensions from classes implementing ITextEventSource

olegsych:
- Extract duplicate methods like `WriteInfo`, `WriteInfoWithId` and similar methods for Error, Warning and Verbose text events from the `ServiceFabricStringEventSource`, `ActorEventSource` and `ServiceEventSource` into a new class `ITextEventSourceExtensions`, where they are now implemented as extension methods for the `ITextEventSource` interface implemented by these classes.
- Fully test the `ITextEventSourceExtensions`.
- Add `using Microsoft.ServiceFabric.Diagnostic.Tracing` directives to the code files that previously used `Write` methods implemented by one of the `*EventSource` classes and now need to use the `ITextEventSourceExtensions` from this namespace.
- Opportunistically move the misplaced using directives above the namespace.


src/Microsoft.ServiceFabric.Diagnostics/Tracing/ITextEventSourceExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+            eventSource.WarningText(string.Empty, type, Message(format, args));
+
+        internal static void WriteWarningWithId(this ITextEventSource eventSource, string type, string id, string message) =>
+            eventSource.WarningText(id, type, message);
+
+        internal static void WriteWarningWithId(this ITextEventSource eventSource, string type, string id, string format, params object[] args) =>
+            eventSource.WarningText(id, type, Message(format, args));
+
+        static string Message(string format, object[] args) =>
+            string.Format(CultureInfo.InvariantCulture, format, args);
```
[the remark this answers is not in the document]
olegsych: That's the job of string.Format.

properties/service_fabric_common.props
```diff
@@ -22,7 +22,7 @@
     <!-- Major Version for Microsoft.ServiceFabric.Diagnostics binary - Bump together with normal MajorVersion -->
     <!-- We migrated Diagnostics package from WindowsFabric repo and we need to keep the versioning consistent -->
     <MajorVersion Condition="'$(MSBuildProjectName)' == 'Microsoft.ServiceFabric.Diagnostics'">12</MajorVersion>
-    <MinorVersion>9</MinorVersion>
+    <MinorVersion>10</MinorVersion>
```
ilijagrbic: In the meantime, I bumped the version to 9.10.0 (and build a new package), so we need to bump it to 11 for this change.

ilijagrbic (approved):
Great follow up on the previous PR @olegsych . 
I like that we removed a lot of the repeated code from event source classes and that this code is now tested as well. :shipit:

test/Microsoft.ServiceFabric.Diagnostics.Tests/Tracing/ITextEventSourceExtensionsTest.cs
```diff
@@ -0,0 +1,151 @@
...
+        // Method parameters
+        readonly string type = "Type" + fuzzy.String();
+        readonly string id = "Id" + fuzzy.String();
+        readonly string message = "Message" + fuzzy.String();
+        readonly string format;
+        readonly object[] args = fuzzy.Array(() => (object)fuzzy.DateTime()); // Culture-sensitive
+
+        public ITextEventSourceExtensionsTest()
+        {
+            var format = new StringBuilder("Format" + fuzzy.String().Replace("{", string.Empty).Replace("}", string.Empty));
```
gobradovic: Did you intentionally use this approach instead of `fuzzy.String().LettersOrDigits()` as in PR #471.

gobradovic (approved):

test/Microsoft.ServiceFabric.Diagnostics.Tests/Tracing/ITextEventSourceExtensionsTest.cs
olegsych: Good catch. I had a feeling there was another instance of this I forgot about, it turns out it was in a different PR. Thanks for paying attention!

ilijagrbic (approved):

PR #474, opened 2025-09-25
Implement TimeSpanMeter classes and refactor Record methods

ilijagrbic:
Changes:
- Implement `TimeSpanMeter/1D/2D/3D` classes in order to support TimeSpan metrics
- Currently every Meter class has its own logic on how to call native telemetry. In order to reduce this duplication and make the code easier to refactor introducing `Record` method in `Meter` base class which handles logic for calling native telemetry implementation and converting dimensions into an appropriate format.
- In all Meter classes, refactor `Record` methods to call the base class's `Record`
- Multiple minor Unit test changes


src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -11,6 +11,8 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 {
     abstract class Meter
     {
+        private readonly Action<long, int, string, string, string> recordAction;
```
olegsych, line 14: This works, although it feels odd to use the same technique we use for de-coupling static method calls to test instance methods in a derived classes. It's fine like this, I just think it's a workaround for cases when dependency between two components is too difficult to test otherwise and try to avoid when alternatives are available.

It's often a sign of a potentially missing higher-level, better-testable abstraction. We tend to default to inheritance, but composition is a more powerful code reuse technique and it's easier to test. For example, suppose this class was called something like `Recorder<T>`, implemented an `IRecord<T>` interface with the three `Record` methods and that the `Int64Meter` received an `IRecorder<long>` as a single constructor parameter instead of the `fabricMeter` and `systemDimensions`. Then you could test the `Recorder` implementation separately, just like you're testing the `Meter` now. However, testing of the `Int64Meter` would need only a straightforward `Mock.Of<IRecord<long>>()` without the need to manipulate the base class via Reflection.

Alternatively, if the code is simple enough, re-testing each derived class without assumptions of whether it uses inherited logic or re-implements it could be reasonable too, particularly if we encapsulate the test logic in a reusable test helper like the EventSourceTest. So even if we end up running more tests, they'll be implemented with reusable code.

olegsych (approved):
The PR is good to go @ilijagrbic. Thanks for iterating over the unit tests, it's very important. I left some thoughts on the design and testability of the Meter class, but they are more of an ongoing conversation than critique. 

:shipit:

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/Meter.cs
ilijagrbic: I think this case is quite specific since we not only have inheritance, but we also have a dependency on the base class's method. I like the idea to use composition instead of inheritance now that the code has evolved. We tend to default to inheritance, especially when it "seems" this obvious, but thinking about it now, there is no behavior we would want to override in this case, we kind of want to do the opposite, which is a sign that inheritance is probably not needed. I don't think this is a huge issue now, and we can easily refactor it, so, lets revisit this in the future iterations and hopefully make it even better. :)

Regarding the reusable test helper approach, although I like it, I wanted to avoid creating reusable test helper for this particular case. It makes sense to invest time and complexity in such an approach if we think it is likely that the tests will need to scale (for example by implementing several additional `ServiceFabricEventSources` we would need more of the same tests). In this case, scaling is really unlikely, perhaps there will be need for more than 3 custom dimensions, but we both think that is really unlikely having in mind we already emit a lot of system dimensions. So, here I preferred to have simpler tests. Even if there is some code that can be reused between them, the tests are already small, and arguable won't get much smaller by introducing additional testing abstractions.

PR #476, opened 2025-09-26
Cleanup InternalsVisibleTo attributes

olegsych:
- Extract public and test keys into reusable constants
  - Create a new project, `Microsoft.ServiceFabric.Constants` to hold compile-time constants that should produce runtime-dependencies.
  - Move the Signature class from the `TestFramework` to the `Constants` project and rename it to `AssemblyInfo`
  - Replace hard-coded public keys in all projects with the `AssemblyInfo` constants.
- Remove unnecessary `InternalsVisibleTo` attributes
  - Assemblies produced by projects in this repository don't use the `TestKey`
  - Assemblies that no longer exist
- Extract the `DynamicProxyGenAssembly2` literal from all projects to `Constants.AssemblyInfo`.
- Rename `Friend*.cs` files to `AssemblyInfo.cs`
  Old file names used inconsistent capitalization, some used underscores. `AssemblyInfo.cs` is a widely-used convention established by the Visual Studio project templates.


src/Microsoft.ServiceFabric.Diagnostics/AssemblyInfo.cs
```diff
@@ -4,44 +4,37 @@
...
+[assembly: InternalsVisibleTo("EventsValidationTest" + PublicKey)]
+[assembly: InternalsVisibleTo("AzureFilesVolumePlugin" + PublicKey)]
+[assembly: InternalsVisibleTo("AzureFilesVolumePluginSetup" + PublicKey)]
+[assembly: InternalsVisibleTo("SFVolumeDiskDebugAgentCommon" + PublicKey)]
+[assembly: InternalsVisibleTo("NodeAgentServiceSetup" + PublicKey)]
+[assembly: InternalsVisibleTo("Microsoft.ServiceFabric.Diagnostics.Tests" + PublicKey)]
+[assembly: InternalsVisibleTo("Microsoft.ServiceFabric.Actors.Tests" + PublicKey)]
+[assembly: InternalsVisibleTo("Microsoft.ServiceFabric.Services.Tests" + PublicKey)]
+[assembly: InternalsVisibleTo("Microsoft.ServiceFabric.TestFramework" + PublicKey)]
+[assembly: InternalsVisibleTo(DynamicProxyGenAssembly2)]
```
ilijagrbic, line 30: Why are we not creating `AssemblyInfo.cs` in this project instead of using old `_friend.cs`? I see that for most project we are using `AssemblyInfo.cs`, except for `Diagnostics` and `Services.Wcf`.

src/Constants/Microsoft.ServiceFabric.Constants.csproj
```diff
@@ -0,0 +1,10 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <!--
+    This project contains only compile-time constants.
+    Projects referencing it should not generate assembly references in their assemblies.
+    The assembly produced by this project should not be included in any of the NuGet packages.
+  -->
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
ilijagrbic, line 8: Why are we using netstandard here instead of having net462 and net8.0 (as in other projects)? I suspect there is a good reason, I just haven't figured it out. :)

ilijagrbic (changes requested):
Hey @olegsych , awesome changes, thanks for doing this. I think the new way to add `InternalsVisibleTo` is much more readable and scalable. 
Before merging I would just consider renaming `_friend.cs` files from Diagnostics/Services.Wcf project into `AsseblyInfo.cs` for consistency through all projects, unless there is some other reason for this that I'm missing

src/Microsoft.ServiceFabric.Diagnostics/Microsoft.ServiceFabric.Diagnostics.csproj
```diff
@@ -7,4 +7,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.ServiceFabric" />
   </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="$(RepoRoot)src\Constants\Microsoft.ServiceFabric.Constants.csproj" />
```
plave0: nit: We should use absolute or relative paths for project references consistently throughout the repo. We use relative paths in other projects, but in this project you used an absolute path.

plave0 (approved):

src/Constants/Microsoft.ServiceFabric.Constants.csproj
olegsych: Since this project will have only constants, making it netstandard2.0 allows us to build it only once. Not much of an optimization, but still.

src/Microsoft.ServiceFabric.Diagnostics/AssemblyInfo.cs
olegsych: Good idea. I already started on this in #480

src/Microsoft.ServiceFabric.Diagnostics/Microsoft.ServiceFabric.Diagnostics.csproj
olegsych: Good point. It looks like we're using relative paths in the src projects and RepoRoot in the test projects.

src/Microsoft.ServiceFabric.Diagnostics/AssemblyInfo.cs
olegsych: Actually, it's a part of this PR. Updated.

olegsych:
Thanks for the review @plave0 and @ilijagrbic. I've addressed your comments. Could you re-approve? I think @ilijagrbic's approval is required because he requested changes.

ilijagrbic (approved):

PR #481, opened 2025-09-30
Enable MeterProvider creation without ServiceContext and minor changes

ilijagrbic:
In order to delete old telemetry API, we need to replace its usage in IS (Infrastructure Service) with the new API.
An issue we found is that Meters in IS are currently created outside of the Service context, before service is instantiated. 

We need to keep the old functionality, so we are enabling creation of MeterProviders with null ServiceContext. MeterProviders created in this way will create Meters that emit not system dimensions based on ServiceContext. Note that system dimensions emitted from the native site (nodeId, runtimeVersion etc.) will still be emitted as normal, this only applies to system dimensions read from ServiceContext.

In addition to this we are making Diatgnostics internals visible to relevant IS projects.


src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/ServiceMeterProvider.cs
```diff
@@ -12,46 +12,50 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
...
+                    serviceContext.PartitionId.ToString(),
+                    serviceContext.ServiceTypeName,
+                    serviceContext.ServiceName.ToString(),
+                    serviceContext.CodePackageActivationContext.ApplicationName,
+                    serviceContext.CodePackageActivationContext.ApplicationTypeName
+                };
+            }
+            else
+            {
+                this.systemDimensionNames = Array.Empty<string>();
```
olegsych: Can we use `Enumerable.Empty<string>()` to match the field type `IEnumerable<string>`?

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/ServiceMeterProvider.cs
```diff
@@ -12,46 +12,50 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
...
-            nameof(ServiceContext.CodePackageActivationContext.ApplicationTypeName)
-        };
+        readonly IEnumerable<string> systemDimensionNames;
         protected readonly IEnumerable<string> systemDimensionValues;
         protected readonly IFabricMeterProvider fabricMeterProvider;
 
         private static Func<IFabricMeterProvider> createFabricMeterProvider = NativeTelemetry.FabricCreateMeterProvider;
 
-        protected ServiceMeterProvider(ServiceContext serviceContext)
+        protected ServiceMeterProvider(ServiceContext serviceContext = null)
```
olegsych: Accepting a nullable ServiceContext kind of changes the purpose of this class, which is no longer a `ServiceMeterProvider`. I wonder if we should make this a plain `MeterProvider` that accepts an `IEnumerable<KeyValuePair<string, string>>` of system dimensions and provide a separate API to extract system dimensions from a `ServiceContext`.

olegsych (approved):

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/ServiceMeterProvider.cs
ilijagrbic: I wanted to rename the class, but forgot to push those changes, so I will rename it to `MeterProvider`. 

Regarding the second point, I would prefer to actually have the API accept `ServiceContext`, so that we are the ones controlling which system dimensions are emitted. If we allow `KeyValuePari` API, it will mean people could start using these system dimensions for unintended use-cases. But anyways, I see your point, we can think about this a bit more and follow up on it in the future if it is needed. :)

src/Microsoft.ServiceFabric.Diagnostics/Metrics/Implementation/ServiceMeterProvider.cs
ilijagrbic: Good point, I have changed this as well.

plave0 (approved):

PR #485, opened 2025-10-06
Enable tests on Linux

olegsych:
- Fix test failures
   - Set `IsTestProject` property to `false` in WCF test projects to prevent test failures when running `dotnet test code.sln --framework net9.0`
   - Remove `TracingTests` from the `Actors.Tests` project because it required ETW and failed on Linux. It was also superceeded by the new event source tests.
- Make `build.yml` fail `test-linux` job on test failures
- Merge the `build` and `test-windows` job so that `test-windows` can run in parallel with `test-linux`


test/Microsoft.ServiceFabric.Actors.Wcf.Tests/Microsoft.ServiceFabric.Actors.Wcf.Tests.csproj
```diff
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <ProjectGuid>{9037d86c-544d-48f2-bdb4-bf68fc8228b7}</ProjectGuid>
     <RootNamespace>Microsoft.ServiceFabric.Actors.Wcf</RootNamespace>
+    <IsTestProject Condition="'$(TargetFramework)' != 'net472'">false</IsTestProject>
```
ilijagrbic, line 5: Isn't this properties condition supposed to be always false, since this project is built only on `net472`? Why did we change this property? I'm not sure if it was done in order to prevent execution of these test projects on Linux, but one concern I have now is that these tests are not executed at all (even on Windows). I'm not sure if this is related to this property though.

Also, shouldn't the `--framework net9.0` flag be sufficient to skip the NETFramework tests in Linux pipeline?

test/Microsoft.ServiceFabric.Services.Wcf.Tests/Microsoft.ServiceFabric.Services.Wcf.Tests.csproj
```diff
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <ProjectGuid>{1bfa8614-0a48-4d17-97e5-a34e630a4c2b}</ProjectGuid>
     <RootNamespace>Microsoft.ServiceFabric.Services.Wcf</RootNamespace>
+    <IsTestProject Condition="'$(TargetFramework)' != 'net472'">false</IsTestProject>
```
ilijagrbic, line 5: This is a similar situation as with `...Actors.Wcf.Tests`, we probably need to address both in a similar manner.

test/Microsoft.ServiceFabric.Actors.Tests/Runtime/TracingTests.cs
```diff
@@ -1,97 +0,0 @@
...
-    using Microsoft.Diagnostics.Tracing.Session;
-    using Microsoft.ServiceFabric.Actors;
-    using Microsoft.ServiceFabric.Actors.Diagnostics;
-    using Xunit;
-
-    /// <summary>
-    /// Tests for ActorFrameworkEventSource
-    /// </summary>
-    public class TracingTests
-    {
```
ilijagrbic, line 22: We should call out in the PRs description why this test was removed.

test/Microsoft.ServiceFabric.Actors.Wcf.Tests/Microsoft.ServiceFabric.Actors.Wcf.Tests.csproj
olegsych: This prevents `dotnet test code.sln --framework net9.0` from failing.

test/Microsoft.ServiceFabric.Services.Wcf.Tests/Microsoft.ServiceFabric.Services.Wcf.Tests.csproj
olegsych: Same reason. Called out in the PR description.

test/Microsoft.ServiceFabric.Actors.Tests/Runtime/TracingTests.cs
olegsych: Updated.

ilijagrbic (approved):

PR #486, opened 2025-10-08
Implement TelemetryDiagnosticEvents in Services.Remoting

ilijagrbic:
- Add new `IDiagnosticEvents` implementation `TelemetryDiagnosticEvents` for metrics
- Integrate `TelemetryDiagnosticEvents` into `AgregatedDiagnosticEvents` used in `FabricTransportMessageHandler`
- Unit test these changes
- Extract telemetry API mock logic into `MockedTelemetryTest` class 
- Other minor improvements for existing Unit Tests


src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/TelemetryDiagnosticEvents.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        public void OnRemotingRequestBegin()
+        {
+            // Intentionally left blank, since we don't observe this
+        }
+
+        public void OnRemotingRequestEnd(DateTime startTime)
+        {
+            requestDeserializationTime.Record(clock.UtcNow - startTime);
+        }
```
gobradovic, line 48: Is `RemotingRequest` only deserialization? If within it is actual remoting call, then the name is misleading.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/TelemetryDiagnosticEvents.cs
```diff
@@ -0,0 +1,65 @@
...
+        internal readonly IMeter<TimeSpan> requestProcessingTime;
+        internal readonly IMeter<TimeSpan> requestDeserializationTime;
+        internal readonly IMeter<TimeSpan> responseSerializationTime;
+
+        public TelemetryDiagnosticEvents(ServiceContext serviceContext, IClock clock)
+        {
+            if (serviceContext == null)
+            {
+                throw new ArgumentException(nameof(serviceContext));
+            }
```
gobradovic: I think we can use a ternary operator here as well as for clock even if we don't assign it:

`serviceContext ?? throw new ArgumentException(nameof(serviceContext));`

test/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/TelemetryDiagnosticEventsTest.cs
```diff
@@ -0,0 +1,134 @@
...
+
+                Assert.NotNull(requestProcessingTime);
+                Assert.NotNull(requestDeserializationTime);
+                Assert.NotNull(responseSerializationTime);
+            }
+
+            [Fact]
+            public void ThrowsOnNullClock()
+            {
+                Assert.Throws<ArgumentException>(() =>
```
olegsych: Why not ArgumentNullException?
Can we verify the actual ParamName of the exception as well?

test/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/TelemetryDiagnosticEventsTest.cs
```diff
@@ -0,0 +1,134 @@
...
+{
+    public class TelemetryDiagnosticEventsTest : MockedTelemetryTest
+    {
+        static readonly IFuzz fuzzy = new RandomFuzz(Environment.TickCount);
+
+        TelemetryDiagnosticEvents sut;
+
+        readonly IClock clock = Mock.Of<IClock>();
+
+        protected TelemetryDiagnosticEventsTest() => sut = new TelemetryDiagnosticEvents(fuzzy.ServiceContext(), clock);
```
olegsych: When constructor is called multiple times, I think it helps to create a field for each parameter, to reduce duplication and communicate the parameter type. We already have a field for the clock, I would add one for the service context as well.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/TelemetryDiagnosticEvents.cs
ilijagrbic: I was thinking about this, and I was not sure what is cleaner/more readable. If you think this is better, I will change it.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/TelemetryDiagnosticEvents.cs
ilijagrbic: The idea is that `OnRemotingRequestEnd` method would handle all of the diagnostics needed on request end. In this case we only handle/emit request deserialization, but hypothetically we could emit multiple relevant metrics/perfCounters/traces.

test/Microsoft.ServiceFabric.TestFramework/IFuzzExtensions.cs
```diff
@@ -74,5 +74,11 @@ public static Type Type(this IFuzz fuzzy)
             mock.SetupGet(_ => _.Name).Returns(fuzzy.String());
             return mock.Object;
         }
+
+        public static double Double(this IFuzz fuzzy, int minValue, int maxValue)
```
plave0, line 78: `fuzzy.Double.Between(minValue, maxValue)` is supported out of the box. I don't think we need to reimplement this.

plave0 (changes requested):

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/TelemetryDiagnosticEvents.cs
```diff
@@ -0,0 +1,62 @@
...
+using Microsoft.ServiceFabric.Diagnostics;
+using Microsoft.ServiceFabric.Diagnostics.Metrics;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Diagnostic
+{
+    internal class TelemetryDiagnosticEvents : IDiagnosticEvents
+    {
+        readonly IClock clock;
+
+        internal readonly IMeter<TimeSpan> requestProcessingTime;
```
olegsych, line 16: Since this is a new class, I think we should implement idiomatic dependency injection and not mix injection with instantiation of dependencies. In particular, I would inject `IMeterProvider<TimeSpan>` as a constructor parameter instead of directly instantiating the `TimeSpanMeterProvider`. Actually, that makes the `ServiceContext` parameter unnecessary, so it could be removed. These dependency parameters should perculate to the highest possible level. In applications, this is called a "composition root" and is usually provided by a dependency injection container. In our case, it's probably going to be a public class that we cannot modify and have to place the instantiation logic there, or introduce a singleton/factory that we can intercept in tests. Still, most classes should have a clean constructor injection and no need for reflection.

This will allow testing `TelemetryDiagnosticEvents` without having to break its encapsulation. In particular, it shouldn't be neccessary to make the `IMeter<TimeSpan>` instances internal so that they can be accessed by tests. Instead, you can setup the `Mock<IMeterProvider<TimeSpan>>` to return `Mock<IMeter<TimeSpan>>` instances given expected metric namespace and name. This way we can validate that the namespace and names are correct and don't need to make any assumptions about their storage in the tests.

src/Microsoft.ServiceFabric.Services.Remoting/V2/FabricTransport/Runtime/FabricTransportMessageHandler.cs
```diff
@@ -45,10 +47,11 @@ public FabricTransportMessageHandler(
 
             this.clock = new SystemClock();
 
-
             this.serviceRemotingPerformanceCounterProvider = new ServiceRemotingPerformanceCounterProvider(this.partitionId, this.replicaOrInstanceId);
             var performanceCounterDiagnosticEvents = new PerformanceCounterDiagnosticEvents(serviceRemotingPerformanceCounterProvider, this.clock);
-            var registeredDiagnosticsEvents = new List<IDiagnosticEvents> { performanceCounterDiagnosticEvents };
+            var telemetryDiagnosticEvents = new TelemetryDiagnosticEvents(serviceContext, this.clock);
+
+            var registeredDiagnosticsEvents = new List<IDiagnosticEvents> { performanceCounterDiagnosticEvents, telemetryDiagnosticEvents };
```
olegsych, line 54: Nit: use inline collection initializer instead of a List variable?

test/Microsoft.ServiceFabric.TestFramework/IFuzzExtensions.cs
olegsych: It may be to avoid fractions. I've had a work item for something like this in the fuzzy repo for a while.

test/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/TelemetryDiagnosticEventsTest.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        public class Class : TelemetryDiagnosticEventsTest
+        {
+            [Fact]
+            public void ImplementsIDiagnosticsEvents()
+            {
+                Assert.True(typeof(IDiagnosticEvents).IsAssignableFrom(sut.GetType()));
```
olegsych: For Single-responsiblity classes, this assertion is better expressed by making the type of the sut field `IDiagnosticEvents` instead of `TelemetryDiagnosticEvents`. Then an explicit test is no longer needed.

test/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/TelemetryDiagnosticEventsTest.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+        }
+
+        public class Constructor : TelemetryDiagnosticEventsTest
+        {
+            [Fact]
+            public void WithParametersSetsClock()
+            {
+                var actualClock = sut.Field<IClock>().Value;
```
olegsych: I try to avoid testing privates, if I can help it. It doesn't matter how the clock is stored, we just want to test that the time it returns is used to calculate the metric values. Since it shouldn't be possible for this test to fail independently of the event method tests, I would remove it.

test/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/TelemetryDiagnosticEventsTest.cs
```diff
@@ -0,0 +1,134 @@
...
+                Assert.Equal(clock, actualClock);
+            }
+
+            [Fact]
+            public void WithParametersSetsMeters()
+            {
+                var requestProcessingTime = sut.Field<IMeter<TimeSpan>>(nameof(sut.requestProcessingTime));
```
olegsych: This test may be useful if we verify that meters were created with the expected namespace and name values. It would still be difficult to make fail independently of the event method tests though, so I usually end up removing the factory call tests as well. You could make a case for it helping make the tests more readable though, so I think this decision can change depending on repetition of the pattern.

test/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/PerformanceCounterDiagnosticEventsTest.cs
```diff
@@ -27,14 +27,11 @@ public class Class : PerformanceCounterDiagnosticEventsTest
             [Fact]
             public void ImplementsIDiagnosticsEvents()
             {
-                var performanceCounterDiagnosticsSourceType = typeof(PerformanceCounterDiagnosticEvents);
-                var iDiagnosticsSourceType = typeof(IDiagnosticEvents);
-
-                Assert.True(iDiagnosticsSourceType.IsAssignableFrom(performanceCounterDiagnosticsSourceType));
+                Assert.True(typeof(IDiagnosticEvents).IsAssignableFrom(sut.GetType()));
```
olegsych: Consider changing type of the `sut` field to `IDiagnosticEvents` and eliminating this test.

test/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/TelemetryDiagnosticEventsTest.cs
```diff
@@ -0,0 +1,134 @@
...
+using Xunit;
+using IClock = Microsoft.ServiceFabric.Diagnostics.IClock;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Diagnostic
+{
+    public class TelemetryDiagnosticEventsTest : MockedTelemetryTest
+    {
+        static readonly IFuzz fuzzy = new RandomFuzz(Environment.TickCount);
+
+        TelemetryDiagnosticEvents sut;
```
olegsych: Can we make `sut` read-only?

test/Microsoft.ServiceFabric.TestFramework/MockedTelemetryTest.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation;
+using Moq;
+
+namespace Microsoft.ServiceFabric.TestFramework
+{
+    public class MockedTelemetryTest
+    {
+        protected MockedTelemetryTest()
+        {
+            typeof(MeterProvider<long>).Field<Func<IFabricMeterProvider>>().Set(() => new Mock<IFabricMeterProvider>() { DefaultValue = DefaultValue.Mock }.Object);
```
olegsych: I think this deserves a doc-comment explaining why we're doing this, or perhaps a better class name. The term _Telemetry_ encompasses not only metrics but logs, events, distributed traces, etc, so it feels too vague in this context.

Don't we need to implement `IDisposable` and revert the static changes as well?

olegsych (changes requested):
Thanks @ilijagrbic I've left a couple of design suggestions, could you take a look?

test/Microsoft.ServiceFabric.TestFramework/IFuzzExtensions.cs
ilijagrbic: I think using `Between()` in such a way for Double is not enabled in the Fuzzy library, correct me if I'm wrong @olegsych. I checked the source code [here](https://github.com/olegsych/fuzzy/blob/master/src/Implementation/FuzzyDouble.cs) - and `FuzzyDouble` does not inherit from `FuzzyRange` (unlike `FuzzyInt64`). I think this is the reason I was getting `InvalidCastExceptions` when trying to use `Between()`, `Minimum()` or `Maximum() `methods.

Perhaps we could contribute this (or similar) extensions method to the original Fuzzy repo, but since this is just one of usage, I could also just revert this change if you guys think it is not needed.

test/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/TelemetryDiagnosticEventsTest.cs
ilijagrbic: Thanks for the suggestion. I wanted to remove all of the fields that are not used afterwards, to make the test more compact. But I see your point about communicating the type so I will extract it as a field.

test/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/TelemetryDiagnosticEventsTest.cs
ilijagrbic: Good suggestion, thanks! I will apply it.

plave0 (approved):

test/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/PerformanceCounterDiagnosticEventsTest.cs
ilijagrbic: Good suggestion, I will apply it.

test/Microsoft.ServiceFabric.TestFramework/MockedTelemetryTest.cs
ilijagrbic: Yeah, I missed this, since in the test we don't need to ever have original static field values. But anyways, it should be reverted, so I will implement it. :)

src/Microsoft.ServiceFabric.Services.Remoting/V2/FabricTransport/Runtime/FabricTransportMessageHandler.cs
ilijagrbic: I prefer shorter line length in this case, so I will keep it as is, unless you think it is critical.

test/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/TelemetryDiagnosticEventsTest.cs
ilijagrbic: Good suggestion, I will apply it.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Diagnostic/TelemetryDiagnosticEvents.cs
ilijagrbic: Thanks for the explanation, Oleg. This makes sense, I will instantiate `MeterProvider `in the `FabricTransportServiceRemotingListener` then, as this is the highest level we can move it to.

test/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/TelemetryDiagnosticEventsTest.cs
ilijagrbic: After I refactored the dependency injection issue, I also changed this test to verify creation of the Meters from the injected `IMeterProvider`. :)

test/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/TelemetryDiagnosticEventsTest.cs
```diff
@@ -0,0 +1,122 @@
...
+        readonly IDiagnosticEvents sut;
+
+        readonly IClock clock = Mock.Of<IClock>();
+        readonly IMeterProvider<TimeSpan> meterProvider = new Mock<IMeterProvider<TimeSpan>>() { DefaultValue = DefaultValue.Mock }.Object;
+
+        protected TelemetryDiagnosticEventsTest() => sut = new TelemetryDiagnosticEvents(meterProvider, clock);
+
+        public class Constructor : TelemetryDiagnosticEventsTest
+        {
+
```
olegsych: Nit: extra blank line?

test/Microsoft.ServiceFabric.Services.Remoting.Tests/V2/Diagnostic/TelemetryDiagnosticEventsTest.cs
```diff
@@ -0,0 +1,122 @@
...
+            [Fact]
+            public void ThrowsOnNullClock()
+            {
+                var exception = Assert.Throws<ArgumentNullException>(() =>
+                {
```
olegsych: Nit: why not eliminate the braces and place Assert.Throws(() => ); on a single line. The constructor is short, so it should be readable.

olegsych (approved):
:shipit:

plave0 (approved):

PR #487, opened 2025-10-08
Small fix in UsesPassedRemotingListenerSettings

gobradovic:
Set a particular value for `RemotingExceptionDepth` instead of
passing the default value to the factory method.

### DIFF
--- a/test/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionSerializerTest.cs
+++ b/test/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionSerializerTest.cs
@@ -82,7 +82,10 @@ namespace Microsoft.ServiceFabric.Services.Remoting.Tests
             [Fact]
             public void UsesPassedRemotingListenerSettings()
             {
-                var expectedSettings = new FabricTransportRemotingListenerSettings();
+                var expectedSettings = new FabricTransportRemotingListenerSettings()
+                {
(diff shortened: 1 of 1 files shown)


test/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionSerializerTest.cs
```diff
@@ -82,7 +82,10 @@ public void UsesDefaultConvertorsIfNullPassed()
             [Fact]
             public void UsesPassedRemotingListenerSettings()
             {
-                var expectedSettings = new FabricTransportRemotingListenerSettings();
+                var expectedSettings = new FabricTransportRemotingListenerSettings()
+                {
+                    RemotingExceptionDepth = 17
```
olegsych, line 87: Hey @gobradovic, why do we need to set the `RemotingExceptionDepth`? I remember removing it because this test uses `Assert.Same`, which verifies object reference equality, so it shouldn't matter what the contents of that object are.

gobradovic: @olegsych Ah yes, I noticed this while I was porting some changes and I thought I deleted it by mistake. Didn't look too deeply into it. Discarding this PR.

gobradovic:
Closing as per discussion in comment section. `Assert.Same` compares by reference, contents. is unimportant.

PR #488, opened 2025-10-08
Enable migration of WCF Remoting from binary serialization to DCS

gobradovic:
Enable migration of WCF Remoting from binary serialization to DCS.

In this version, WCF Remoting Client is capable of receiving both `RemoteException` (BinaryFormatter) and `RemoteException2` (DataContractSerialization), while WCF Remoting Listener will emit either `RemoteException` or `RemoteException2` depending on the `WcfRemotingListenerSettings.ExceptionSerializationTechnique` (`Default` for `RemoteException2` and `BinaryFormatter` for `RemoteException`).

We also add several missing exceptions to the lists of known exceptions that were previously added on develop branch.

Bump minor version from `8.0.16` to `8.1.0`.


src/Microsoft.ServiceFabric.Services.Remoting/FabricTransport/Runtime/FabricTransportRemotingListenerSettings.cs
```diff
@@ -44,24 +44,6 @@ private FabricTransportRemotingListenerSettings(FabricTransportListenerSettings
             this.listenerSettings = listenerSettings;
         }
 
-        /// <summary>
```
olegsych, line 47: We shouldn't be removing the `ExceptionSerialization` enum, correct? This is a breaking change.

src/Microsoft.ServiceFabric.Services.Remoting/FabricTransport/FabricTransportRemotingSettings.cs
```diff
@@ -44,25 +45,7 @@ internal FabricTransportRemotingSettings(FabricTransportSettings fabricTransport
         }
 
         /// <summary>
-        /// Exception Deserialization option to use(applies to V2 Remoting only).
-        /// </summary>
-        [Obsolete(DeprecationMessage.RemotingV1)]
-        public enum ExceptionDeserialization
```
olegsych, line 50: We shouldn't remove `ExceptionDeserialization`; it's a breaking change.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Client/ExceptionDeserialization.cs
```diff
@@ -0,0 +1,25 @@
...
+// ------------------------------------------------------------
+
+using System;
+using Microsoft.ServiceFabric.Services.Remoting;
+
+/// <summary>
+/// Exception Deserialization option to use(applies to V2 Remoting only).
+/// </summary>
+[Obsolete(DeprecationMessage.RemotingV1)]
+public enum ExceptionDeserialization
```
olegsych: We have to keep the original nested enum type. I'm not sure if we need this one. Can we define IExceptionDeserializerSettings with the ExceptionDeserializationTechnique property of type FabricTransportRemotingSettings.ExceptionDeserialization instead?

src/Microsoft.ServiceFabric.Services.Remoting/V2/Runtime/ExceptionSerialization.cs
```diff
@@ -0,0 +1,27 @@
...
+
+using System;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Runtime
+{
+    /// <summary>
+    /// Exception serialization option to use(applicable only to V2 Remoting).
+    /// </summary>
+    [Obsolete(DeprecationMessage.RemotingV1)]
+    public enum ExceptionSerialization
```
olegsych: Same concern as for `ExceptionDeserialization`

test/unittests/Microsoft.ServiceFabric.Actors.Wcf.Tests/AssemblyInfo.cs
```diff
@@ -0,0 +1,9 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System.Runtime.CompilerServices;
+using Microsoft.ServiceFabric.Actors.Remoting.Wcf;
+
+[assembly: WcfActorRemotingProviderAttribute()]
```
olegsych: This looks like a shared test fixture. Can we easily eliminate this attribute? If not add a short comment why it's needed.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Client/ExceptionDeserialization.cs
olegsych: Oh, I see. Having to use FabricTransportRemotingSettings.ExceptionDeserialization for the WcfRemotingSettings.ExceptionDeserialization property would be odd. As a point in time fix, it can be OK. Or We could define a nested enum types for Wcf settings as well, implement the IException*Settings interfaces explicitly and convert from the nested enums to the plain ones.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Client/WcfServiceRemotingClientFactory.cs
```diff
@@ -54,12 +55,84 @@ public class WcfServiceRemotingClientFactory : IServiceRemotingClientFactory
...
+        ///     It indicates whether the remoting method parameters should be wrapped or not before sending it over the wire.
+        ///     When UseWrappedMessage is set to false, parameters  will not be wrapped. When this value is set to true, the
+        ///     parameters will be wrapped.Default value is false.
+        /// </param>
+        /// <remarks>
+        ///     This factory uses <see cref="WcfExceptionHandler"/> and <see cref="ServiceRemotingExceptionHandler"/> in addition to the
+        ///     exception handlers supplied to the constructor.
+        /// </remarks>
+        [Obsolete]
+        public WcfServiceRemotingClientFactory(
```
olegsych, line 73: Could you double-check the Obsolete attributes? I think this constructor exists in develop without it, however the new one below doesn't.

olegsych: We may have a weird situation where we need to add an obsolete constructor that takes WcfRemotingSettings only to remove it in the next version.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Client/WcfRemotingSettings.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// Settings that configures the Wcf Remoting.
+    /// </summary>
+    public class WcfRemotingSettings : IExceptionDeserializerSettings
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="WcfRemotingSettings"/> class with default values.
+        /// </summary>
+        public WcfRemotingSettings()
+        {
+#pragma warning disable 618
```
olegsych: Shouldn't the whole class be obsolete?

olegsych (changes requested):

src/Microsoft.ServiceFabric.Services.Remoting/V2/FabricExceptionKnownTypes.cs
```diff
@@ -26,6 +26,14 @@ internal class FabricExceptionKnownTypes
...
                     }
                 },
+                {
+                    "System.Fabric.FabricInsufficientMaxLoadCapacityException", new ConvertorFuncs()
+                    {
+                        ToServiceExFunc = ex => ToServiceException(ex),
+                        FromServiceExFunc = (svcEx, innerEx) => FromServiceException<FabricInsufficientMaxLoadCapacityException>(svcEx, innerEx),
+                        InnerExFunc = ex => GetInnerExceptions(ex),
+                    }
+                },
```
ilijagrbic, line 36: Why are we changing the known exception types?

gobradovic: Several of these were missing so we added them on develop branch. As we're retrofitting bunch of similar changes back to release 8, I thought it was a good idea to add the missing exceptions here as well.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Client/ExceptionConversionHandler.cs
```diff
@@ -3,30 +3,46 @@
...
-
     internal class ExceptionConversionHandler
     {
         private static readonly string TraceEventType = "ExceptionConversionHandler";
         private IEnumerable<IExceptionConvertor> convertors;
-        private FabricTransportRemotingSettings remotingSettings;
+        private IExceptionDeserializerSettings remotingSettings;
 
-        public ExceptionConversionHandler(IEnumerable<IExceptionConvertor> convertors, FabricTransportRemotingSettings remotingSettings)
+        public ExceptionConversionHandler(IEnumerable<IExceptionConvertor> convertors, IExceptionDeserializerSettings remotingSettings)
```
ilijagrbic: Can we make this constructor private, so that nobody uses it by mistake? 
Currently it is used in `CreateDefault` method and in `DummyFabricTransportRemotingClient` class. Perhaps the later usage can be removed.

src/Microsoft.ServiceFabric.Services.Remoting/V2/FabricTransport/Client/DummyFabricTransportRemotingClient.cs
```diff
@@ -20,7 +20,8 @@ public DummyFabricTransportRemotingClient(
                 serializersManager,
                 fabricTransportClient,
                 null,
-                FabricTransportRemotingSettings.GetDefault())
+                FabricTransportRemotingSettings.GetDefault(),
+                new ExceptionConversionHandler(null, null))
```
ilijagrbic: Can we use `CreateDefault` method here instead?

src/Microsoft.ServiceFabric.Services.Remoting/V2/FabricTransport/Runtime/FabricTransportServiceRemotingListener.cs
ilijagrbic, line 133: Can we use `CreateDefault()` here?

src/Microsoft.ServiceFabric.Services.Remoting/V2/Runtime/ExceptionConversionHandler.cs
```diff
@@ -122,41 +135,21 @@ public List<ArraySegment<byte>> SerializeRemoteException(RemoteException2 remote
             }
         }
 
+        internal RemoteException2 BuildRemoteException(Exception exception)
```
ilijagrbic, line 139: Consider making this method private.

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Runtime/WcfServiceRemotingListener.cs
```diff
@@ -38,22 +39,71 @@ public class WcfServiceRemotingListener : IServiceRemotingListener
...
+        ///     should be used to create the address for the listener. If the endpointResourceName is not specified or null,
+        ///     the default value "ServiceEndpointV2" is used.
         /// </param>
         /// <param name="useWrappedMessage">
-        /// It indicates whether the remoting method parameters should be wrapped or not before sending it over the wire. When UseWrappedMessage is set to false, parameters  will not be wrapped. When this value is set to true, the parameters will be wrapped.Default value is false.</param>
+        ///     It indicates whether the remoting method parameters should be wrapped or not before sending it over the wire.
+        ///     When UseWrappedMessage is set to false, parameters  will not be wrapped. When this value is set to true, the
+        ///     parameters will be wrapped.Default value is false.
+        /// </param>
+        [Obsolete("This constructor is obsolete. Use the one with exceptionConvertors parameter instead.")]
```
ilijagrbic: nit: Consider extracting this string since it repeats.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingListenerTest.cs
```diff
@@ -0,0 +1,82 @@
...
+
+                var mockServiceContext = new Mock<ServiceContext>(
+                    new NodeContext("NodeName", new NodeId(0, 1), 0, "NodeType", "127.0.0.1"),
+                    mockCodePackageActivationContext,
+                    "DummyServiceType",
+                    new Uri("fabric:/DummyApp/DummyService"),
+                    null,
+                    Guid.NewGuid(),
+                    1L
+                );
```
ilijagrbic, line 46: Not sure if I missed something, but we can use `fuzzy.ServiceContext()` instead of manually creating `ServiceContext` mock since as far as I can see, we don't verify anything on it nor do we need a specific `ServiceContext` for this test.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingClientFactoryTest.cs
```diff
@@ -0,0 +1,59 @@
...
+                IEnumerable<IExceptionConvertor> expectedConvertors = expectedExceptionDeserializer.Field<IEnumerable<IExceptionConvertor>>().Value;
+
+                Assert.NotNull(actualConvertors);
+
+                // Compare that the types of the convertors in both lists are the same
+                Assert.Equal(expectedConvertors.Count(), actualConvertors.Count());
+                for (int i = 0; i < actualConvertors.Count(); i++)
+                {
+                    Assert.Equal(expectedConvertors.ElementAt(i).GetType(), actualConvertors.ElementAt(i).GetType());
+                }
```
ilijagrbic: Perhaps we could try to remove the for loop. For example, we could do something like this:

```
var expectedTypes = actualExceptionDeserializer.Field<IEnumerable<IExceptionConvertor>>().Value.Select(c => c.GetType()).ToArray();
var actualTypes   = expectedExceptionDeserializer.Field<IEnumerable<IExceptionConvertor>>().Value.Select(c => c.GetType()).ToArray();

Assert.Equal(expectedTypes, actualTypes);
```

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfRemotingListenerSettingsTest.cs
```diff
@@ -0,0 +1,70 @@
...
+            [Fact]
+            public void ShouldInitializeRemotingExceptionDepthToDefaultValue()
+            {
+                Assert.Equal(ExceptionConversionHandler.DefaultRemotingExceptionDepth, sut.RemotingExceptionDepth);
+            }
+
+            [Fact]
+            public void ShouldImplementIExceptionSerializerSettings()
+            {
+                Assert.IsAssignableFrom<IExceptionSerializerSettings>(sut);
```
ilijagrbic, line 27: nit: Recently Oleg pointed out to me that these kinds of tests are not really needed and that the same assert can be implicitly accomplished but declaring the `sut` field as the interface type:
`IExceptionSerializerSettings sut = new WcfRemotingListenerSettings();`

test/unittests/Microsoft.ServiceFabric.Services.Tests/Remoting/FabricTransport/Runtime/FabricTransportRemotingListenerSettingsTest.cs
```diff
@@ -0,0 +1,69 @@
...
+            [Fact]
+            public void ShouldInitializeRemotingExceptionDepthToDefaultValue()
+            {
+                Assert.Equal(ExceptionConversionHandler.DefaultRemotingExceptionDepth, sut.RemotingExceptionDepth);
+            }
+
+            [Fact]
+            public void ShouldImplementIExceptionSerializerSettings()
+            {
+                Assert.IsAssignableFrom<IExceptionSerializerSettings>(sut);
```
ilijagrbic, line 26: Consider using implicit assert for this:
`IExceptionSerializerSettings sut = new FabricTransportRemotingListenerSettings();`

test/unittests/Microsoft.ServiceFabric.Services.Remoting.Tests/ExceptionSerializerTest.cs
```diff
@@ -0,0 +1,96 @@
...
+                // Act
+                ExceptionConversionHandler serializer = ExceptionConversionHandler.CreateDefault(null, new FabricTransportRemotingListenerSettings());
+
+                // Assert
+                IEnumerable<IExceptionConvertor> actualConvertors = serializer.Field<IEnumerable<IExceptionConvertor>>().Value;
+                Assert.Equal(expectedConvertors.Count, actualConvertors.Count());
+                for (int i = 0; i < expectedConvertors.Count; i++)
+                {
+                    Assert.IsType(expectedConvertors[i].GetType(), actualConvertors.ElementAt(i));
+                }
```
ilijagrbic: nit: I would try to avoid for loop for asserting, if possible.

test/unittests/Microsoft.ServiceFabric.Actors.Wcf.Tests/WcfActorServiceRemotingListenerTest.cs
```diff
@@ -0,0 +1,99 @@
...
+
+                var statefulServiceContext = new StatefulServiceContext(
+                    new NodeContext("NodeName", new NodeId(0, 1), 0, "NodeType", "127.0.0.1"),
+                    mockCodePackageActivationContext,
+                    "DummyServiceType",
+                    new Uri("fabric:/DummyApp/DummyService"),
+                    null,
+                    Guid.NewGuid(),
+                    1L
+                );
```
ilijagrbic, line 65: Unless there is a reason to construct `ServiceContext`, I would try to use `IFuzzy.ServiceContext()` instead.

src/Microsoft.ServiceFabric.Services.Remoting/V2/FabricExceptionKnownTypes.cs
ilijagrbic: Ah makes sense, got it. Perhaps we should mention the reason we are introducing them in the PR description.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Runtime/ExceptionConversionHandler.cs
gobradovic: It is used in test currently.

test/unittests/Microsoft.ServiceFabric.Actors.Wcf.Tests/AssemblyInfo.cs
gobradovic: It can be eliminated, but we need to add a fair bit of code to mock it. I've pushed the changes without `AssemblyInfo`, but I can revert back to this if we consider it cleaner/more readable. WDYT?

src/Microsoft.ServiceFabric.Services.Wcf/Remoting/V2/Wcf/Client/WcfServiceRemotingClientFactory.cs
gobradovic: Constructor with the same signature exists and is obsolete in develop as well.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Client/ExceptionConversionHandler.cs
gobradovic: We can (by adding `InternalsVisibleTo` for the tests that use it). However, I'm not sure if such change would make proper sense. As a user I'd be a bit confused if I can create only a "default" instance of an object. That _would_ make sense if we'd have two factory methods, one that adds default constructors, and one that doesn't. In any case, such refactoring makes much more sense on develop branch rather than on release branch.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingListenerTest.cs
gobradovic: We don't have `IFuzzExtensions` class on release 8 branch. We can do this on develop branch though.

test/unittests/Microsoft.ServiceFabric.Actors.Wcf.Tests/WcfActorServiceRemotingListenerTest.cs
gobradovic: We don't have `IFuzzExtensions` class on release 8 branch. We can do this on develop branch though.

test/unittests/Microsoft.ServiceFabric.Services.Tests/Remoting/FabricTransport/Runtime/FabricTransportRemotingListenerSettingsTest.cs
gobradovic: Interface doesn't allow setting of `RemotingExceptionDepth` which we do in almost all tests in this class. We'd have to cast `sut` to `FabricTransportRemotingListenerSettings` in all these places. I think that this would make it less readable so I'd rather keep it as is.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfRemotingListenerSettingsTest.cs
gobradovic: As discussed in other comment, (`FabricTransportRemotingListenerSettingsTest`), this approach would require us to cast `sut` to `WcfRemotingListenerSettings` almost every time we use it. I think that would reduce readability.

test/unittests/Microsoft.ServiceFabric.Services.Wcf.Tests/WcfServiceRemotingClientFactoryTest.cs
gobradovic: Good idea, done.

src/Microsoft.ServiceFabric.Services.Remoting/V2/Runtime/IExceptionSerializerSettings.cs
```diff
@@ -0,0 +1,17 @@
...
+// ------------------------------------------------------------
+
+using Microsoft.ServiceFabric.Services.Remoting.FabricTransport.Runtime;
+
+namespace Microsoft.ServiceFabric.Services.Remoting.V2.Runtime
+{
+    interface IExceptionSerializerSettings
+    {
+        int RemotingExceptionDepth { get; }
+        #pragma warning disable 618
```
olegsych: Just a nit, since we're doing this in the release branch and it's fine as is. VS insists on placing compiler directives in the beginning of the line, at least it does for #ifs. Because of this, I think the typical code style is to place directives at the start of the line and keep the code alignment as is. In other words, the ExceptionSerializationTechnique property would be at the same level as the RemotingExceptionDepth. It kind of makes sense logically as well, since they are both properties of this interface.

olegsych (approved):
:shipit:

PR #489, opened 2025-10-09
Refactor Actors performance counter emission

ilijagrbic:
Refactor the way we emit diagnostic data in Actors. Implement more testable abstraction in order to prepare a path for Metrics instrumentation.
- Delete old `DiagnosticManager`/`DiagnosticEventManager` and clean up old code
- Introduce `IDiagnostics`
- Implement composite pattern in `AggregatedDiagnoticEvents`
- Implement `EventSourceDiagnoticEvents` and `PerformanceCounterDiagnoticEvents`
- Replace old `DiagnoticManager` usage with new `AggregatedDiagnoticEvents`


src/Microsoft.ServiceFabric.Actors/Diagnostics/AggregatedDiagnosticEvents.cs
```diff
@@ -0,0 +1,176 @@
...
+            if (diagnosticEvents.Any(d => d == null))
+            {
+                throw new ArgumentException(nameof(diagnosticEvents));
+            }
+
+            this.diagnosticEvents = diagnosticEvents;
+        }
+        public void AcquireActorLockFailed(DiagnosticsManagerActorContext diagnosticContext)
+        {
+            Interlocked.Decrement(ref diagnosticContext.PendingActorMethodCalls);
```
ilijagrbic: This is how PendingMethodCalls were counted in the previous DiagnosticsManager. Think about if there is a better way to do this.

src/Microsoft.ServiceFabric.Actors/Diagnostics/Util.cs
```diff
@@ -5,9 +5,13 @@
...
+    internal class Util
     {
-        internal bool IsActivationEvent;
-        internal ActorId ActorId;
+        public static long GetInterfaceMethodKey(uint interfaceId, uint methodId)
+        {
+            var key = (ulong)methodId;
+            key = key | (ulong)interfaceId << 32;
+            return (long)key;
+        }
```
ilijagrbic, line 15: Consider a better place for this logic.

src/Microsoft.ServiceFabric.Actors/Runtime/ActorManager.cs
```diff
@@ -34,26 +35,40 @@ internal sealed class ActorManager : IActorManager
...
         private readonly ConcurrentDictionary<ActorId, ActorBase> activeActors;
         private readonly ConcurrentDictionary<ActorId, ConcurrentDictionary<string, ActorReminder>> remindersByActorId;
-        private readonly DiagnosticsEventManager diagnosticsEventManager;
         private readonly IActorEventManager eventManager;
-        private IDiagnosticsManager diagnosticsManager;
         private bool isClosed;
 
+        private readonly IDiagnosticEvents diagnosticEvents;
+        private readonly IClock clock;
+        private static Func<ActorService, IClock, IDiagnosticEvents> createDiagnosticEvents = (actorService, clock) =>
```
ilijagrbic: We need this static field in order to mock `diagnosticEvents` in certain test paths, where `ActorManager` is not injected but rather created during the process.

test/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/PerformanceCounterDiagnosticEventsTest.cs
```diff
@@ -0,0 +1,367 @@
...
+            }
+
+            public class ActorState : OnEvents
+            {
+                [Fact]
+                public void SaveEmitsPerfCounter()
+                {
+                    sut.SaveActorStateFinish(actorId, startTime);
+
+                    Mock.Get(actorSaveStateTimeCounterWriter).Verify(p => p.UpdateCounterValue(It.Is<ActorStateDiagnosticData>(data => data.OperationTime.Value.TotalMilliseconds == operationDurationMillis && data.ActorId == actorId)), Times.Once);
```
ilijagrbic: TODO: Use Equals here instead of manually checking fields.

src/Microsoft.ServiceFabric.Actors/Diagnostics/Data/ActorMethodDiagnosticData.cs
olegsych, line 1: We should keep the folder structure consistent with the namespace structure for maintainability.
With only three types, I don't think we have enough to justify the cognitive overhead of an extra sub-namespace/folder for Data. .NET used to have an FxCop rule advising against namespaces with fewer than 5 types, if memory serves me well. To be on the safe side, my rule of thumb to decide that a new namespace is needed once the number of candidate types reaches 7.

src/Microsoft.ServiceFabric.Actors/Diagnostics/PerformanceCounters/ActorMethodExceptionFrequencyCounterWriter.cs
```diff
@@ -11,14 +11,15 @@ namespace Microsoft.ServiceFabric.Actors.Diagnostics
     // frequency at which a particular actor method throws exceptions.
     internal class ActorMethodExceptionFrequencyCounterWriter : FabricBaselessPerformanceCounterWriter
     {
+        protected ActorMethodExceptionFrequencyCounterWriter() { }
```
olegsych, line 14: Could you double-check the code style? This constructor is missing a blank line.

src/Microsoft.ServiceFabric.Actors/Diagnostics/PerformanceCounters/ActorLockContentionCounterWriter.cs
```diff
@@ -11,14 +11,16 @@ namespace Microsoft.ServiceFabric.Actors.Diagnostics
     // number of pending actor calls that are waiting for the actor lock.
```
olegsych, line 11: We have enough types to justify a separate folder for PerfCounters, but let's add a matching sub-namespace

src/Microsoft.ServiceFabric.Actors/Diagnostics/PerformanceCounters/ActorLockContentionCounterWriter.cs
```diff
@@ -11,14 +11,16 @@ namespace Microsoft.ServiceFabric.Actors.Diagnostics
...
     internal class ActorLockContentionCounterWriter : FabricBaselessPerformanceCounterWriter
     {
+        protected ActorLockContentionCounterWriter() { }
```
olegsych, line 14: Why do we need protected constructors? I'm guessing we need to dynamically subclass them in Moq, but doesn't the InternalsVisibleTo already cover it?

src/Microsoft.ServiceFabric.Actors/Diagnostics/PerformanceCounters/ActorLockContentionCounterWriter.cs
```diff
@@ -11,14 +11,16 @@ namespace Microsoft.ServiceFabric.Actors.Diagnostics
...
+
         internal ActorLockContentionCounterWriter(FabricPerformanceCounterSetInstance counterSetInstance)
             : base(
                 counterSetInstance,
                 ActorPerformanceCounters.ActorCallsWaitingForLockCounterName)
         {
         }
 
-        internal void UpdateCounterValue(PendingActorMethodDiagnosticData pendingMethodData)
+        internal virtual void UpdateCounterValue(PendingActorMethodDiagnosticData pendingMethodData)
```
olegsych, line 23: I wonder if we should add a one-line comment to the top of the class to explain that it's not sealed to enable dynamic mocking.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodInfo.cs
```diff
@@ -0,0 +1,24 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System.Reflection;
+
+namespace Microsoft.ServiceFabric.Actors.Diagnostics
+{
+    internal class ActorMethodInfo
```
olegsych: Could you omit `internal` in new code? C# is a very well-designed language and defaults,, such as types are internal unless specified otherwise, help to reduce the noise and make the code more readable. 

On the other hand, we should make classes sealed as much as possible to communicate the whether we intend them to be subclassed.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodInfo.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        internal ActorMethodInfo() { }
```
olegsych: This constructor appears to be redundant because C# should provide an empty constructor by default.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodInfo.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        internal ActorMethodInfo(MethodInfo methodInfo)
+        {
+            MethodName = string.Concat(methodInfo.DeclaringType.Name, ".", methodInfo.Name);
+            MethodSignature = methodInfo.ToString();
+        }
+
+        internal string MethodName { get; set; }
```
olegsych: Can we define these as read-only fields instead of read-write properties? The design guideline recommending properties shouldn't apply to internal APIs.

src/Microsoft.ServiceFabric.Actors/Diagnostics/AggregatedDiagnosticEvents.cs
```diff
@@ -0,0 +1,176 @@
...
+using System.Fabric;
+using System.Linq;
+using System.Threading;
+using Microsoft.ServiceFabric.Services.Remoting;
+
+namespace Microsoft.ServiceFabric.Actors.Diagnostics
+{
+    internal class AggregatedDiagnosticEvents : IDiagnosticEvents
+    {
+        readonly IEnumerable<IDiagnosticEvents> diagnosticEvents;
```
olegsych: Could you double-check for blank lines between type members?

src/Microsoft.ServiceFabric.Actors/Diagnostics/AggregatedDiagnosticEvents.cs
```diff
@@ -0,0 +1,176 @@
...
+        public AggregatedDiagnosticEvents(IEnumerable<IDiagnosticEvents> diagnosticEvents)
+        {
+            _ = diagnosticEvents ?? throw new ArgumentNullException(nameof(diagnosticEvents));
+            if (diagnosticEvents.Any(d => d == null))
+            {
```
olegsych: I would love to eliminate curly braces around single line statements. The bloat is particularly noticeable in a class like this.

src/Microsoft.ServiceFabric.Actors/Diagnostics/AggregatedDiagnosticEvents.cs
olegsych: This definitely smells like an unrelated responsibility in the context of the AggregatedDiagnosticEvents. 
This could be a distinct class implementating of the IDiagnosticEvents with the single responsibility of tracking the PendingActorMethodCalls, but since only two methods would be doing the actual work, it feels like an overkill. Why couldn't this be done by the caller?

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnosticEvents.cs
```diff
@@ -0,0 +1,189 @@
...
+        readonly ServiceContext serviceContext;
+        readonly IClock clock;
+        readonly string actorType;
+        readonly ActorFrameworkEventSource eventSource;
+        readonly Dictionary<long, ActorMethodInfo> actorMethodInfo;
+
+        public EventSourceDiagnosticEvents(ActorFrameworkEventSource eventSource, IClock clock, ServiceContext serviceContext, ActorMethodFriendlyNameBuilder nameBuilder, ActorTypeInformation typeInfo)
+        {
+            this.eventSource = eventSource ?? throw new ArgumentNullException(nameof(eventSource));
+            this.serviceContext = serviceContext;
```
olegsych: Shouldn't we throw on null ServiceContext as well?

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnosticEvents.cs
```diff
@@ -0,0 +1,189 @@
...
+            this.clock = clock ?? throw new ArgumentNullException(nameof(clock));
+            _ = nameBuilder ?? throw new ArgumentNullException(nameof(nameBuilder));
+            _ = typeInfo ?? throw new ArgumentNullException(nameof(typeInfo));
+            this.actorType = typeInfo.ImplementationType.ToString();
+            this.actorMethodInfo = InitializeActorMethodInfo(nameBuilder, typeInfo);
+        }
+
+        private Dictionary<long, ActorMethodInfo> InitializeActorMethodInfo(ActorMethodFriendlyNameBuilder nameBuilder, ActorTypeInformation typeInfo)
```
olegsych: Same ask to avoid specifying private for type members, which are private by default.

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnosticEvents.cs
```diff
@@ -0,0 +1,189 @@
...
+        {
+            var actorMethodInfo = new Dictionary<long, ActorMethodInfo>();
+
+            foreach (var actorInterfaceType in typeInfo.InterfaceTypes)
```
olegsych: What is the actual type of 'var` here?

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnosticEvents.cs
olegsych: Does InitializeActorMethodInfo initialize this.actorMethodInfo or only builds it? Can this method be static?

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnosticEvents.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        public void AcquireActorLockFailed(DiagnosticsManagerActorContext diagnosticContext)
+        {
+            // Intentionally left blank, since we don't track
+        }
+
+        public void AcquireActorLockFinish(PendingActorMethodDiagnosticData diagnosticData, DateTime startTime)
+        {
+            if (this.eventSource.IsPendingMethodCallsEventEnabled())
```
olegsych: We can omit this. without losing any readability in small methods like this. I always omit this. and when the method starts to feel like it needs it, that's a sign to break it up.

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnosticEvents.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        public void ActorMethodFinish(DateTime startTime, ActorId actorId, long interfaceMethodKey, Exception e, RemotingListenerVersion remotingListener)
+        {
+            var methodInfo = this.actorMethodInfo[interfaceMethodKey];
+
+            if (e != null)
+            {
+                this.eventSource.ActorMethodThrewException(
+                   e.ToString(),
```
olegsych: Ugh... I really dislike the convention of placing method call parameters on separate lines. As soon as we start doing this, we throw away an important complexity indicator and ignore the opportunity to improve the method design. Offhand, this method smells of "primitives obsession" and asking to take an Exception instead of a string, a DateTime instead of a long, an ActorMethodInfo instead of two strings.

src/Microsoft.ServiceFabric.Actors/Remoting/V2/Runtime/ActorServiceRemotingDispatcher.cs
```diff
@@ -26,8 +26,8 @@ namespace Microsoft.ServiceFabric.Actors.Remoting.V2.Runtime
     public class ActorServiceRemotingDispatcher : ServiceRemotingMessageDispatcher
     {
         private static readonly string TraceType = typeof(ActorServiceRemotingDispatcher).Name;
-        private readonly ActorService actorService;
-        private readonly ServiceRemotingCancellationHelper cancellationHelper;
+        private ActorService actorService;
```
olegsych: Why are we making these fields writeable? If this is for tests, then readonly instance fields can be modified by reflection, only readonly static fields cannot be.

src/Microsoft.ServiceFabric.Actors/Runtime/ActorStateManager.cs
```diff
@@ -12,21 +12,27 @@ namespace Microsoft.ServiceFabric.Actors.Runtime
...
+        readonly ActorBase actor;
+        readonly IClock clock;
+        readonly IDiagnosticEvents diagnosticEvents;
 
         internal ActorStateManager(ActorBase actor, IActorStateProvider actorStateProvider)
         {
             this.actor = actor;
             this.stateProvider = actorStateProvider;
             this.stateChangeTracker = new Dictionary<string, StateMetadata>();
+            this.clock = actor.ActorService.Clock;
```
olegsych: This is breaks the law of Demeter and creates a tight coupling between components. We should strive to maintain direct dependencies only. For internal classes, this would be as straightforward as declaring additional constructor parameters.

test/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/PerformanceCounterDiagnosticEventsTest.cs
```diff
@@ -0,0 +1,369 @@
...
+
+                var clockField = sut.Field<IClock>().Value;
+                Assert.Equal(clock, clockField);
+            }
+
+            [Fact]
+            public void ThrowsOnNullProvider()
+            {
+                var exception = Assert.Throws<ArgumentNullException>(() =>
+                {
```
olegsych: Could you simplify the lambdas? Even if you don't want to place it on a single line, there is no need or benefit to the curly braces.

olegsych (commented):
Hey @ilijagrbic, this is a large change and I haven't been able to finish thoroughly reviewing it. The comments below are mostly small style and design nits. 

At a higher level, I'm not sure replacing DiagnosticsEventManager with the IDiagnosticEvents hierarchy is worth the risk of such a significant change. Other than consistency with the design we're using in other projects, what are the comparative advantages here? The old design is based on .NET delegates which already implement the composite pattern and don't require implementing AggregatedDiagnosticEvents. Otherwise, but both designs appear to produce the same number of classes and could be fully tested. Doesn't that make the delegate-based design simpler?

ilijagrbic:
Hey @olegsych , thank you for the review. There are couple of points I feel like are better in this design:
- There is smaller number of classes for this design. We now don't have some specific data classes only used to repack method parameters. We also don't have two `EventSourceProviders`. Perhaps total number of files seems the same since I extracted some classes into separate files, such as Util.
- Previously logic in `DiagnosticEventManager` used to change struct values in `DiagnoticContext` before passing it allong to `EventSource`/`PerfCounterProviders`.
- We removed entire `Stopwatch `logic. It was in my opinion overly complex, especially for nested method call and really not needed if replaced by the pattern we used in Remoting.
- Both patterns do essentially achieve the same thing, I just think new pattern is easier to test and mock.
- I also think there is value in consistency across our projects, especially for such a similar process. Future readers would understand it easier and would not need to wonder why we have such different architecture for the same thing.

On the other hand, I see where your concerns come from, I had similar concerns and entire change grew a bit too much I feel. We can discuss if this refactor is worth the risk and if not, we can always move on instrumenting Actors code by using the old pattern.

src/Microsoft.ServiceFabric.Actors/Diagnostics/AggregatedDiagnosticEvents.cs
ilijagrbic: Thanks for the comment, Oleg, this is a good point. Another reason I didn't want to implement it as an another `IDiagnoticEvents` since in that case logic for other `IDiagnoticEvents` would depend on their ordering in the `AggregatedDiagnoticEvents` list, which would me an implicit constraint which could easily be broken by accident. I think we can move this to be done by the caller, it would be just a bit harder to test, but I will look into moving it there.

src/Microsoft.ServiceFabric.Actors/Diagnostics/Data/ActorMethodDiagnosticData.cs
ilijagrbic: Good point, I will move them to the parent namespace, thanks.

src/Microsoft.ServiceFabric.Actors/Diagnostics/PerformanceCounters/ActorLockContentionCounterWriter.cs
ilijagrbic: I think that `InternalsVisibleTo` would allow `Moq` to access internal methods/fields. However, since we defined a constructor with parameters, no default constructor without parameters is present in this class and I needed it since I didn't want to pass `FabricPerformanceCounterSetInstance` in order to create a Mock object. Anyways that's why I defined this protected constructor.

src/Microsoft.ServiceFabric.Actors/Diagnostics/PerformanceCounters/ActorLockContentionCounterWriter.cs
ilijagrbic: Good idea, I will add a comment explaining this.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodInfo.cs
ilijagrbic: Maybe I'm wrong about this but I think default constructor is provided only if there are no other constructors defined. But I will look into if I can remove the usage of this altogether.

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnosticEvents.cs
ilijagrbic: Yeah, that is correct, I will add a check for this.

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnosticEvents.cs
ilijagrbic: It's `System.Type`, I thought it is self-explanatory from the variable name, but I will add explicit types here to make it more clear.

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnosticEvents.cs
ilijagrbic: Good catch, initially it was initializing the field, but I refactored it in order to be able to make the field `readonly`. Anyways, I will rename the method and change it to be static. :)

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnosticEvents.cs
ilijagrbic: Good point, I will remove it where it is not needed (which should be everywhere except for constructor).

src/Microsoft.ServiceFabric.Actors/Remoting/V2/Runtime/ActorServiceRemotingDispatcher.cs
ilijagrbic: Ahhh,, sorry, I was removing the `private` modifier on a bunch of fields, and I removed `readonly` by mistake here...

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodInfo.cs
ilijagrbic: Since this class is really simple, I have moved the logic from this constructor to the caller and now the class just sets fields to passed parameters. I also removed constructor without parameters and set properties.

src/Microsoft.ServiceFabric.Actors/Runtime/ActorStateManager.cs
ilijagrbic: The way I was thinking about it was that either I will access the `IClock/IDiagnoticEvents` here or I will inject them from `ActorService`. However, in `ActorService` I would basically have the same logic (`actor.ActorService.Clock` and `actor.Manager.DiagnosticsEvents`) or I would have to change the public API. Tight coupling already existed here, since we were accessing old `DiagnoticManager` in a similar manner.

Anyways, I see a benefit for injecting these anyways, particularly in the test code, where I had to set some fields in `ActorService` in order to test the `ActorStateManager`. So, I will inject these as constructor parameters and move the logic to access them to `ActorService`, let me know what you think about this. :)

src/Microsoft.ServiceFabric.Actors/Diagnostics/AggregatedDiagnosticEvents.cs
ilijagrbic: I'm planning to move the logic for this into the `DiagnosticsManagerActorContext` class, and unit test it that way. I will probably finish that up on Monday. :)

ilijagrbic: In the end I moved the logic to the `DiagnosticsManagerActorContext`, which also allowed to me to remove some not needed APIs from the `IDiagnoticEvents` (`AcquireActorLockStart` and `AcquireActorLockFailed`) which were only needed in order to increment/decrement pending calls.

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnosticEvents.cs
ilijagrbic: I agree on the number of parameters and on this method. I will refactor this method to accept `ActorMethodDiagnosticData` instead of multiple primitive parameters. However, I will keep the call to `ActorFrameworkEventSource` as it is, since that would be yet another change in an already big PR.

test/Microsoft.ServiceFabric.Actors.Tests/Runtime/ActorManagerTests.cs
```diff
@@ -0,0 +1,343 @@
...
+        public ActorManagerTests()
+        {
+            actorId = ActorId.CreateRandom();
+            actorService = TestMocksRepository.GetActorService<MockActor>();
+
+            var friendlyNameBuilder = new ActorMethodFriendlyNameBuilder(actorService.ActorTypeInformation);
+            actorService.InitializeInternal(friendlyNameBuilder);
+        }
+
+        public class Remainder : ActorManagerTests
```
ilijagrbic: Note: `Remainder` class contains only old tests for `ActorManager`. They are shown as new code since the file was moved to this package.

src/Microsoft.ServiceFabric.Actors/Diagnostics/PerformanceCounters/ActorLockContentionCounterWriter.cs
```diff
@@ -3,22 +3,24 @@
...
         {
```
olegsych, line 24: Forgot to mention last time. When I was thinking about how to unit test this code, I though we could use Inspector to get to the FabricPerformanceCounter obtained by the base class and use its GetValue method. This is not as invasive as making the UpdateCounterValue methods virtual and would be closer to the state-based testing, which is generally more maintainable than behavior-based testing we do with mocks. It's a pretty wide gray area though, so just a suggestion to consider.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorFrameworkEventSource.cs
```diff
@@ -11,17 +11,20 @@
...
+    // method calls in the unit tests.
     [EventSource(Name = "Microsoft-ServiceFabric-Actors", LocalizationResources = "Microsoft.ServiceFabric.Actors.SR", Guid = "0e1ec353-9f02-55d7-fbb8-f3857458acbd")]
-    internal sealed class ActorFrameworkEventSource : ServiceFabricEventSource
+    internal class ActorFrameworkEventSource : ServiceFabricEventSource
     {
         internal static ActorFrameworkEventSource Writer { get; private set; } = new ActorFrameworkEventSource();
 
         [NonEvent]
-        internal void ReplicaChangeRoleToPrimary(ServiceContext serviceContext)
+        internal virtual void ReplicaChangeRoleToPrimary(ServiceContext serviceContext)
```
olegsych, line 27: I don't remember if I suggested before and you probably already know that we can observe event method calls using EventListener without making the "NonEvent" methods virtual. However, verifying the high-level event methods with mocks is much more concise so overall, I think this approach is *simpler*.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodDiagnosticData.cs
```diff
@@ -12,8 +12,16 @@ internal struct ActorMethodDiagnosticData
     {
         internal ActorId ActorId;
         internal long InterfaceMethodKey;
-        internal TimeSpan? MethodExecutionTime;
         internal Exception Exception;
         internal RemotingListenerVersion RemotingListener;
+        internal TimeSpan? MethodExecutionTime;
+
+        internal ActorMethodDiagnosticData(ActorId actorId, long interfaceMethodKey, Exception exception, RemotingListenerVersion remotingListener)
```
olegsych: Now that we have a constructor, can we make the fields read-only as well?

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodDiagnosticData.cs
```diff
@@ -12,8 +12,16 @@ internal struct ActorMethodDiagnosticData
...
+        {
```
olegsych: It's odd that we don't have the method execution time with the exception. I could see them being mutually exclusive, but there's value in measuring the execution time even when an exception is thrown.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodDiagnosticData.cs
```diff
@@ -12,8 +12,16 @@ internal struct ActorMethodDiagnosticData
...
+            ActorId = actorId;
```
olegsych: I'm not sure why we're adding this constructor, but since we're doing it, don't we want to validate the parameters? I assume we'd want to check ActorId and Exception parameters for null values.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodInfo.cs
```diff
@@ -0,0 +1,23 @@
...
+        readonly string methodName;
+        readonly string methodSignature;
+
+        internal ActorMethodInfo(string methodName, string methodSignature)
+        {
+            this.methodName = methodName;
+            this.methodSignature = methodSignature;
+        }
+
+        internal string MethodName => methodName;
```
olegsych: Since this class is internal, I would make MethodName and MethodSignature internal readonly fields rather than properties. In API design, properties allow changing public APIs in the future, so I prefer the simplicity of readonly fields. But you could make a case for internal APIs due to our excessive use of InternalsVisibleTo. So, if you strongly prefer properties, make them get-only and remove the fields.

src/Microsoft.ServiceFabric.Actors/Diagnostics/AggregatedDiagnosticEvents.cs
```diff
@@ -0,0 +1,145 @@
...
+
+namespace Microsoft.ServiceFabric.Actors.Diagnostics
+{
+    sealed class AggregatedDiagnosticEvents : IDiagnosticEvents
+    {
+        readonly IEnumerable<IDiagnosticEvents> diagnosticEvents;
+
+        internal AggregatedDiagnosticEvents(IEnumerable<IDiagnosticEvents> diagnosticEvents)
+        {
+            _ = diagnosticEvents ?? throw new ArgumentNullException(nameof(diagnosticEvents));
```
olegsych, line 19: My OCD compels me to dislike using a throw-away variable assignment to eliminate an extra line of an if statement. I'm not sure it's rational though. Do you know if this is optimized away by the compiler?

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorFrameworkEventSource.cs
```diff
@@ -140,13 +143,13 @@ internal void ActorMethodStart(
...
-        internal bool IsActorMethodStopEventEnabled()
+        internal virtual bool IsActorMethodStopEventEnabled()
         {
             return this.IsEnabled(EventLevel.Verbose, Keywords.ActorMethod);
         }
 
         [NonEvent]
-        internal void ActorMethodStop(
+        internal virtual void ActorMethodStop(
            long methodExecutionTimeTicks,
```
olegsych, line 153: I think TimeSpan to ticks conversion would be *simpler* to encapsulate here rather than in the EvenSourceDiagnosticEvents

test/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/DiagnosticsManagerActorContextTest.cs
```diff
@@ -0,0 +1,49 @@
+using System;
+using Fuzzy;
+using Inspector;
+using Xunit;
+
+namespace Microsoft.ServiceFabric.Actors.Diagnostics
+{
+    public class DiagnosticsManagerActorContextTest
```
olegsych: Let's make the test class match the name of the SUT.

src/Microsoft.ServiceFabric.Actors/Diagnostics/DiagnosticActorContext.cs
```diff
@@ -0,0 +1,32 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System.Threading;
+
+namespace Microsoft.ServiceFabric.Actors.Diagnostics
+{
+    internal class DiagnosticActorContext
```
olegsych: This reads "Context of a Diagnostic Actor". I think "ActorDiagnosticsContext" would be more accurate and consistent with the ActorBase.DiagnosticsContext property.

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnosticEvents.cs
```diff
@@ -0,0 +1,175 @@
...
+                   TicksSinceStart(startTime),
+                   methodInfo.MethodName,
+                   methodInfo.MethodSignature,
+                   actorType,
+                   actorMethodDiagnosticData.ActorId,
+                   serviceContext);
+                return;
+            }
+
+            if (eventSource.IsActorMethodStopEventEnabled())
```
olegsych, line 99: I wonder if this logic would be simpler to encapsulate inside the eventSource.ActorMethodStop.

src/Microsoft.ServiceFabric.Actors/Runtime/ActorBase.cs
```diff
@@ -99,7 +102,7 @@ public ActorService ActorService
 
         internal bool IsInitialized { get; set; }
 
-        internal DiagnosticsManagerActorContext DiagnosticsContext
+        internal DiagnosticActorContext DiagnosticsContext
```
olegsych: Context is a filler word. Since we're changing the property type, I would consider dropping it and naming the property simply Diagnostics.

src/Microsoft.ServiceFabric.Actors/Runtime/ActorBase.cs
```diff
@@ -120,9 +123,12 @@ internal IActorManager Manager
 
         internal async Task OnActivateInternalAsync()
         {
-            this.Manager.DiagnosticsEventManager.ActorOnActivateAsyncStart(this);
+            var startTime = clock.UtcNow;
```
olegsych: Let's use explicit types instead of var.

src/Microsoft.ServiceFabric.Actors/Runtime/ActorStateManager.cs
```diff
@@ -12,21 +12,27 @@ namespace Microsoft.ServiceFabric.Actors.Runtime
...
+        readonly IClock clock;
+        readonly IDiagnosticEvents diagnosticEvents;
 
-        internal ActorStateManager(ActorBase actor, IActorStateProvider actorStateProvider)
+        internal ActorStateManager(ActorBase actor, IActorStateProvider actorStateProvider, IDiagnosticEvents diagnosticEvents)
         {
             this.actor = actor;
             this.stateProvider = actorStateProvider;
             this.stateChangeTracker = new Dictionary<string, StateMetadata>();
+            this.clock = new SystemClock();
```
olegsych, line 34: We should inject IClock instead of creating new instances in different classes. Although we can use Inspector/Reflection to manipulate the field, expressing the dependency explicitly, as a constructor parameter is better.

src/Microsoft.ServiceFabric.Actors/Runtime/ActorManager.cs
```diff
@@ -34,26 +35,41 @@ internal sealed class ActorManager : IActorManager
...
-        private readonly DiagnosticsEventManager diagnosticsEventManager;
         private readonly IActorEventManager eventManager;
-        private IDiagnosticsManager diagnosticsManager;
         private bool isClosed;
 
+        private readonly IDiagnosticEvents diagnosticEvents;
+        private readonly IClock clock;
+        private readonly PerformanceCounterProviderV2 performanceCounterProvider;
+
+        private static Func<ActorService, IClock, PerformanceCounterProviderV2, IDiagnosticEvents> createDiagnosticEvents = (actorService, clock, performanceCounterProvider) =>
```
olegsych: I think this logic better fits in the ActorService class because it also creates the ActorStateManager, which also needs IDiagnosticEvents. So, having it here introduces unnecessary coupling. For example, ActorManager creates and disposes the IDiagnosticEvents, but ActorStateManager uses it and that dependency is not visible here. This class could decide to dispose and recreate it breaking the ActorStateManager. However, if the ActorService creates an instance and gives it to both ActorManager and ActorStateManager, the intent would be clear.

ActorService is also a better fit for the role of the *composition root*, so I would expect dependencies like IDiagnosticEvents, IClock, etc to bubble up to it.

src/Microsoft.ServiceFabric.Actors/Diagnostics/IDiagnosticEvents.cs
```diff
@@ -0,0 +1,43 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using System.Fabric;
+
+namespace Microsoft.ServiceFabric.Actors.Diagnostics
+{
+    internal interface IDiagnosticEvents
```
olegsych: After commenting on the name of the ActorBase.DiagnosticsContext property, it occurred to me that the term Events is also redundant. So the interface name could be simplified to *IDiagnostics* without losing descriptiveness.

test/Microsoft.ServiceFabric.Actors.Tests/Runtime/ActorManagerTests.cs
```diff
@@ -0,0 +1,345 @@
...
+using Inspector;
+using Microsoft.ServiceFabric.Actors.Diagnostics;
+using Microsoft.ServiceFabric.Actors.Tests;
+using Microsoft.ServiceFabric.Diagnostics;
+using Moq;
+using Xunit;
+
+namespace Microsoft.ServiceFabric.Actors.Runtime
+{
+    public class ActorManagerTests
```
olegsych: Test class names should be singular - ActorManagerTest, otherwise the test names don't make sense in English.

test/Microsoft.ServiceFabric.Actors.Tests/ActorServiceRemotingDispatcherTest.cs
```diff
@@ -0,0 +1,67 @@
...
+            sut.Field<IClock>().Set(clock);
+        }
+
+        public class DiagnosticEvents : ActorServiceRemotingDispatcherTest
+        {
+            readonly Func<IActorRemotingMessageHeaders, IServiceRemotingRequestMessageBody, CancellationToken, Task<IServiceRemotingResponseMessageBody>> handleActorMethodDispatchAsync;
+
+            public DiagnosticEvents()
+            {
+                var method = sut.GetType().GetMethod("HandleActorMethodDispatchAsync",
```
olegsych: Ugh. Can't we use Inspector to make this more compact?

Something along the lines of https://github.com/olegsych/inspector/blob/de276bbfb1357bf9415bd35f07daeb7b138c88a5/examples/ObjectAccessExample.cs#L104?

test/Microsoft.ServiceFabric.Actors.Tests/AssemblyInfo.cs
```diff
@@ -3,10 +3,16 @@
...
+using static Microsoft.ServiceFabric.Constants.AssemblyInfo;
+
 
 // This attribute is used to run tests in the same assembly in sequence.
 // It is necessary for ActorRemotingProviderAttributeTest suit to run properly,
 // because ActorRemotingProviderAtrribut has static state and cannot be teste in parallel.
 [assembly: CollectionBehavior(CollectionBehavior.CollectionPerAssembly)]
 [assembly: CaptureConsole]
+
+// Used to test mocks
```
olegsych, line 17: Although it's not immediately obvious that DynamicProxyGenAssembly is used for mocking, it's pretty close; plus we have a doc comment for the constant explaining this. I would remove the line comment.

olegsych (approved):
I've completed another review pass @ilijagrbic. Aside from naming refinements, my main suggestion is to use ActorService as the composition root and create the IDiagnosticEvents objects there instead of the ActorManager and express dependencies explicitly with constructor parameters. The PR already feels too large, so feel free to merge it and iterate.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorFrameworkEventSource.cs
ilijagrbic: Yeah, I have decided to use the simpler approach here. This way a boundary of what is being testes is clearer and I don't have to know anything about the internals of the `ActorFrameworkEventSource`, but a good suggestion nonetheless. :)

src/Microsoft.ServiceFabric.Actors/Diagnostics/AggregatedDiagnosticEvents.cs
ilijagrbic: Main reason was to avoid having an extra line of an `if` statement. I'm kind of struggling to decide which of those two options were better and I got PR comments lately that `if` can be avoided this way. Anyways, I started to like the uniformity of how parameters are checked now.

src/Microsoft.ServiceFabric.Actors/Diagnostics/DiagnosticActorContext.cs
ilijagrbic: I like this suggestion, this name was kind of a left over from the old name for this class. I will rename the class into `DiagnosticContext` though, since I think it is implied that this context is related to Actor when it is its member.

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnosticEvents.cs
ilijagrbic: I agree, I will keep it outside of this PR, but I will address this in the future.

src/Microsoft.ServiceFabric.Actors/Diagnostics/IDiagnosticEvents.cs
ilijagrbic: I like how this is an example of "less is more" when it comes to naming things. We converged to a one-word name after thinking about way more complicated names in the beginning.

test/Microsoft.ServiceFabric.Actors.Tests/Runtime/ActorManagerTests.cs
ilijagrbic: I completely missed this since it was a preexisting class. I will rename it.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorFrameworkEventSource.cs
ilijagrbic: I agree with this, I will address this in a follow up PR together with other logic that should be moved into the EventSource class.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodInfo.cs
ilijagrbic: I was unsure about this as well and I mostly added the properties to avoid having externally visible fields, since that seemed wrong to me. Now that I look at it again, the properties seem rather artificial in this place, so I will take your suggestion and make the fields internal.

test/Microsoft.ServiceFabric.Actors.Tests/ActorServiceRemotingDispatcherTest.cs
ilijagrbic: Initially I had some issue with implementing the Inspector approach here, so I settled to just use reflection. I managed to get it to work now, so I will implement this suggestion. :)

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodDiagnosticData.cs
ilijagrbic: Main reason why I added this constructor is to improve readability. Using for example `new ActorMethodDiagnosticData() { actorId = ActorId, ... }` often lead to longer line length than I would like to have. I will probably revert this change as I think it raised more questions than it helped solve.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodDiagnosticData.cs
ilijagrbic: I wanted to keep the responsibility of calculating the execution time in the respective `IDiagnotic` classes, so that's why this constructor doesn't accept that info. Ideally, I would like to change the signature of PerfCounter methods to not accept this structure, since they actually don't use most of the data from it. This would also enable me to remove the execution time from this struct and make all other fields `readonly`.

Anyways, I will remove the constructor, and I will tackle this in the follow up PR.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodDiagnosticData.cs
ilijagrbic: I removed the constructor, and I will tackle this issue in the follow up PR.

src/Microsoft.ServiceFabric.Actors/Diagnostics/PerformanceCounters/ActorLockContentionCounterWriter.cs
ilijagrbic: Thanks for the comment, I also think that state-based testing is a more maintainable approach. However, it would require us to introduce some changes into the PerfCounter classes. While these changes would be minor (for example initializing `FabricPerformanceCounter` in constructors without fields), I didn't feel like doing this and going through tedious release process for WindowsFabric NuGet packages is worth the improvements we would get that way. 
Also, I try to avoid interacting too much with the internals of classes that are not being tested since I think that introduces unnecessary complexity. I'll think about this some more, and we can refactor these tests if it turns out that current approach significantly impacts the maintainability of these test.

src/Microsoft.ServiceFabric.Actors/Runtime/ActorManager.cs
ilijagrbic: Thanks for explaining this. I was biased when thinking about this since the original location for old `DiagnosticManager` was in the `ActorManager`. Anyways, I think this should be tackled (together with the `IClock` issue) and I will do it in a follow up PR.

olegsych (commented):
:shipit:

olegsych (approved):

PR #491, opened 2025-10-15
Remove Actor state migration APIs

### LINKED ISSUE #448: Remove Microsoft.ServiceFabric.Actors.KVSToRCMigration
The KVS to RC migration projects were created years ago based on the direction from the Service Fabric Data team, who didn't want to support automatic migration. Since then, the strategy has changed, the migration effort was abandoned and these projects haven't been maintained.

- Remove the following projects
  Microsoft.ServiceFabric.Actors.KVSToRCMigration.csproj
  Microsoft.ServiceFabric.Actors.StateMigration.Tests.csproj
- Remove any references to these assemblies from the remaining project files and source code, including the InternalVisibleToAttributes in the Friends.cs files.
- Remove the state migration APIs from the Microsoft.ServiceFabric.Actors project.

olegsych:
Actor state migration is very complex, requires elaborate implementation and orchestration in user applications and doesn't solve the problem for non-Actor stateful services. Because of this the data migration from KVS to RC will be handled by the Data stack transparently for the customer applications. 

#490 deprecates all migration-related APIs. This PR completes their removal.

Most of the changes are a straightforward removal of the migration APIs and their implementation with the exception of the `FabricTransportActorServiceRemotingListener`, where migration-related parameters were removed from the public constructors. These changes need to be reviewed together with #490 where constructors with old signatures were deprecated and new constructors with new signatures were introduced. Another non-trivial change is in the `ActorService` where some internal constructors were removed because their signatures became identical to other constructors after removal of migration-related parameters.

Bump internal package version to 12/9.12.0 for integration with WindowsFabric.

This resolves #448


src/Microsoft.ServiceFabric.Actors/FabricActorExceptionKnownTypes.cs
```diff
@@ -83,30 +82,6 @@ internal class FabricActorExceptionKnownTypes
                         InnerExFunc = ex => GetInnerExceptions(ex),
                     }
                 },
-                {
-                    "Microsoft.ServiceFabric.Actors.Migration.Exceptions.ActorCallsDisallowedException", new ConvertorFuncs()
-                    {
-                        ToServiceExFunc = ex => ToServiceException(ex),
-                        FromServiceExFunc = (svcEx, innerEx) => FromServiceException<ActorCallsDisallowedException>(svcEx, innerEx),
```
gobradovic, line 90: If caller service is on the new version and don't know these exceptions and callee service is on the old version and throws some of these, could we have a problem?

gobradovic (approved):

src/Microsoft.ServiceFabric.Actors/FabricActorExceptionKnownTypes.cs
olegsych: They'd just get a ServiceException, correct? I think that should be fine.

PR #492, opened 2025-10-16
Remove Services.Tests.csproj from list of test projects in netframework.targets

olegsych:
This project doesn't target net472 and fails to build in the ADO pipeline due to a missing Main entry point. It is not clear why this error doesn't happen during GitHub builds, however since we're removing the old build infrastructure, it's probably not worth the effort to get to the root cause of this.

### DIFF
--- a/properties/service_fabric_managed_netframework.targets
+++ b/properties/service_fabric_managed_netframework.targets
@@ -13,7 +13,6 @@
 
     <Projects_Test_netframework Include="$(RepoRoot)test\Microsoft.ServiceFabric.Actors.Tests\Microsoft.ServiceFabric.Actors.Tests.csproj;
                                    $(RepoRoot)test\Microsoft.ServiceFabric.Services.Remoting.Tests\Microsoft.ServiceFabric.Services.Remoting.Tests.csproj;
-                                   $(RepoRoot)test\Microsoft.ServiceFabric.Services.Tests\Microsoft.ServiceFabric.Services.Tests.csproj;
                                    $(RepoRoot)test\Microsoft.ServiceFabric.Diagnostics.Tests\Microsoft.ServiceFabric.Diagnostics.Tests.csproj;
                                    $(RepoRoot)test\Microsoft.ServiceFabric.Actors.IntegrationTests\Microsoft.ServiceFabric.Actors.IntegrationTests.csproj" />
   </ItemGroup>


properties/service_fabric_managed_netframework.targets
```diff
@@ -13,7 +13,6 @@
 
     <Projects_Test_netframework Include="$(RepoRoot)test\Microsoft.ServiceFabric.Actors.Tests\Microsoft.ServiceFabric.Actors.Tests.csproj;
                                    $(RepoRoot)test\Microsoft.ServiceFabric.Services.Remoting.Tests\Microsoft.ServiceFabric.Services.Remoting.Tests.csproj;
-                                   $(RepoRoot)test\Microsoft.ServiceFabric.Services.Tests\Microsoft.ServiceFabric.Services.Tests.csproj;
```
ilijagrbic, line 16: I think our current build configuration is a bit missleading. This `service_fabric_managed_netframework.targets` is used to define and build both netfx and netcore projects, even though the name would suggest otherwise. On the other hand, `service_fabric_managed_netcore.targets` only copies the already build binaries to their respective locations in the drop directory.

ilijagrbic: I managed to reproduce the issue locally (the missing Main method error CS5001). It's a really weird issue; it goes away when I change the `csproj` files `TargetFramework` by adding `net472`:
`<TargetFrameworks>net9.0;net472</TargetFrameworks>`

I'm not sure why would this be the case, but on the other hand all of our test projects are being built with both frameworks (except for WCF which is `net472` only). Which I think might be a better way to solve this until we figure out the issue and refactor the target files to be less misleading.

One reason why I'm concerned about just removing this project from targets is that I'm not sure if it's being built any longer. I can't find a log in the build pipeline that `Services.Tests` is built. And when I run build locally it is also not built. One thing I don't understand is how are tests being run for `Services.Tests` then in this PRs pipeline? Perhaps the project is restored but I don't know how?

ilijagrbic (changes requested):
Hey @olegsych I would like to discuss this today before approving. I don't think we should remove this project from targets, I wrote a more detailed comment regarding all of this.

ilijagrbic (approved):

properties/service_fabric_managed_netframework.targets
olegsych: Thanks @ilijagrbic I also managed to reproduce this problem locally, the trick was to use the full command line from our ADO pipeline `dotnet msbuild buildAll.proj /property:Configuration=Release /t:BuildCode` and particularly the `BuildCode` target. There must be some build .props and .targets in the xunit.v3 package to affect this behavior and I suspect package restore is involved. I would like to keep enabling Service.Tests on net472 out of scope for now I'm leaving it as is as we've discussed.

PR #496, opened 2025-10-25
Make build infrastructure support dotnet pack

olegsych:
- Define common package properties in Directory.Build.props and Directory.Build.targets. Because Directory.Build.props is imported very early, properties based on other properties may need to be defined in Directory.Build.targets instead.
- Disable packing in test projects as well as in FabActUtil.csproj which is included in the Microsoft.ServiceFabric.Actors package.
- Build .snupkg symbols packages in addition to regular .nupkg packages.
- Make Microsoft.ServiceFabric.Constants project references use PrivateAssets=All. This is a compile-only dependency and shouldn't generate package dependencies in the generated packages.
- Make projects use common build output directories.
  - Binaries are placed in the bin directory. We will sign files in this directory before generating packages.
  - Test results are placed in the out/tests directory.
  - Packages are placed in the out/packages directory.
- Add transitive package references to Directory.Packages.props to prevent assembly version conflicts producing double writes at build time and test failures at run time.
- Make build.yml build, test and pack code.sln
  - Generate .binlog files during build
  - Organize build artifacts in sub-folders for logs, packages and test results.


olegsych:
No, this would need a lot more work.

PR #501, opened 2025-11-13
Refactor ActorService to be composition root for IClock and IDiagnostics

ilijagrbic:
- Move creation of `IDiagnostics` and `IClock` to `ActorService`
- Use Factory pattern to instantiate `IDiagnostics` in the `DiagnosticsFactory`
- `DiagnosticsFactory` creates and disposes of `PerformanceCounterProviderV2` needed for `AggregatedDiagnosticsEvents`.
- Since some of the classes have public APIs, we use `IDiagnostics` and `IClock`  from `ActorService` by reading it from `ActorService` parameter - no injection (`ActorServiceRemotingDispatcher` and `ActorBase`)
- For classes where it was possible we inject `IDiagnostics` and `IClock`. (`ActorStateManager` and `ActorManager`)


olegsych (approved):
Feels like a step in the right direction. Looks like we're reducing the depth of Law of Demeter violations. Perhaps at some point the DiagnosticsFactory could become a singleton reused rather than created by the ActorService to shift more complexity to the new class.

PR #504, opened 2025-11-20
Implement MetricsDiagnostics in Actors

ilijagrbic:
- Implement `MetricsDiagnostics`, emit metrics analogous to PerfCounters already emitted
- Integrate metrics into `DiagnosticsFactory` and `ActorService`
- All of test depending on `ActorService` or `DiagnosticsFactory` now inherit from `MocketMetricsTest` class in order to mock Interop calls to FabricTelemetry.dll


src/Microsoft.ServiceFabric.Actors/Diagnostics/DiagnosticsFactory.cs
```diff
@@ -27,13 +31,16 @@ public DiagnosticsFactory(ServiceContext serviceContext, ActorTypeInformation ty
 
             performanceCounterProvider = new PerformanceCounterProviderV2(serviceContext.PartitionId, typeInformation);
             performanceCounterProvider.InitializeActorMethodInfo(this.friendlyNameBuilder);
+            timeSpanMeterProvider = new TimeSpanMeterProvider(serviceContext);
+            longMeterProvider = new Int64MeterProvider(serviceContext);
```
ilijagrbic: I was thinking if instantiation of `MeterProviders` should be moved up the hierarchy to `ActorService` composition root and then injected into the factory. On the other hand, I like that all Diagnostics related dependencies are created in the Factory and that `ActorService` itself only cares about `IDiagnostics` and the factory itself. Also, `MeterProviders` seem like something that won't ever be needed outside the factory context.

olegsych: Yeah, the complexity balance seems to be better with creation logic encapsulated in the DiagnosticsFactory. Another option to consider is singleton meter provider factories, but the trade-offs aren't immediately obvious to me. I think this looks good for now.

src/Microsoft.ServiceFabric.Actors/Diagnostics/DiagnosticsFactory.cs
```diff
@@ -27,13 +31,16 @@ public DiagnosticsFactory(ServiceContext serviceContext, ActorTypeInformation ty
...
         }
 
         public virtual IDiagnostics CreateDiagnostics(IClock clock)
         {
             var performanceCounterDiagnostics = new PerformanceCounterDiagnostics(performanceCounterProvider, clock);
             var eventSourceDiagnostics = new EventSourceDiagnostics(ActorFrameworkEventSource.Writer, clock, serviceContext, friendlyNameBuilder, typeInformation);
-            var registeredDiagnostics = new List<IDiagnostics> { performanceCounterDiagnostics, eventSourceDiagnostics };
+            var metricDiagnostics = new MetricDiagnostics(longMeterProvider, timeSpanMeterProvider, clock);
+            var registeredDiagnostics = new List<IDiagnostics> { performanceCounterDiagnostics, eventSourceDiagnostics, metricDiagnostics };
```
olegsych: We won't be creating many instances of these, so the perf overhead of lists is not an issue. But since the AggregatedDiagnostics takes an IEnumerable<T>, I think a `new IDiagnostics[]` would expresses the intent better, with a little less verbiage. Just a nit though.

src/Microsoft.ServiceFabric.Actors/Diagnostics/MetricDiagnostics.cs
```diff
@@ -0,0 +1,125 @@
...
+// ------------------------------------------------------------
+
+using System;
+using System.Fabric;
+using Microsoft.ServiceFabric.Diagnostics;
+using Microsoft.ServiceFabric.Diagnostics.Metrics;
+
+namespace Microsoft.ServiceFabric.Actors.Diagnostics
+{
+    internal class MetricDiagnostics : IDiagnostics
```
olegsych: Can we seal this class and remove the redundant internal?

src/Microsoft.ServiceFabric.Actors/Diagnostics/MetricDiagnostics.cs
```diff
@@ -0,0 +1,125 @@
...
+        readonly IMeter<TimeSpan> loadStateDuration;
+        readonly IMeter<TimeSpan> saveStateDuration;
+
+        public MetricDiagnostics(IMeterProvider<long> longMeterProvider, IMeterProvider<TimeSpan> timeSpanProvider, IClock clock)
+        {
+            _ = longMeterProvider ?? throw new ArgumentNullException(nameof(longMeterProvider));
+            _ = timeSpanProvider ?? throw new ArgumentNullException(nameof(timeSpanProvider));
+            this.clock = clock ?? throw new ArgumentNullException(nameof(clock));
+
+            this.actorLockContention = longMeterProvider.CreateMeter("Actor", "ActorLockContention");
```
olegsych: Let's create constants for the namespace and try using nameof instead of hardcoding the metric and dimension names.

olegsych: I think ActorLockContention doesn't explain itself and would need documentation. Why don't we name this metric with nameof(PendingActorMethodDiagnosticData.PendingActorMethodCalls)?

src/Microsoft.ServiceFabric.Actors/Diagnostics/MetricDiagnostics.cs
```diff
@@ -0,0 +1,125 @@
...
+            this.acquireLockDuration = timeSpanProvider.CreateMeter("Actor", "AcquireLockDuration");
+            this.releaseLockDuration = timeSpanProvider.CreateMeter("Actor", "ReleaseLockDuration");
+            this.methodExceptionCount = longMeterProvider.CreateMeter("Actor", "MethodExceptionCount", "MethodId");
```
olegsych: We shouldn't emit a separate metric for exception count. Instead, the MethodExecutionDuration metric should have a dimension for exceptions. I would call it Exception and provide dimension value "None" when the method call is completed successfully and when an exception was thrown, provide type name of the exception as the dimension value. 

With this metric, we can determine not just the count of the exceptions, but their distribution by type and correlation with the method call durations.

test/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/DiagnosticsFactoryTest.cs
```diff
@@ -128,14 +137,23 @@ public void CreatesPerformanceCounterEventsFirst()
...
 
                 Assert.IsAssignableFrom<EventSourceDiagnostics>(eventSourceDiagnostics);
                 Assert.Equal(ActorFrameworkEventSource.Writer, eventSourceDiagnostics.Field<ActorFrameworkEventSource>().Value);
                 Assert.Equal(serviceContext, eventSourceDiagnostics.Field<ServiceContext>().Value);
                 Assert.Equal(clock, eventSourceDiagnostics.Field<IClock>().Value);
                 Assert.Equal(typeInformation.ImplementationType.ToString(), eventSourceDiagnostics.Field<string>().Value);
             }
+
+            [Fact]
+            public void CreatesPerformanceCounterEventsThird()
```
olegsych: Is the test name accurate? what's the significance of the third item?

test/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/DiagnosticsFactoryTest.cs
```diff
@@ -72,14 +74,21 @@ public void InstantiatesPerfCounterProvider()
...
+            [Fact]
+            public void InstantiatesMeterProviders()
+            {
+                Assert.NotNull(sut.Field<IMeterProvider<TimeSpan>>().Value);
+                Assert.NotNull(sut.Field<IMeterProvider<long>>().Value);
+            }
         }
 
-        public class Dispose : DiagnosticsFactoryTest
+        public class LifeCycle : DiagnosticsFactoryTest
```
olegsych: Why change? Doesn't this test class test the Dispose method?

test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorIntegrationTests/RegisterReminderAsync.cs
```diff
@@ -1,184 +1,185 @@
...
 using System.Threading.Tasks;
+using Fuzzy;
+using Microsoft.ServiceFabric.Actors.Runtime;
+using Microsoft.ServiceFabric.TestFramework;
 using Xunit;
 
 namespace Microsoft.ServiceFabric.Actors.ActorIntegrationTests;
 
-    public class RegisterReminderAsync
+public class RegisterReminderAsync : MockedMetricsTest
```
olegsych, line 12: Wow, what happened here? Are these white space changes?

test/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/MetricDiagnosticsTest.cs
```diff
@@ -0,0 +1,233 @@
...
+            [Fact]
+            public void ThrowsOnNullTimeSpanMeterProvider()
+            {
+                var exception = Assert.Throws<ArgumentNullException>(() => new MetricDiagnostics(mockLongMeterProvider, null, clock));
+                Assert.Equal("timeSpanProvider", exception.ParamName);
+            }
+
+            [Fact]
+            public void WithParametersCreatesMeters()
+            {
```
olegsych, line 86: Since we're testing that expected meters were created and used to emit expected metrics, I would typically write this test a bit differently. This is pseudo-code, I still can't write Moq code on the fly.

```C#
readonly IMeter<TimeSpan> actorLockDuration = Mock.Of<IMeter<TimeSpan>>();

public MetricDiagnosticsTest()
{
    timeSpanMeterProvider.Setup(_ => _.Create("Actor", "AcquireLockDuration")).Returns(actorLockDuration);
    ...
    sut = new MetricDiagnostics(mockLongMeterProvider, mockTimeSpanMeterProvider, clock)
}
```

This setup code also serves as a test for meter creation, so we don't have to test it explicitly. As long as we test that actorLockDuration.Record is called as expected, we know that it was created correctly.

src/Microsoft.ServiceFabric.Actors/Diagnostics/MetricDiagnostics.cs
```diff
@@ -0,0 +1,125 @@
...
+            this.methodExecutionDuration = timeSpanProvider.CreateMeter("Actor", "MethodExecutionDuration", "MethodId");
```
olegsych: Are MethodIds stable between processes, nodes, app versions, etc? At the minimum we should be able to convert them back to the actor names long time after the app was built and deployed. I think it would be best to use method name, possibly method signature as the dimension value, like the EventSourceDiagnostics does.

src/Microsoft.ServiceFabric.Actors/Diagnostics/MetricDiagnostics.cs
```diff
@@ -0,0 +1,125 @@
...
+            this.onActivateAsyncDuration = timeSpanProvider.CreateMeter("Actor", "OnActivateAsyncDuration");
```
olegsych: I think we also need an actor name dimension.

olegsych (approved):
@ilijagrbic I think this is good enough as the first implementation; it looks like we're achieving parity with the perf counters. However, the metrics still need to be improved to be usable for monitoring and troubleshooting actors.

src/Microsoft.ServiceFabric.Actors/Diagnostics/MetricDiagnostics.cs
ilijagrbic: I'm creating mocks of this class, so sealing it would require some test changes which I don't think would be worthwhile right now. I will remove the redundant internal modifier though.

test/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/DiagnosticsFactoryTest.cs
ilijagrbic: Yeah, the test name is off, thanks for noticing this. The test itself is accurate as it verifies that we are creating MetricsDiagnostics. Also, I prefer to create these events in a deterministic order, so that's another reason why I'm specifically verifying each item separately, instead of just verifying that collection has items of all needed types (which could probably be done in one shorter test).

test/Microsoft.ServiceFabric.Actors.IntegrationTests/ActorIntegrationTests/RegisterReminderAsync.cs
ilijagrbic: Yup, all of the code was moved by one "tab" to the right, and when I saved the file autoformat was applied. Normally I don't like to introduce huge changes because of formatting in old files, but his just seemed like an unintended mistake when the file was created (and it was created recently). :)

src/Microsoft.ServiceFabric.Actors/Diagnostics/MetricDiagnostics.cs
ilijagrbic: Thanks, this is a good suggestion. I kept the naming of the PerfCounters, but it should definitely be improved.

ilijagrbic: I decided to not use `nameof`, since all other names are hardcoded. I understand this is not ideal, but we can iterate on this and keep it for now as is.

test/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/DiagnosticsFactoryTest.cs
ilijagrbic: Great and interesting question. All of these tests are implicitly implementing `IDisposable`, since they inherit from `MockedMetricsTest`. Anyways these classes already have a member called `Dispose` as a method, and adding a class with the same name introduced warnings which are reported as errors during the build. I decided to rename the class to quickly resolve this issue.
Here is the error I encountered:
```
D:\a\service-fabric-services-and-actors-dotnet\service-fabric-services-and-actors-dotnet\test\Microsoft.ServiceFabric.Actors.Tests\Diagnostics\DiagnosticsFactoryTest.cs(86,22): error CS0108: 'DiagnosticsFactoryTest.Dispose' hides inherited member 'MockedMetricsTest.Dispose()'. Use the new keyword if hiding was intended. [D:\a\service-fabric-services-and-actors-dotnet\service-fabric-services-and-actors-dotnet\test\Microsoft.ServiceFabric.Actors.Tests\Microsoft.ServiceFabric.Actors.Tests.csproj::TargetFramework=net9.0]
```

test/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/MetricDiagnosticsTest.cs
ilijagrbic: Thanks, this is a great suggestion. I need to try to incorporate this kind of thinking about tests more often. I will adjust the test code as you suggested.

src/Microsoft.ServiceFabric.Actors/Diagnostics/MetricDiagnostics.cs
ilijagrbic: Thank for noticing this as well, I wanted to improve this, but it slipped by me later on. I will use the same logic we use for `EventSourceDiagnostics`.

src/Microsoft.ServiceFabric.Actors/Diagnostics/MetricDiagnostics.cs
ilijagrbic: This is something I was considering and in my previous experience has been a good practice. I was just not sure whether this would impact the metrics aggregations and if cardinality of the Exception dimensions could be a problem, so I wanted to discuss this and follow up in iterations. Anyways, I will refactor the code according to your suggestions, as I think this is way better as well. :)

src/Microsoft.ServiceFabric.Actors/Diagnostics/MetricDiagnostics.cs
ilijagrbic: I thought that this was not needed since we emit `ServicName` and `ServiceTypeName` as system dimensions. I'm not sure, but I thought `ActorName` can be determined from all of the system dimensions.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodInfoUtil.cs
```diff
@@ -0,0 +1,40 @@
...
+// ------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using Microsoft.ServiceFabric.Actors.Runtime;
+using Microsoft.ServiceFabric.Services.Remoting.Description;
+
+namespace Microsoft.ServiceFabric.Actors.Diagnostics
+{
+    internal class ActorMethodInfoUtil
```
olegsych: Could you make this class static and remove the default internal visibility? We only need it for functions.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodInfoUtil.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        internal static long GetInterfaceMethodKey(uint interfaceId, uint methodId)
+        {
+            var key = (ulong)methodId;
+            key = key | (ulong)interfaceId << 32;
+            return (long)key;
+        }
+
+        internal static Dictionary<long, ActorMethodInfo> BuildActorMethodInfo(ActorMethodFriendlyNameBuilder nameBuilder, ActorTypeInformation typeInfo)
```
olegsych: Can we return an IReadOnlyDictionary?

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnostics.cs
```diff
@@ -18,6 +17,7 @@ sealed class EventSourceDiagnostics : IDiagnostics
         readonly ServiceContext serviceContext;
         readonly IClock clock;
         readonly string actorType;
+
```
olegsych, line 20: Intentional?

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodInfoUtil.cs
olegsych: Why did you choose to make this a static utility method rather than an ActorTypeInformation method?

src/Microsoft.ServiceFabric.Actors/Diagnostics/MetricDiagnostics.cs
```diff
@@ -0,0 +1,130 @@
...
+        readonly IMeter<long> pendingMethodCalls;
+        readonly IMeter<TimeSpan> acquireLockDuration;
+        readonly IMeter<TimeSpan> releaseLockDuration;
+        readonly IMeter3D<TimeSpan> methodExecutionDuration;
+        readonly IMeter<TimeSpan> onActivateAsyncDuration;
+        readonly IMeter<TimeSpan> requestProcessingDuration;
+        readonly IMeter<TimeSpan> loadStateDuration;
+        readonly IMeter<TimeSpan> saveStateDuration;
+
+        readonly Dictionary<long, ActorMethodInfo> actorMethodInfo;
```
olegsych: Yup, this should really be an IReadOnlyDictionary.

test/Microsoft.ServiceFabric.Actors.Tests/Diagnostics/DiagnosticsFactoryTest.cs
olegsych: Ah, makes sense. I usually solve this problem by making the class name DisposeTest. Although the term Test appears in the full class name multiple times, it still maintains direct mapping of tests with their "subjects".

olegsych (approved):

src/Microsoft.ServiceFabric.Actors/Diagnostics/MetricDiagnostics.cs
olegsych: Would it work when a service registers multiple actor types or when a client creates multiple proxy types?
https://dev.azure.com/msazure/One/_git/WindowsFabric?path=/src/test/current/source/sdk/Actors/ReentrancyTests/Program.cs&version=GBdevelop&_a=contents

src/Microsoft.ServiceFabric.Actors/Diagnostics/EventSourceDiagnostics.cs
ilijagrbic: Yeah, I like to separate the fields in groups, roughly by their purpose. Here the first 4 fields are in my mind "core" for this class, are needed in basically every method. On the other hand, this dictionary is specific for only one of the methods, and a bit less significant than other fields, so that's why I separated them. Let me know what you think about this and if you think it is not needed. I guess in the end it comes down to preference regarding these things.

src/Microsoft.ServiceFabric.Actors/Diagnostics/ActorMethodInfoUtil.cs
ilijagrbic: Good question, my reasoning was that `ActorTypeInformation` represents a concept we need throughout the system and is references by a lot of other components. This method was not intended (at least for now) to be used outside of specifically Actor diagnostics. And in case it turns out we need a similar method in other places, I would introduce a proper abstraction for it, perhaps as a member of `ActorTypeInformation` or perhaps in some other way. But for now, it made sense to me to keep it contained within this namespace as a simple utility method. 

Another way was to include this method in `ActorMethodFriendlyNameBuilder` and refactor that class a bit. But it had a similar issue as `ActorTypeInformation`, because that class was also used throughout the system, although to a lesser extent and I wanted to keep the changes localized as much as possible for this iteration.

src/Microsoft.ServiceFabric.Actors/Diagnostics/MetricDiagnostics.cs
ilijagrbic: Yup, thanks for suggesting this, I will fix this.

src/Microsoft.ServiceFabric.Actors/Diagnostics/MetricDiagnostics.cs
ilijagrbic: Regarding the multiple actor types, isn't each Actor supposed to be a standalone StatefulService? As far as I can see in the ApplicationManifest and ServiceManifest, there are two StatefulServices defined in this example. And to my knowledge it is not possible to register multiple actor types within the same service, the way I see it - there is a 1-1 mapping between ActorType-ActorService-StatefulService. Correct me if I'm wrong on this, perhaps I'm misunderstanding something.

olegsych (commented):
:shipit:

olegsych (approved):

PR #505, opened 2025-11-27
Adding support for .NET 10 SDK and runtime

plave0:
- Set minimum required version of SDK to 10.0.0. With the implemented configuration, the latest version of SDK 10.x which is available will be used.
- Build unit test for .NET 10. This confirms that prog model libraries built for .NET 8 will work for .NET 10 as well.


global.json
```diff
@@ -0,0 +1,6 @@
+{
+    "sdk": {
+        "version": "10.0.0",
+        "rollForward": "latestMajor"
```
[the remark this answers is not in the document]
ilijagrbic: @plave0 do we need to address this? I think we probably do.

ilijagrbic (approved):
Hey @plave0 , I would just double check the Copilots comment, other than that - :shipit:

ilijagrbic (approved):

.github/workflows/build.yml
```diff
@@ -40,7 +40,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - run: dotnet build -bl:"$Logs/build.binLog"
-      - run: dotnet test --no-build --logger trx --framework net9.0 -bl:"$Logs/test.binLog"
+      - run: dotnet test --no-build --logger trx --framework net10.0 -bl:"$Logs/test.binLog"
```
olegsych, line 43: @plave0 Can we dotnet test for multiple frameworks? I wonder if it’s possible or a good idea in a single step?

plave0: Good catch
I will include dotnet test for .NET 9 on Linux build in my text PR

PR #506, opened 2025-12-04
Add System.Fabric.Common.Test to assemblyinfo

TypoInUsrename:
This PR adds InternalsVisibleTo attributes to allow the `System.Fabric.Common.Test` assembly to access internal members of the `Microsoft.ServiceFabric.Diagnostics` assembly used in tests.


ilijagrbic (changes requested):
@TypoInUsrename Lets also update the package `Revision` in this PR, so that we can build new packages and integrate them into WindowsFabric.

ilijagrbic (approved):

PR #515, opened 2025-12-24
Move projects from service-fabric-aspnetcore

olegsych:
We will ship `Microsoft.ServiceFabric.AspNetCore.*` packages from this repo starting with Service Fabric version 12.

Changes made to integrate this code:
- Make `.csproj` files conform to current conventions in this repo.
- Make `.cs` files place usings above namespace, remove unnecessary usings.
- Add new assemblies to `SkipStrongName.ps1` and resort the list.
- Expand the `.AspNetCore.Tests.csproj` to `net472`, `net8.0` and `net10.0` in addition to `net9.0`.
- Temporarily allow new `net10.0` warning `ASPDEPR008` in the `.AspNetCore.Tests.csproj`.
- Update package references to
  - Resolve deprecation errors.
  - Prevent double-writes breaking the `net472` tests with assembly binding errors.
- Make `service_fabric_managed_common.props` preserve the `<RootNamespace>` specified in the `AspNetCore` projects. Without it the resources are embedded with incorrect name.

Also:
- Create `init.cmd` for configuring pre-requisites.
- Upgrade `xunit` to latest version, just in case it has anything to do with test failures.
- Move usings above namespaces to better match our current coding style.
- Bump package revision number to 19 for WindowsFabric integration
- Add AspNetCore packages to the list in `README.md`


src/AspNetCore.HttpSys/Microsoft.ServiceFabric.AspNetCore.HttpSys.csproj
```diff
@@ -0,0 +1,32 @@
...
+  <Import Project="..\..\properties\service_fabric_managed_common.props" />
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ServiceFabric" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' != 'net462'">
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
+    <PackageReference Include="Microsoft.AspNetCore.Server.HttpSys"/>
+  </ItemGroup>
```
ilijagrbic, line 15: This is not something this PR introduced, but it is a good opportunity to remedy it. I think we should consider moving the TargetFramework target versions into the service_fabric_managed_common.props, instead of having them hardcoded in both `<TargetFrameworks>` and every `$(TargetFramework)` check later on. It should be a simple change, and it would lower the probability of accidental mistakes. Feel free to merge without this, I don't think it is critical.

ilijagrbic (approved):
I haven't reviewed the .cs files, since I assume the have no major changes. Everything else looks good. :shipit:

PR #517, opened 2026-01-08
Move projects from service-fabric-client-dotnet

olegsych:
We will ship the `Microsoft.ServiceFabric.Client.Http` package and the `Microsoft.ServiceFabric.PowerShell.Http` module from this repo starting with Service Fabric version 12.

An in-depth review of the code was performed as part of this integration in order to gain deep understanding of the code-base and its suitability for further evolution of the Service Fabric .NET SDK. 

The `Microsoft.ServiceFabric.Client.dll` and `Microsoft.ServiceFabric.Common.dll` were never shipped separately before. They were bundled in the `.Http` NuGet package and PowerShell module. While their overall design is a good idea, unfortunately the code wasn't clean enough to be shipped in a stable package version, particularly the code generated from SF-Swagger. On the other hand, we need these assemblies to start pulling apart the monolithic `System.Fabric.dll` and shiping the SDK independently of the Runtime. To avoid locking in another huge set of public APIs, the `.Client` and `.Common` DLLs have been eliminated and their code moved to the `Microsoft.ServiceFabric.Client.Http.dll`. We will use TypeForwardedToAttribute in both `System.Fabric.dll` and `Microsoft.ServiceFabric.Client.Http.dll` when moving types to the clean `Microsoft.ServiceFabric.Common.dll` and `Microsoft.ServiceFabric.Client.dll` we will create shortly.

As part of the review, a safe cleanup of new projects was performed using refactoring tools. The generated code wasn't cleaned up.
  - Move misplaced using directives above namespace
  - Remove unused code and redundant comments
  - Remove unnecessary parenthesis and braces
  - Rename/move files to match their name and namespace
  - Replace var with explicit type where it doesn't already appear in the initialization statement
  - Simplify default value declaration
  - Remove redundant visibility specifiers and comments
  - For internal types, where possible
    - Reduce visibility of type members
    - Implement interfaces explicitly
    - Make sealed or abstract


code.slnx
```diff
@@ -12,23 +12,27 @@
...
     <Project Path="src/Services.Remoting/Microsoft.ServiceFabric.Services.Remoting.csproj" />
     <Project Path="src/Services.Wcf/Microsoft.ServiceFabric.Services.Wcf.csproj" />
     <Project Path="src/Services/Microsoft.ServiceFabric.Services.csproj" />
   </Folder>
   <Folder Name="/test/">
-    <Project Path="test/AspNetCore/Microsoft.ServiceFabric.AspNetCore.Tests.csproj" />
     <Project Path="test/Actors.Integration/Microsoft.ServiceFabric.Actors.IntegrationTests.csproj" />
-    <Project Path="test/Actors/Microsoft.ServiceFabric.Actors.Tests.csproj" />
     <Project Path="test/Actors.Wcf/Microsoft.ServiceFabric.Actors.Wcf.Tests.csproj" />
+    <Project Path="test/Actors/Microsoft.ServiceFabric.Actors.Tests.csproj" />
```
olegsych, line 28: Project list resorted in VSCode

refs/Microsoft.ServiceFabric.YamlUtils.dll
gobradovic, line 1: Can we avoid having binaries in repo?

olegsych: I hope so. I suspect we only need these yaml dependencies for the Mesh APIs that we never shipped and should remove. Let me take a closer look at this.

olegsych: Well, the good news is that my guess that we only need Yaml DLLs for the Mesh commandlets, which we don't support. The bad news is that properly removing Mesh will probably take a few days to track down all unused APIs across three different repos. I'll push a commit removing Mesh to this PR branch, but it'll be big - more than two hundred files. It would somewhat reduce the number of files to review in this PR, but not if you're reviewing each commit separately. Let me know if you would prefer to separate it.

olegsych: Actually, I didn't realize we've accumulated a lot of Swagger changes that haven't been re-generated, so I think it'll be better to do it in a separate PR. Otherwise, I'll have to edit large, generated files by hand

gobradovic:
You have duplicate "client" in PR title: "service-fabric-client-client-dotnet"

gobradovic (approved):

PR #519, opened 2026-01-21
Replace usage of MockedMetricsTest with AssemblyFixture

ilijagrbic:
- Remove usage of MockedMetricsTest class, since it seems to cause issues, the cause of this could be possible parallel execution of these tests, because of multiple collection defined in it.
- Instead of relying on MockedMetricsTest, define an assembly level fixture to override the static fields that call the native interop


ilijagrbic:
@copilot Can you run the Microsoft.ServiceFabric.Actors.IntegrationTests.dll tests multiple, like 20, times and report the results?

ilijagrbic:
@copilot  Can you run the Microsoft.ServiceFabric.Actors.IntegrationTests.dll tests multiple, like 20, times and report the results?

test/TestFramework/TelemetryApiFixture.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+using System.Threading.Tasks;
+using Inspector;
+using Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation;
+using Moq;
+using Xunit;
+
+namespace Microsoft.ServiceFabric.TestFramework
+{
+    public class TelemetryApiFixture : IAsyncLifetime
```
olegsych: I would try to make the class name more specific so that we don't have to add a comment every time we use it.

olegsych: It looks like we don't need async initialization or cleanup. Can we simplify this from `IAsyncLifetime` to `IDisposable`?
https://xunit.net/docs/shared-context#assembly-fixture

test/TestFramework/FabricTelemetryDllFixture.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Threading.Tasks;
+using Inspector;
+using Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation;
+using Moq;
+using Xunit;
+
+namespace Microsoft.ServiceFabric.TestFramework
+{
+    public class TelemetryApiFixture : IAsyncLifetime
+    {
```
olegsych, line 14: The new fixture looks very similar to the `MockedMetricsTest`. What if we reuse the existing class instead of creating a new one?

olegsych (approved):
Looks good. Some thoughts I had while reviewing are in the comments.

test/TestFramework/TelemetryApiFixture.cs
ilijagrbic: I will rename it to "FabricTelemetryDllFixture" and remove the comments. Main reason why I added the comments is in case the reader is unfamiliar with the Assembly Fixtures concept, but I can see that it might not be necessary.

ilijagrbic: Good point, I will change this.

test/TestFramework/FabricTelemetryDllFixture.cs
ilijagrbic: Good point, I will unify them and use the new class instead.

ilijagrbic:
Thanks for the review @olegsych , I've addressed the comments you've left. Feel free to reapprove if you think this is good to go, I enabled auto-merge. :)

olegsych (approved):
:shipit:

PR #522, opened 2026-01-22
Fix PowerShell.Http module publishing

olegsych:
- Regenerate Client/Powershell.Http code from latest SF-Swagger definitions.
- Ensure that all Powershell.Http dependencies are published for the netstandard version of the module
- Bump package version number to 12.0.0-beta.21

### DIFF
--- a/src/Client.Http/Common/Generated/ServiceUpdateDescription.cs
+++ b/src/Client.Http/Common/Generated/ServiceUpdateDescription.cs
@@ -43,8 +43,8 @@ namespace Microsoft.ServiceFabric.Common
         /// - InstanceRestartWaitDuration - Indicates the InstanceCloseDelayDuration property is set. The value is 32768.
         /// - DropSourceReplicaOnMove - Indicates the DropSourceReplicaOnMove property is set. The value is 65536.
         /// - ServiceDnsName - Indicates the ServiceDnsName property is set. The value is 131072.
-        /// - TagsForPlacement - Indicates the TagsForPlacement property is set. The value is 1048576.
-        /// - TagsForRunning - Indicates the TagsForRunning property is set. The value is 2097152.
+        /// - ServiceTags TagsRequiredToPlace - Indicates the TagsRequiredToPlace property is set. The value is 1048576.
+        /// - ServiceTags TagsRequiredToRun - Indicates the TagsRequiredToRun property is set. The value is 2097152.
         /// </param>
         /// <param name="placementConstraints">The placement constraints as a string. Placement constraints are boolean
         /// expressions on node properties and allow for restricting a service to particular nodes based on the service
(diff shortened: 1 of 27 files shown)


src/Client.Http/Common/Generated/ServiceUpdateDescription.cs
```diff
@@ -43,8 +43,8 @@ public abstract partial class ServiceUpdateDescription
         /// - InstanceRestartWaitDuration - Indicates the InstanceCloseDelayDuration property is set. The value is 32768.
         /// - DropSourceReplicaOnMove - Indicates the DropSourceReplicaOnMove property is set. The value is 65536.
         /// - ServiceDnsName - Indicates the ServiceDnsName property is set. The value is 131072.
-        /// - TagsForPlacement - Indicates the TagsForPlacement property is set. The value is 1048576.
-        /// - TagsForRunning - Indicates the TagsForRunning property is set. The value is 2097152.
+        /// - ServiceTags TagsRequiredToPlace - Indicates the TagsRequiredToPlace property is set. The value is 1048576.
+        /// - ServiceTags TagsRequiredToRun - Indicates the TagsRequiredToRun property is set. The value is 2097152.
```
gobradovic, line 47: Shouldn't here be RepartitionDescription?

src/Client.Http/Common/Generated/ServiceUpdateDescription.cs
```diff
@@ -116,8 +116,8 @@ protected ServiceUpdateDescription(
...
```
gobradovic, line 120: Shouldn't here be RepartitionDescription?

src/Client.Http/Common/Generated/StatefulServiceUpdateDescription.cs
```diff
@@ -42,8 +42,8 @@ public partial class StatefulServiceUpdateDescription : ServiceUpdateDescription
         /// - InstanceRestartWaitDuration - Indicates the InstanceCloseDelayDuration property is set. The value is 32768.
         /// - DropSourceReplicaOnMove - Indicates the DropSourceReplicaOnMove property is set. The value is 65536.
         /// - ServiceDnsName - Indicates the ServiceDnsName property is set. The value is 131072.
-        /// - TagsForPlacement - Indicates the TagsForPlacement property is set. The value is 1048576.
-        /// - TagsForRunning - Indicates the TagsForRunning property is set. The value is 2097152.
+        /// - ServiceTags TagsRequiredToPlace - Indicates the TagsRequiredToPlace property is set. The value is 1048576.
+        /// - ServiceTags TagsRequiredToRun - Indicates the TagsRequiredToRun property is set. The value is 2097152.
```
gobradovic, line 46: Shouldn't here be RepartitionDescription?

gobradovic (approved):

src/Client.Http/Common/Generated/ServiceUpdateDescription.cs
olegsych: It looks that way. We'll need to make sense of the flags property at some point. It looks like the pattern we use in the native API, where we set a flags bit in the first struct of a linked chain. It makes no sense for an HTTP JSON API, but we keep doing it. Regardless, this is was generated from Swagger: https://msazure.visualstudio.com/One/_git/SF-Swagger?path=/src/spec/sfclient/src/definitions/Native/ServiceUpdateDescription.yaml&version=GBdevelop&line=40&lineEnd=41&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

That's where it needs to be fixed.

PR #526, opened 2026-01-31
Update markdown and script files for contributor onboarding

olegsych:
- Cleanup `README.md`
- Update `SECURITY.md` based on the latest Microsoft template.
- Simplify the `init.cmd` and include optional tools for Windows.
- Create `init.sh` for Ubuntu.
- Rewrite `CONTRIBUTING.md` with working Windows and Linux instructions, plus creating pull requests.

### DIFF
--- a/init.cmd
+++ b/init.cmd
@@ -1,11 +1,7 @@
-: Run as admin to configure local development pre-requisites
-
 winget install Microsoft.DotNet.AspNetCore.8
 winget install Microsoft.DotNet.AspNetCore.9
-winget install Microsoft.DotNet.AspNetCore.10
 winget install Microsoft.DotNet.Runtime.8
 winget install Microsoft.DotNet.Runtime.9
-winget install Microsoft.DotNet.Runtime.10
 winget install Microsoft.DotNet.SDK.10
 
 pwsh SkipStrongName.ps1
--- a/init.sh
+++ b/init.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+add-apt-repository -y ppa:dotnet/backports
+apt-get update
+
+apt-get install -y aspnetcore-runtime-8.0
+apt-get install -y aspnetcore-runtime-9.0
+apt-get install -y dotnet-sdk-10.0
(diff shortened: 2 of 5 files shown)


init.cmd
```diff
@@ -1,11 +1,22 @@
...
 winget install Microsoft.DotNet.AspNetCore.8
 winget install Microsoft.DotNet.AspNetCore.9
-winget install Microsoft.DotNet.AspNetCore.10
 winget install Microsoft.DotNet.Runtime.8
 winget install Microsoft.DotNet.Runtime.9
-winget install Microsoft.DotNet.Runtime.10
 winget install Microsoft.DotNet.SDK.10
+winget install Microsoft.PowerShell
 
+: Optional tools
```
plave0: Even though they are listed in the "Optional tools" section, they will be installed regardless. I think we should either make these tools opt-in, or they should not be installed by the init script at all. This init script should install only the dependencies which are required of this repo to be built.

init.cmd
```diff
@@ -1,11 +1,22 @@
-: Run as admin to configure local development pre-requisites
+: Run as admin
 
+: Required tools
+winget install Git.Git
```
plave0: Git has been added to both the "Required tools" and "Optional tools" section. But I also think Git is redundant here. Someone who has cloned this repo will already have git installed.

init.sh
```diff
@@ -0,0 +1,11 @@
...
+
+add-apt-repository -y ppa:dotnet/backports
+apt-get update
+
+apt-get install -y aspnetcore-runtime-8.0
+apt-get install -y aspnetcore-runtime-9.0
+apt-get install -y dotnet-sdk-10.0
+
+# Optional tools
+apt-get install -y powershell
```
plave0: The same as "Optional tools" in `init.cmd`

plave0 (changes requested):

init.cmd
olegsych: Good point. Removed.

init.cmd
olegsych: That's fair. Removed.

plave0 (approved):

PR #529, opened 2026-02-08
Add GetFailoverManagerManagerInformation API 

ccrystalwu:
Added GetFailoverManagerManagerInformation API to the Service Fabric .NET SDK cluster management APIs. This API returns the information (node id, node name, node instance id) of the node that FMM is on.


olegsych (changes requested):
Thanks @ccrystalwu. Could you update the PR description to explain what we're changing and why?
https://cbea.ms/git-commit#why-not-how

Also, it looks like the PR validation build failed on one of our flakey tests. We are fixing it but in the meantime need to rerun the failing stages. Could you make sure the build is green after you update your branch?

olegsych (approved):
:shipit:

PR #530, opened 2026-02-13
Cleanup ActorService

olegsych:
Remove
- Redundant comments
- Unnecessary `this.`
- Unnecessary fields
- Unreferenced methods left over from removed actor state migration

Also,
- Convert single-line methods to expression style.
- Convert properties to auto style

### DIFF
--- a/src/Actors/Runtime/ActorService.cs
+++ b/src/Actors/Runtime/ActorService.cs
@@ -12,10 +12,12 @@ using System.Threading.Tasks;
 using Microsoft.ServiceFabric.Actors.Diagnostics;
 using Microsoft.ServiceFabric.Actors.Query;
 using Microsoft.ServiceFabric.Actors.Remoting;
+using Microsoft.ServiceFabric.Actors.Remoting.V2.Runtime;
 using Microsoft.ServiceFabric.Diagnostics;
 using Microsoft.ServiceFabric.Diagnostics.Tracing;
 using Microsoft.ServiceFabric.Services;
 using Microsoft.ServiceFabric.Services.Communication.Runtime;
+using Microsoft.ServiceFabric.Services.Remoting.Runtime;
 using Microsoft.ServiceFabric.Services.Runtime;
 
 namespace Microsoft.ServiceFabric.Actors.Runtime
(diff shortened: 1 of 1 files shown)


src/Actors/Runtime/ActorService.cs
```diff
@@ -29,354 +31,170 @@ namespace Microsoft.ServiceFabric.Actors.Runtime
...
-        internal IActorStateManager CreateStateManager(ActorBase actor)
-        {
-            return this.stateManagerFactory.Invoke(actor, this.StateProvider);
-        }
+        internal IActorStateManager CreateStateManager(ActorBase actor) => 
+            stateManagerFactory.Invoke(actor, StateProvider);
 
         internal void InitializeInternal(ActorMethodFriendlyNameBuilder methodNameBuilder)
         {
-            this.methodFriendlyNameBuilder = methodNameBuilder;
```
olegsych, line 256: @ilijagrbic This looks suspicious to me because the InitializeInternal takes the methodNameBuilder, but it's never used. The constructor creates its own ActorMethodFriendlyNameBuilder instance and the only time it's needed is to create a DiagnosticsFactory.

ilijagrbic: Hey @olegsych , I haven't noticed this while I was initially working on this class. But it seems like something we shouldn't do. Old `methodFriendlyNameBuilder` field was only used by the `MethodFriendlyNameBuilder` property which is now not used anywhere. Previously, MethodFriendlyNameBuilder was fetched by other classes such as `ActorManager`, in order to use it in telemetry emission, but after we did some refactors, we removed these strong dependencies. 
I'm struggling to understand why this field would be initialized manually by the `ActorServiceFactory`, as this class is quite simple and self-contained.

Anyways, I don't see any issues with these changes, since we instantiate the field in the constructor. We could also remove the `methodNameBuilder` parameter as it is no longer used and updated the call sites, but this is also fine for now.

src/Actors/Runtime/ActorService.cs
```diff
@@ -29,354 +31,170 @@ namespace Microsoft.ServiceFabric.Actors.Runtime
...
-        /// <para>
         /// Failing to do so can cause failover, reconfiguration or upgrade of your actor service to get stuck and
-        /// can impact availibility of your service.
-        /// </para>
+        /// can impact availability of your service.
         /// </remarks>
-        protected override Task RunAsync(CancellationToken cancellationToken) =>
-            ActorManager.StartLoadingRemindersAsync(cancellationToken);
+        protected override Task RunAsync(CancellationToken cancellation) =>
+            ActorManager.StartLoadingRemindersAsync(cancellation);
```
ilijagrbic, line 149: Just wondering, why are we changing this? Is there a convention that should be followed for naming the `CancellationTokens`?

ilijagrbic (approved):

src/Actors/Runtime/ActorService.cs
olegsych: I personally don't like the term Token because it feels unnecessary in parameter names. I've never come access a function where dropping it from the parameter name made it ambiguous, so I've been doing it a while to make functions more compact.

olegsych: This is subjective of course, if you prefer otherwise, I think I can convince my OCD 😂

ilijagrbic: I definitely agree that method names/parameters should be as compact as possible, so this makes sense to me. I don't think that this particular case was that long, but in principle I agree. :)

olegsych: It sounds like our OCDs should talk 😆

olegsych:
Abandoning until we can get Copilot to reliably retrofit unit tests.

PR #533, opened 2026-02-20
Remove heap allocation from Meter Records

ilijagrbic:
Remove `string[]` allocation from the `Meter.Record()` to avoid heap allocation during this hot path

Change `IFabricMeter.Record` signiture from `string[]` to `IntPtr` for the `dimensionValues` parameter:
- `string[]` are subject to Marshalling with .NET interop, we are trying to avoid any copying that may occur as part of such marshalling, to achieve this, we are pinning all of the strings in memory, and passing an array of pointers to them
- Use `IntPtr` in favour of `char*/char**`, pointer types are unsafe and would not allow us to unit test this method, since the Moq API uses generic classes/methods, which can't accept pointer types.

### DIFF
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -67,23 +67,22 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
                 for (int i = 0; i < systemDimensionValues.Length; i++)
                 {
                     allDimensionPins[i] = GCHandle.Alloc(systemDimensionValues[i], GCHandleType.Pinned);
-                    allDimensionValuesPointers[i] = allDimensionPins[i].AddrOfPinnedObject();
                 }
 
                 if (customDimensionCount > 0)
-                {
                     allDimensionPins[systemDimensionValues.Length] = GCHandle.Alloc(customDimension1, GCHandleType.Pinned);
-                    allDimensionValuesPointers[systemDimensionValues.Length] = allDimensionPins[systemDimensionValues.Length].AddrOfPinnedObject();
-                }
+
                 if (customDimensionCount > 1)
-                {
                     allDimensionPins[systemDimensionValues.Length + 1] = GCHandle.Alloc(customDimension2, GCHandleType.Pinned);
-                    allDimensionValuesPointers[systemDimensionValues.Length + 1] = allDimensionPins[systemDimensionValues.Length + 1].AddrOfPinnedObject();
-                }
+
                 if (customDimensionCount > 2)
-                {
                     allDimensionPins[systemDimensionValues.Length + 2] = GCHandle.Alloc(customDimension3, GCHandleType.Pinned);
-                    allDimensionValuesPointers[systemDimensionValues.Length + 2] = allDimensionPins[systemDimensionValues.Length + 2].AddrOfPinnedObject();
+
+
+                for (int i = 0; i < totalDimensionCount; i++)
+                {
+                    // for strings, AddrOfPinnedObject() returns a pointer to the first character - https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.gchandle.addrofpinnedobject
+                    allDimensionValuesPointers[i] = allDimensionPins[i].AddrOfPinnedObject();
                 }
 
                 fabricMeter.Record(value, (uint)totalDimensionCount, (IntPtr)allDimensionValuesPointers);
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -88,6 +88,10 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 
                 fabricMeter.Record(value, (uint)totalDimensionCount, (IntPtr)allDimensionValuesPointers);
             }
+            catch (Exception)
+            {
+                throw;
+            }
             finally
             {
                 for (int i = 0; i < totalDimensionCount; i++)
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -6,6 +6,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.InteropServices;
 
 namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 {
@@ -37,30 +38,52 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
             recordAction.Invoke(value, customDimensionCount, customDimension1, customDimension2, customDimension3);
         }
 
-        private void RecordViaNative(long value, int customDimensionCount, string customDimension1, string customDimension2, string customDimension3)
+        unsafe private void RecordViaNative(long value, int customDimensionCount, string customDimension1, string customDimension2, string customDimension3)
         {
             if (customDimensionCount < 0 || customDimensionCount > 3)
             {
                 throw new ArgumentOutOfRangeException(nameof(customDimensionCount));
             }
 
-            var allDimensionArray = new string[systemDimensionValues.Length + customDimensionCount];
-            systemDimensionValues.CopyTo(allDimensionArray, 0);
+            int totalDimensionCount = systemDimensionValues.Length + customDimensionCount;
 
-            if (customDimensionCount > 0)
-            {
-                allDimensionArray[systemDimensionValues.Length] = customDimension1;
-            }
-            if (customDimensionCount > 1)
-            {
-                allDimensionArray[systemDimensionValues.Length + 1] = customDimension2;
-            }
-            if (customDimensionCount > 2)
-            {
-                allDimensionArray[systemDimensionValues.Length + 2] = customDimension3;
-            }
+            GCHandle* allDimensionPins = stackalloc GCHandle[totalDimensionCount];
+            IntPtr* allDimensionValuesPointers = stackalloc IntPtr[totalDimensionCount];
 
-            fabricMeter.Record(value, (uint)allDimensionArray.Length, allDimensionArray);
+            try
+            {
+                for (int i = 0; i < systemDimensionValues.Length; i++)
+                {
+                    allDimensionPins[i] = GCHandle.Alloc(systemDimensionValues[i], GCHandleType.Pinned);
+                    allDimensionValuesPointers[i] = allDimensionPins[i].AddrOfPinnedObject();
+                }
+
+                if (customDimensionCount > 0)
+                {
+                    allDimensionPins[systemDimensionValues.Length] = GCHandle.Alloc(customDimension1, GCHandleType.Pinned);
+                    allDimensionValuesPointers[systemDimensionValues.Length] = allDimensionPins[systemDimensionValues.Length].AddrOfPinnedObject();
+                }
+                if (customDimensionCount > 1)
+                {
+                    allDimensionPins[systemDimensionValues.Length + 1] = GCHandle.Alloc(customDimension2, GCHandleType.Pinned);
+                    allDimensionValuesPointers[systemDimensionValues.Length + 1] = allDimensionPins[systemDimensionValues.Length + 1].AddrOfPinnedObject();
+                }
+                if (customDimensionCount > 2)
+                {
+                    allDimensionPins[systemDimensionValues.Length + 2] = GCHandle.Alloc(customDimension3, GCHandleType.Pinned);
+                    allDimensionValuesPointers[systemDimensionValues.Length + 2] = allDimensionPins[systemDimensionValues.Length + 2].AddrOfPinnedObject();
+                }
+
+                fabricMeter.Record(value, (uint)totalDimensionCount, (IntPtr)allDimensionValuesPointers);
+            }
+            finally
+            {
+                for (int i = 0; i < totalDimensionCount; i++)
+                {
+                    if (allDimensionPins[i].IsAllocated)
+                        allDimensionPins[i].Free();
+                }
+            }
         }
     }
 }
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -35,16 +35,28 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 
         protected void Record(long value, int customDimensionCount = 0, string customDimension1 = null, string customDimension2 = null, string customDimension3 = null)
         {
+            if (customDimensionCount < 0 || customDimensionCount > 3)
+            {
+                throw new ArgumentOutOfRangeException(nameof(customDimensionCount));
+            }
+            if (customDimension3 == null && customDimensionCount == 3)
+            {
+                throw new ArgumentException(nameof(customDimension3));
+            }
+            if (customDimension2 == null && customDimensionCount >= 2)
+            {
+                throw new ArgumentException(nameof(customDimension2));
+            }
+            if (customDimension1 == null && customDimensionCount >= 1)
+            {
+                throw new ArgumentException(nameof(customDimension1));
+            }
+
             recordAction.Invoke(value, customDimensionCount, customDimension1, customDimension2, customDimension3);
         }
 
         unsafe private void RecordViaNative(long value, int customDimensionCount, string customDimension1, string customDimension2, string customDimension3)
         {
-            if (customDimensionCount < 0 || customDimensionCount > 3)
-            {
-                throw new ArgumentOutOfRangeException(nameof(customDimensionCount));
-            }
-
             int totalDimensionCount = systemDimensionValues.Length + customDimensionCount;
 
             GCHandle* allDimensionPins = stackalloc GCHandle[totalDimensionCount];
--- a/test/Diagnostics/Metrics/Implementation/MeterTest.cs
+++ b/test/Diagnostics/Metrics/Implementation/MeterTest.cs
@@ -101,17 +101,17 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
                     sutMethod.Invoke(value, fuzzy.Int32().Minimum(4), customDimension1, customDimension2, customDimension3));
             }
 
-            [Fact]
-            public void ThrowsExceptionIfCustomDimensionIsExpectedButNull()
+            [Theory]
+            [InlineData(null, "DimensionValue", "DimensionValue")]
+            [InlineData("DimensionValue", null, "DimensionValue")]
+            [InlineData("DimensionValue", "DimensionValue", null)]
+            [InlineData("DimensionValue", null, null)]
+            [InlineData(null, "DimensionValue", null)]
+            [InlineData(null, null, "DimensionValue")]
+            [InlineData(null, null, null)]
+            public void ThrowsExceptionIfCustomDimensionIsExpectedButNull(string customDimension1, string customDimension2, string customDimension3)
             {
-                Assert.Throws<ArgumentException>(() =>
-                    sutMethod.Invoke(value, 1, null, null, null));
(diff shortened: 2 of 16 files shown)


test/Diagnostics/Properties/launchSettings.json
```diff
@@ -0,0 +1,9 @@
+{
+  "profiles": {
+    "Microsoft.ServiceFabric.Diagnostics.Tests": {
+      "commandName": "Project",
+      "nativeDebugging": true,
```
ilijagrbic: By default, enable native debugging for this Assembly.
Not sure if these kinds of settings should be stored in the repo, let me know if this should be omitted.

src/Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -33,34 +34,72 @@ protected long ConvertTimeSpanToLong(TimeSpan value)
...
+                    allDimensionValuesPointers[systemDimensionValues.Length + 2] = allDimensionPins[systemDimensionValues.Length + 2].AddrOfPinnedObject();
+                }
+
+                fabricMeter.Record(value, (uint)totalDimensionCount, (IntPtr)allDimensionValuesPointers);
             }
-            if (customDimensionCount > 1)
+            catch (Exception)
             {
-                allDimensionArray[systemDimensionValues.Length + 1] = customDimension2;
+                throw;
```
ilijagrbic, line 92: I'm not sure if the try-catch is really needed here. 
My reasoning was that I want to make sure that even in the case some unexpected exception occurs, we free the pins from memory, to avoid a potential memory leak. But should we also propagate the exceptions if they do occur, since they could crash code that is using metrics? 
My thoughts are that there are no expected exceptions that can occur here. Also, I would want to avoid silent failing of emission, so having all that in mind, I think it makes sense to rethrow the exception and let the caller deal with it. Let me know what you think @olegsych. :)

P.S I decided against the `GetPinnableReference` approach in favour of GCHandles, since that API is not available on .NETFramework, and I wanted to avoid having different code on .NETCore and .NETFramework.

test/Diagnostics/Properties/launchSettings.json
olegsych: I haven't encountered this type of file before. It looks like a VSCode config file. Why is it here rather than in the /.vscode folder? If it needs to be here, can we place it in the same folder with the project file?

src/Diagnostics/Metrics/Implementation/Meter.cs
olegsych: I don't know what it compiles into, but catching all exceptions and rethrowing them like this doesn't do anything functionally because the finally section is independent of any catch sections the try statement may have.

Could you expand the PR description to explain why we're using IntPtr/GCHandles instead of fixed char pointers?

src/Diagnostics/Metrics/Implementation/Meter.cs
olegsych, line 27: An extra array allocation is probably not that big of a deal but why do we convert an IEnumerable<string> to a mutable string[] here? My guess is that we need to determine the number of system dimensions so that we aren't recalculating them for each Record call. If that's the case, why not take IReadOnlyCollection<string> instead of IEnumerable<string> as a parameter?

src/Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -33,34 +34,72 @@ protected long ConvertTimeSpanToLong(TimeSpan value)
...
-        {
-            recordAction.Invoke(value, customDimensionCount, customDimension1, customDimension2, customDimension3);
-        }
-
-        private void RecordViaNative(long value, int customDimensionCount, string customDimension1, string customDimension2, string customDimension3)
         {
             if (customDimensionCount < 0 || customDimensionCount > 3)
             {
                 throw new ArgumentOutOfRangeException(nameof(customDimensionCount));
             }
```
olegsych, line 41: Could you maintain blank lines after closing braces for readability? Our coding style makes an exception only if the next line is another closing brace. But in this particular case, I would omit the braces entirely because a single statement inside an if is readable enough. In fact, I find braces around single-line statements to be less readable because they reduce density of the code and force you to read more blank lines, which is a form of noise.

src/Diagnostics/Metrics/Implementation/Meter.cs
olegsych, line 15: Why do we need an extra test seam - the recordAction field if we already have IFabricMeter we can mock?

src/Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -33,34 +34,72 @@ protected long ConvertTimeSpanToLong(TimeSpan value)
...
-                allDimensionArray[systemDimensionValues.Length + 1] = customDimension2;
+                throw;
             }
-            if (customDimensionCount > 2)
+            finally
             {
-                allDimensionArray[systemDimensionValues.Length + 2] = customDimension3;
+                for (int i = 0; i < totalDimensionCount; i++)
+                {
+                    if (allDimensionPins[i].IsAllocated)
```
olegsych, line 98: Why do we need to check if the GCHandle is allocated when the totalDimensionCount already accounts for that?

src/Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -33,34 +34,72 @@ protected long ConvertTimeSpanToLong(TimeSpan value)
...
+            IntPtr* allDimensionValuesPointers = stackalloc IntPtr[totalDimensionCount];
+
+            try
             {
-                allDimensionArray[systemDimensionValues.Length] = customDimension1;
+                for (int i = 0; i < systemDimensionValues.Length; i++)
+                {
+                    allDimensionPins[i] = GCHandle.Alloc(systemDimensionValues[i], GCHandleType.Pinned);
+                    allDimensionValuesPointers[i] = allDimensionPins[i].AddrOfPinnedObject();
+                }
```
olegsych, line 70: Can we extract the two statements we repeat for every dimension value into a separate function? I think this would help to
- Reduce duplication, obviously
- Introduce a local variable for GCHandle and eliminate unnecessary indexing into the allDimensionsPins just to get its AddrOfPinnedObject
- Add a comment explaining that AddrOfPinnedObject returns address of the first character for strings. This is not obvious from the API itself and makes this code impossible to understand without reading the GCHandle docs.

We could also separate initialization of the GCHandle array and the IntPtr arrays to reduce duplication and complexity without extracting the function. The function would need two ref or out parameters, which is complexity on its own, so this might be simpler.

I think I care less about redundant indexing than about reducing duplication and adding a comment explaining the AddrOfPinnedObject magic.

src/Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -33,34 +34,72 @@ protected long ConvertTimeSpanToLong(TimeSpan value)
...
-            recordAction.Invoke(value, customDimensionCount, customDimension1, customDimension2, customDimension3);
-        }
-
-        private void RecordViaNative(long value, int customDimensionCount, string customDimension1, string customDimension2, string customDimension3)
         {
             if (customDimensionCount < 0 || customDimensionCount > 3)
             {
                 throw new ArgumentOutOfRangeException(nameof(customDimensionCount));
             }
+            if (customDimension3 == null && customDimensionCount == 3)
```
olegsych, line 42: I like validation of arguments, but this significantly increases cyclomatic complexity. I think it's a consequence of using procedural logic for what would idiomatically be implemented as separate sub classes. I think we chose this implementation because an idiomatic class hierarchy would be more complex in the beginning. Perhaps now would be a good time to re-evaluate. I suspect the next time we'll need to do this when someone needs a fourth dimension and we'll break the Open/Closed principle and keep extending this class instead.

test/Diagnostics/Metrics/Implementation/MeterTest.cs
```diff
@@ -100,44 +101,87 @@ public void ThrowsExceptionIfNumberOfCustomDimenionsHigherThanSupported()
                     sutMethod.Invoke(value, fuzzy.Int32().Minimum(4), customDimension1, customDimension2, customDimension3));
             }
 
+            [Fact]
```
olegsych: Can we use a parameterized Theory instead of packing multiple tests into a single Fact.

test/Diagnostics/Metrics/Implementation/MeterTest.cs
```diff
@@ -100,44 +101,87 @@ public void ThrowsExceptionIfNumberOfCustomDimenionsHigherThanSupported()
...
+                    sutMethod.Invoke(value, 3, customDimension1, customDimension2, null));
+            }
+
             [Fact]
             public void CallsNativeMeterRecordWithZeroCustomDimensionsAndAllSystemDimensions()
             {
                 var expectedArray = systemDimensions.ToArray();
+                Mock.Get(fabricMeter)
+                    .Setup(m => m.Record(It.IsAny<long>(), It.IsAny<uint>(), It.IsAny<IntPtr>()))
+                    .Callback<long, uint, IntPtr>((value, count, stringPtrs) => Assert.True(StringPointersAreExpected(stringPtrs, count, expectedArray)));
```
olegsych: Does this produce comprehensible failure messages? I try to avoid putting assertions into the arrange logic because they are usually harder to understand and debug. Instead, I use arrange logic to capture the actual values so that they can be verified in the assert section. This is available out of the box for managed types with Moq, but we need to do this by hand here.

If you do want to keep the verification logic as part of the arrange section, I would move asserts into the validation function itself, something like AssertExpectedDimensionValues(IEnumerable<string> expected, uint count, IntPtr actual), to make the asserts more expressive than a pass/fail.

test/Diagnostics/Properties/launchSettings.json
olegsych: Found this: https://code.visualstudio.com/docs/csharp/debugger-settings#_approach-2-launchjson, not sure if we can accomplish the same thing with /.vscode/launch.json. Intuitively, I dislike the launchSettings.json in the project sub-folder because of the long history of VS creating junk files that people didn't understand and ended up committing to source control. The /.vscode folder is cleaner and self-explanatory, so if this configuration will be useful for others or even if it's just for you on ongoing basis, I would put it there. But if it's just a one-time thing, I wouldn't commit it. It looks like a fairly basic debugger config that anyone can re-create quickly.

olegsych (commented):
Thanks @ilijagrbic This is functionally solid. I would only recommend a few more iterations of refactoring before we move on, so that the code is left in a state that will be easy to maintain by others in the future. Lots of questions in the comments, some of them I probably already asked before and forgot 😅

olegsych (approved):

src/Diagnostics/Metrics/Implementation/Meter.cs
ilijagrbic: Agreed, I still have the habit of adding braces around single line statements. Thanks for noticing this, I will fix it.

src/Diagnostics/Metrics/Implementation/Meter.cs
ilijagrbic: My reasoning was that since `Free()` method throws an exception if `GCHandle` is not allocated I wanted to avoid a case where new exception is thrown in the finally block, which could cause some of the allocated handles to not be freed. It is definitely not necessary in a happy path, and I expect the handles to be allocated in the regular case.

src/Diagnostics/Metrics/Implementation/Meter.cs
ilijagrbic: It was used for testing of Int64Meter/TimeSpanMeter classes, which extend this class. If you think this should be avoided at all costs, I can refactor it in a separate PR, since I want to keep the changeset for this on as concise as possible.

test/Diagnostics/Metrics/Implementation/MeterTest.cs
ilijagrbic: Good point, thanks for noticing, I will fix it.

src/Diagnostics/Metrics/Implementation/Meter.cs
ilijagrbic: I would not change the API design significantly this late in the project. What I do think can be done here is to move the validations away from the base class to the extending classes ie. Int64Meter1D, Int64Meter2D etc. Validation would be much simpler there since we know with how many dimensions we deal with.

Let me do this refactoring, and we can then do another review pass. :)

test/Diagnostics/Properties/launchSettings.json
ilijagrbic: Agreed, I will skip it for now.

src/Diagnostics/Metrics/Implementation/Meter.cs
olegsych: These are `.Implementation` classes though, so we have the freedom to change them without affecting the user code. But even if they were used directly, I think adding an extra class in the inheritance hierarchy would still be a build- and run-time compatible change. To be clear, I think we should consider introducing Meter1D, Meter2D and Meter3D between Int64Meter1D, etc. and Meter. That way this logic could also be reused by the TimeSpanMeter1D, etc.

src/Diagnostics/Metrics/Implementation/Meter.cs
olegsych: I don't think it is as impactful as heap allocations, but it is an extra virtual method call. I also think that eliminating it would make the product code _simpler_ in terms of the four rules of simple design, although some of this complexity would shift to the code mocking the optimized IFabricMeter interface. I do like how well you've encapsulated it in the MeterTest, so if we make this logic reusable in Int63/TimeSpan meter tests, I think the tests wouldn't lose much _simplicity_, if any.

src/Diagnostics/Metrics/Implementation/Meter.cs
olegsych: Good point. I looked at the managed and native implementation, and I realized that I assumed GCHandle.Alloc to be much simpler than it actually is. How does this implementation compare to what the fixed keyword would do with String.GetPinnableReference?

ilijagrbic: I decided against the _fixed + GetPinnableReference_ implementation since it has some problems I think would not be worth solivng:
- First, `GetPinnableReference` API is only available on .NETCore, so we would have to maintain a different implementation for .NETFramework or figure out a different way to get the pinnable reference.
- `GetPinnableReference` is not [intended to be called by the user code.](https://learn.microsoft.com/en-us/dotnet/api/system.string.getpinnablereference?view=net-10.0)
- Another issue is that we need to pin all of the strings, and I don't think there is a way to use the fixed statement with variable number of fixed variables - for example - we can pin multiple strings, but only if we know which ones in advance, and using a loop around it would not be sufficient, since we need all of the strings to be pinned during the call. It's quite interesting, it turns out that `string[]` interop especially in our use case is not trivial at all, even though it seems simple at first sight.

Perhaps that approach could be made to work, but I only benefit I see is the usage of C# language construct (`fixed`), while on the other hand having unified implementation for both .NET versions and we are using official API to pin the strings. Also, the code here is not simple, but it is relatively self-explanatory what we are trying to do.

src/Diagnostics/Metrics/Implementation/Meter.cs
ilijagrbic: This conversion allowed us to initially lower the number of heap allocation in `Record()` method and use `Array.CopyTo()` method to copy the values of system dimensions before emission. This is in addition to what you pointed out about the count of system dimensions.

I will think about the usage of `IReadOnlyCollection`, and possibly tackle it in a follow up PR (alongside with some other code improvements)

ilijagrbic: Makes sense, I will take note and work on it in a follow up.

src/Diagnostics/Metrics/Implementation/Meter.cs
ilijagrbic: Yeah, I forgot to write a follow up on this, but I came to a similar conclusion. I will keep the change as is in this PR and I will do the refactoring in a follow up PR.

src/Diagnostics/Metrics/Implementation/Meter.cs
ilijagrbic: I understand that finally section is independent. I want to have a try-finally section in order to make sure we are releasing the pins from memory. 
The main question I have is how should be handle the exceptions in this case. I see 2 ways:
1. We don't rethrow the exceptions, but just trace an error
2. We rethrow the exceptions and let the caller deal with them
With option 1. I'm concerned that if metrics fail, we will not know they are failing (unless we check the traces), so they could fail silently. 
With option 2. my concern is that in case there is a bug in the metrics code, exception would occur in the caller's code. So, users of the API would have to think about how to handle unexpected exceptions, and for example if they should: ignore any metrics related exceptions, trace error messages or also rethrow the exception. 
I'm fairly certain we should go with the option 2., but I wanted to sanity-check my thought process.

Also, I have expanded the PR description.

src/Diagnostics/Metrics/Implementation/Meter.cs
ilijagrbic: Thanks Oleg, the complexity of a function with two ref parameters was why I decided against extracting this logic into a separate function. But I like the idea to separate the initialization, and I will add a comment explaining the `magic` :)

Regarding indexing, I don't think it should be a huge deal performance wise, in case we need to hyper-optimize it further, we can look into removing the indexing as well as the virtual function calls you called out.

test/Diagnostics/Metrics/Implementation/MeterTest.cs
ilijagrbic: This was a great suggestion, it seems obvious now, but I haven't tried to capture the values. Now I also managed to simplify the setup logic itself and move it from the individual methods into the constructor. :)

src/Diagnostics/Metrics/Implementation/Meter.cs
olegsych: Not sure if you meant to fix it in a follow-up PR, reactivating in case you meant to fix it here.

src/Diagnostics/Metrics/Implementation/Meter.cs
olegsych: This is _framework code_, so the .NET guidelines are clear: DO NOT swallow errors by catching nonspecific exceptions, such as System.Exception, System.SystemException, and so on, in framework code.
https://learning.oreilly.com/library/view/framework-design-guidelines/9780135896457/ch07.xhtml#sec7_2_2

It's ironic that we're discussing the same topic I debated with the Application Insights team roughly 10 years ago. The leadership decided that "thou shall not throw" and we swallowed all exceptions before returning to the user code. As you can imagine, there were many painful RCAs of monitoring blindness outages; it's too bad that leadership wasn't involved in them. 

I still believe that the instrumentation APIs should [fail fast](https://martinfowler.com/ieeeSoftware/failFast.pdf), probably even more today because of the ever-increasing emphasis on security. I think a monitoring blindness during a security incident is much worse than an incident caused by a bug in the instrumentation. The latter is an annoying problem you'll discover and have to fix during the first broad rollout, the former is a time bomb that might be ticking for years before it goes off, and that's if you're lucky to detect it.

Besides, the OutOfMemoryException, which is the most likely, if not the only exception type that can be thrown by GCHandle.Alloc cannot be caught by the managed code anyway.

olegsych (approved):

src/Diagnostics/Metrics/Implementation/Meter.cs
ilijagrbic: It is fixed in a follow up PR, resolving for that reason. :)

src/Diagnostics/Metrics/Implementation/Meter.cs
ilijagrbic: Yeah, totally agree with you Oleg! I also had similar experiences in my career. My reasoning why I was even considering an option to not re-throw was that we would not be sacrificing any functionality in case system starts to silently fail. But yeah, ideally it should fail fast, so I will keep the change as is. Thanks for talking through the change with me. :)

PR #537, opened 2026-03-03
Handle null values in replica and instance factories

plave0:
### Null validation

 - Constructor parameters runtimeContext and serviceFactory now throw ArgumentNullException when null.
 - CreateReplica / CreateInstance throw InvalidOperationException when the user-supplied factory delegate returns null.
 - Dispose() no longer uses null-conditional on runtimeContext since null is rejected at construction.

### Testability

 - RuntimeContext.Dispose() made virtual to allow mocking in tests.

### Code modernization

 - File-scoped namespaces, removed redundant this. qualifiers, expression-bodied Dispose(), removed unnecessary private 
on fields.

### Tests

 - Added StatefulServiceReplicaFactoryTest and StatelessServiceInstanceFactoryTest covering:
  - Constructor null-argument validation
  - CreateReplica / CreateInstance delegates to the service factory with the correct ServiceContext
  - Returns the expected adapter wrapping the service
  - Throws when the factory delegate returns null
  - Dispose forwards to RuntimeContext


src/Services/Runtime/StatelessServiceInstanceFactory.cs
```diff
@@ -3,47 +3,43 @@
...
+            serviceName,
+            initializationData,
+            partitionId,
+            instanceId);
 
-        public void Dispose()
-        {
-            this.runtimeContext?.Dispose();
-        }
+        StatelessService service = serviceFactory(instanceContext) ?? throw new InvalidOperationException($"{nameof(serviceFactory)} return null");
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Typo in production exception message: `"return null"` should be `"returned null"` to match the message in `StatefulServiceReplicaFactory.cs`.

test/Services/Runtime/StatefulServiceReplicaFactoryTest.cs
```diff
@@ -0,0 +1,136 @@
...
+    protected StatefulServiceReplicaFactoryTest()
+    {
+        runtimeContext = Mock.Of<RuntimeContext>(ctx =>
+            ctx.NodeContext == fuzzy.NodeContext() &&
+            ctx.CodePackageContext == fuzzy.ICodePackageActivationContext());
+
+        serviceFactory = Mock.Of<Func<StatefulServiceContext, StatefulServiceBase>>();
+    }
+
+    StatefulServiceReplicaFactory CreateSut() =>
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Per `test.instructions.md` and the conventions in `StatefulServiceReplicaAdapterTest` / `StatelessServiceInstanceAdapterTest`, declare a `sut` field at the top of the base test class, followed by `// Constructor parameters` then `// Test fixture` (with `fuzzy`). This file uses a `CreateSut()` factory method and places `fuzzy` before the constructor parameters. Same applies to `StatelessServiceInstanceFactoryTest.cs`.

test/Services/Runtime/StatefulServiceReplicaFactoryTest.cs
```diff
@@ -0,0 +1,136 @@
...
+                sut.CreateReplica(serviceTypeName, serviceName, initializationData, partitionId, replicaId));
+        }
+    }
+
+    public sealed class Dispose : StatefulServiceReplicaFactoryTest
+    {
+        [Fact]
+        public void DisposesRuntimeContext()
+        {
+            var runtimeContextMock = Mock.Get(runtimeContext);
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") `runtimeContextMock` redundantly suffixes `Mock` and shadows the field. Use a short local name like `runtime`, or call `Mock.Get(runtimeContext).Verify(...)` directly. Same applies to `StatelessServiceInstanceFactoryTest.cs`.

src/Services/Runtime/StatefulServiceReplicaFactory.cs
```diff
@@ -3,47 +3,43 @@
...
 
-        public StatefulServiceReplicaFactory(
-            RuntimeContext runtimeContext,
-            Func<StatefulServiceContext, StatefulServiceBase> serviceFactory)
-        {
-            this.serviceFactory = serviceFactory;
-            this.runtimeContext = runtimeContext;
-        }
+namespace Microsoft.ServiceFabric.Services.Runtime;
+internal class StatefulServiceReplicaFactory : IStatefulServiceFactory, IDisposable
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Inconsistent visibility on sibling factory types: this declares `internal class` while `StatelessServiceInstanceFactory` declares `class`. Per the C# conventions, drop the redundant `internal` keyword here so the two siblings match.

src/Services/Runtime/RuntimeContext.cs
```diff
@@ -93,7 +93,7 @@ public static async Task<RuntimeContext> GetOrCreateAsync(
             return sharedContext;
         }
 
-        public void Dispose()
+        public virtual void Dispose()
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Making `Dispose()` `virtual` solely so Moq can `Verify(c => c.Dispose())` leaks a test-framework concern into production polymorphism. Consider verifying disposal via observable state (e.g., asserting that `Runtime`/`CodePackageContext` were disposed, or that subsequent operations fail) instead of mocking the method. Impact is bounded since the class is `internal`, but the design signal is worth addressing.

test/Services/Runtime/StatefulServiceReplicaFactoryTest.cs
```diff
@@ -0,0 +1,136 @@
...
+
+    public sealed class CreateReplica : StatefulServiceReplicaFactoryTest
+    {
+        readonly string serviceTypeName = fuzzy.String();
+        readonly Uri serviceName = fuzzy.Uri();
+        readonly byte[] initializationData = fuzzy.Array(fuzzy.Byte);
+        readonly Guid partitionId = Guid.NewGuid();
+        readonly long replicaId = fuzzy.Int64();
+
+        readonly Mock<StatefulServiceBase> mockService;
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Per `moq.instructions.md`, don't include `Mock` in `Mock<T>` variable names. Rename `mockService` → `service` (and apply the same to `StatelessServiceInstanceFactoryTest.cs`).

src/Services/Runtime/StatelessServiceInstanceFactory.cs
```diff
@@ -3,47 +3,43 @@
 // Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
 // ------------------------------------------------------------
 
-namespace Microsoft.ServiceFabric.Services.Runtime
-{
-    using System;
-    using System.Fabric;
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [💡](# "Suggestion") Extra blank line between the copyright header and the first `using`. Existing files in the repo use a single blank line — applies to `StatefulServiceReplicaFactory.cs` too.

olegsych (changes requested):
[:copilot:](https://docs.github.com/copilot/responsible-use/code-review) ⚠️ **Needs Changes** — fail-fast null validation and tests are correct and CI is green, but a missing PR description and a typo in a new exception message must be fixed before merge; several convention deviations should also be addressed.

## Detailed Assessment

**What it does.** Adds `ArgumentNullException` guards on constructor parameters of `StatefulServiceReplicaFactory` and `StatelessServiceInstanceFactory`; throws `InvalidOperationException` when the user-supplied `serviceFactory` delegate returns `null`; makes `RuntimeContext.Dispose()` `virtual` to enable Moq verification; reformats both factory files to file-scoped namespaces; adds unit tests for both factories.

**Why it exists.** Without these guards, null inputs/returns fail later as `NullReferenceException` at `CreateReplica`/`CreateInstance` or when the adapter accesses `service.Context`. Surfacing the failure at the boundary improves diagnosability.

**Approach.** Correct location for the checks. Note: `ServiceRuntime.RegisterServiceAsync` already validates `serviceFactory` via `Requires.ThrowIfNull`, so the constructor guard is defense-in-depth — acceptable for an internal class.

**Cost-Benefit.** Net positive. Small, focused runtime change; the structural reformat is acceptable since the files are substantially modified. Tests are thorough.

## Issues

### ⚠️ Missing PR description — merge-blocking
The PR body is empty. [CONTRIBUTING.md](https://github.com/microsoft/service-fabric-dotnet/blob/develop/CONTRIBUTING.md) requires the PR description to explain *what* changed and *why* (not *how*). Please add the motivation (e.g., observed symptom or linked issue).

@plave0 please take a look at the inline comments and the missing PR description.

src/Services/Runtime/StatefulServiceReplicaFactory.cs
olegsych: Could you add blank lines after namespace declarations?

src/Services/Runtime/StatelessServiceInstanceFactory.cs
olegsych: Actually, I would remove this entire throw change because the StatefulServiceReplicaAdapter throws ArgumentNullException, so additional validation here is redundant.

src/Services/Runtime/StatefulServiceReplicaFactory.cs
```diff
@@ -3,47 +3,43 @@
...
+using System.Fabric;
 
-        public StatefulServiceReplicaFactory(
-            RuntimeContext runtimeContext,
-            Func<StatefulServiceContext, StatefulServiceBase> serviceFactory)
-        {
-            this.serviceFactory = serviceFactory;
-            this.runtimeContext = runtimeContext;
-        }
+namespace Microsoft.ServiceFabric.Services.Runtime;
```
olegsych, line 9: These files are small enough, so the review is manageable. But for the future, let's do style changes in separate PRs that don't change any functionality, so that functional changes are easier to review.

olegsych (changes requested):

src/Services/Runtime/StatelessServiceInstanceFactory.cs
plave0: We need this check, because if `serviceFactory` returns null then `service.Context` in the next line will throw a null exception. 

This validation is the motivation for this PR. I worked on an incident where the customer's factory method was returning null, but the null reference exception was not clear enough.

src/Services/Runtime/RuntimeContext.cs
plave0: I went with creating a mock of `CodePackageContext` which is disposed when the runtime context is disposed. We can verify that the `Dispose` in the mock `CodePackageContext` is called.

test/Services/Runtime/StatelessServiceInstanceFactoryTest.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        [Fact]
+        public void ThrowsInvalidOperationExceptionWhenServiceFactoryReturnsNull()
+        {
+            Mock.Get(serviceFactory)
+                .Setup(f => f(It.IsAny<StatelessServiceContext>()))
+                .Returns((StatelessService)null);
+
+            Assert.Throws<InvalidOperationException>(() =>
+                sut.CreateInstance(serviceTypeName, serviceName, initializationData, partitionId, instanceId));
```
olegsych, line 109: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [❗](# "Must fix") This test asserts only the exception type, but the actionable diagnostic message is the feature this PR introduces — the whole motivation is replacing an unclear `NullReferenceException` with a clear failure. A regression to an unclear or typoed message would still pass. Capture the exception and assert its message, e.g. `var exception = Assert.Throws<InvalidOperationException>(...); Assert.Equal($"{nameof(serviceFactory)} returned null", exception.Message);`. See `test.instructions.md` on asserting exception properties.

test/Services/Runtime/StatefulServiceReplicaFactoryTest.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        [Fact]
+        public void ThrowsInvalidOperationExceptionWhenServiceFactoryReturnsNull()
+        {
+            Mock.Get(serviceFactory)
+                .Setup(f => f(It.IsAny<StatefulServiceContext>()))
+                .Returns((StatefulServiceBase)null);
+
+            Assert.Throws<InvalidOperationException>(() =>
+                sut.CreateReplica(serviceTypeName, serviceName, initializationData, partitionId, replicaId));
```
olegsych, line 115: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [❗](# "Must fix") This test asserts only the exception type, but the actionable diagnostic message is the feature this PR introduces — the whole motivation is replacing an unclear `NullReferenceException` with a clear failure. A regression to an unclear or typoed message (the earlier `"return null"` vs `"returned null"` thread shows this is a real risk) would still pass. Capture the exception and assert its message, e.g. `var exception = Assert.Throws<InvalidOperationException>(...); Assert.Equal($"{nameof(serviceFactory)} returned null", exception.Message);`. See `test.instructions.md` on asserting exception properties.

src/Services/Runtime/StatefulServiceReplicaFactory.cs
```diff
@@ -3,47 +3,43 @@
...
-        {
-            var serviceContext = new StatefulServiceContext(
-                this.runtimeContext.NodeContext,
-                this.runtimeContext.CodePackageContext,
-                serviceTypeName,
-                serviceName,
-                initializationData,
-                partitionId,
-                replicaId);
+    public StatefulServiceReplicaFactory(
```
olegsych, line 16: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") This `public` constructor lives on an `internal` class, so its effective visibility is `internal`. The only callers are `ServiceRuntime.cs` (same assembly) and the tests (via `InternalsVisibleTo`), so declare it `internal`. `csharp.instructions.md`: "the effective visibility of a public method in an internal class is internal, specify internal instead."

src/Services/Runtime/StatelessServiceInstanceFactory.cs
```diff
@@ -3,47 +3,43 @@
...
-        {
-            var instanceContext = new StatelessServiceContext(
-                this.runtimeContext.NodeContext,
-                this.runtimeContext.CodePackageContext,
-                serviceTypeName,
-                serviceName,
-                initializationData,
-                partitionId,
-                instanceId);
+    public StatelessServiceInstanceFactory(
```
olegsych, line 16: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") This `public` constructor lives on an `internal` class, so its effective visibility is `internal`. The only callers are `ServiceRuntime.cs` (same assembly) and the tests (via `InternalsVisibleTo`), so declare it `internal`. `csharp.instructions.md`: "the effective visibility of a public method in an internal class is internal, specify internal instead."

src/Services/Runtime/StatefulServiceReplicaFactory.cs
```diff
@@ -3,47 +3,43 @@
...
+namespace Microsoft.ServiceFabric.Services.Runtime;
 
-        public StatefulServiceReplicaFactory(
-            RuntimeContext runtimeContext,
-            Func<StatefulServiceContext, StatefulServiceBase> serviceFactory)
-        {
-            this.serviceFactory = serviceFactory;
-            this.runtimeContext = runtimeContext;
-        }
+class StatefulServiceReplicaFactory : IStatefulServiceFactory, IDisposable
```
olegsych, line 11: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Nothing derives from this type (verified across `src` and `test`), so mark it `sealed`. `csharp.instructions.md`: "Make types and members abstract or sealed if possible."

src/Services/Runtime/StatelessServiceInstanceFactory.cs
```diff
@@ -3,47 +3,43 @@
...
+namespace Microsoft.ServiceFabric.Services.Runtime;
 
-        public StatelessServiceInstanceFactory(
-            RuntimeContext runtimeContext,
-            Func<StatelessServiceContext, StatelessService> serviceFactory)
-        {
-            this.runtimeContext = runtimeContext;
-            this.serviceFactory = serviceFactory;
-        }
+class StatelessServiceInstanceFactory : IStatelessServiceFactory, IDisposable
```
olegsych, line 11: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Nothing derives from this type (verified across `src` and `test`), so mark it `sealed`. `csharp.instructions.md`: "Make types and members abstract or sealed if possible."

test/Services/Runtime/StatefulServiceReplicaFactoryTest.cs
```diff
@@ -0,0 +1,134 @@
...
+
+public abstract class StatefulServiceReplicaFactoryTest
+{
+    readonly IStatefulServiceFactory sut;
+
+    // Constructor parameters
+    readonly RuntimeContext runtimeContext;
+    readonly Func<StatefulServiceContext, StatefulServiceBase> serviceFactory;
+
+    // Test fixture
```
olegsych, line 24: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [💡](# "Consider changing") `test.instructions.md` prescribes only `// Constructor parameters` and `// Method parameters` section comments, and its example shows `fuzzy` with no section comment. Consider dropping this `// Test fixture` label for consistency.

test/Services/Runtime/StatefulServiceReplicaFactoryTest.cs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    public sealed class Dispose : StatefulServiceReplicaFactoryTest
+    {
+        [Fact]
+        public void DisposesRuntimeContext()
+        {
+            var codePackageContext = new Mock<ICodePackageActivationContext>();
+            var runtimeContext = new RuntimeContext();
+            runtimeContext.Field<ICodePackageActivationContext>().Set(codePackageContext.Object);
```
olegsych, line 126: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") This local `runtimeContext` (and the local `sut` below) shadow the base class fields of the same name, which hurts readability. Rename the locals to make the distinction clear — this is the residue of an earlier resolved shadowing thread.

test/Services/Runtime/StatefulServiceReplicaFactoryTest.cs
```diff
@@ -0,0 +1,134 @@
...
+    readonly IStatefulServiceFactory sut;
+
+    // Constructor parameters
+    readonly RuntimeContext runtimeContext;
+    readonly Func<StatefulServiceContext, StatefulServiceBase> serviceFactory;
+
+    // Test fixture
+    static readonly IFuzz fuzzy = new RandomFuzz(Environment.TickCount);
+
+    protected StatefulServiceReplicaFactoryTest()
```
olegsych, line 27: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Keep the base test class constructor `private`, not `protected` — nested classes can still call it, and it discourages unintended inheritance. See `test.instructions.md`.

test/Services/Runtime/StatefulServiceReplicaFactoryTest.cs
```diff
@@ -0,0 +1,134 @@
...
+            service = new Mock<StatefulServiceBase>(
+                fuzzy.StatefulServiceContext(),
+                Mock.Of<IStateProviderReplica>())
+            {
+                DefaultValue = DefaultValue.Mock,
+            };
+
+            Mock.Get(serviceFactory)
+                .Setup(f => f(It.IsAny<StatefulServiceContext>()))
+                .Callback((StatefulServiceContext ctx) => actualServiceContext = ctx)
```
olegsych, line 79: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Use a discard for this lambda parameter since the name isn't meaningful: `_ => _(It.IsAny<StatefulServiceContext>())`. Same applies to the `c => c.Dispose()` verify in the `Dispose` test below. See `moq.instructions.md` ("Use discards to reduce noise"). Keep the meaningful `ctx => actualServiceContext = ctx` callback as-is.

test/Services/Runtime/StatefulServiceReplicaFactoryTest.cs
```diff
@@ -0,0 +1,134 @@
...
+            Assert.Equal(replicaId, actualServiceContext.ReplicaId);
+        }
+
+        [Fact]
+        public void ReturnsStatefulServiceReplica()
+        {
+            IStatefulServiceReplica result = sut.CreateReplica(serviceTypeName, serviceName, initializationData, partitionId, replicaId);
+
+            var adapter = Assert.IsType<StatefulServiceReplicaAdapter>(result);
+            Assert.Same(service.Object, adapter.Field<IStatefulUserServiceReplica>().Value);
```
olegsych, line 103: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Don't use `Assert.IsType<T>()` to cast before further assertions. `test.instructions.md`: "Don't use `Assert.IsAssignableFrom<T>()` or `Assert.IsType<T>()` ... Use type casting instead", e.g. `var adapter = (StatefulServiceReplicaAdapter)result;`.

test/Services/Runtime/StatelessServiceInstanceFactoryTest.cs
```diff
@@ -0,0 +1,128 @@
...
+            Assert.Equal(instanceId, actualServiceContext.InstanceId);
+        }
+
+        [Fact]
+        public void ReturnsStatelessServiceInstance()
+        {
+            IStatelessServiceInstance result = sut.CreateInstance(serviceTypeName, serviceName, initializationData, partitionId, instanceId);
+
+            var adapter = Assert.IsType<StatelessServiceInstanceAdapter>(result);
+            Assert.Same(service.Object, adapter.Field<IStatelessUserServiceInstance>().Value);
```
olegsych, line 97: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Don't use `Assert.IsType<T>()` to cast before further assertions. `test.instructions.md`: "Don't use `Assert.IsAssignableFrom<T>()` or `Assert.IsType<T>()` ... Use type casting instead", e.g. `var adapter = (StatelessServiceInstanceAdapter)result;`.

test/Services/Runtime/StatelessServiceInstanceFactoryTest.cs
```diff
@@ -0,0 +1,128 @@
...
+    readonly IStatelessServiceFactory sut;
+
+    // Constructor parameters
+    readonly RuntimeContext runtimeContext;
+    readonly Func<StatelessServiceContext, StatelessService> serviceFactory;
+
+    // Test fixture
+    static readonly IFuzz fuzzy = new RandomFuzz(Environment.TickCount);
+
+    protected StatelessServiceInstanceFactoryTest()
```
olegsych, line 26: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Keep the base test class constructor `private`, not `protected` — nested classes can still call it, and it discourages unintended inheritance. See `test.instructions.md`.

test/Services/Runtime/StatelessServiceInstanceFactoryTest.cs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    public sealed class Dispose : StatelessServiceInstanceFactoryTest
+    {
+        [Fact]
+        public void DisposesRuntimeContext()
+        {
+            var codePackageContext = new Mock<ICodePackageActivationContext>();
+            var runtimeContext = new RuntimeContext();
+            runtimeContext.Field<ICodePackageActivationContext>().Set(codePackageContext.Object);
```
olegsych, line 120: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") This local `runtimeContext` (and the local `sut` below) shadow the base class fields of the same name, which hurts readability. Rename the locals to make the distinction clear — this is the residue of an earlier resolved shadowing thread.

test/Services/Runtime/StatelessServiceInstanceFactoryTest.cs
```diff
@@ -0,0 +1,128 @@
...
+
+public abstract class StatelessServiceInstanceFactoryTest
+{
+    readonly IStatelessServiceFactory sut;
+
+    // Constructor parameters
+    readonly RuntimeContext runtimeContext;
+    readonly Func<StatelessServiceContext, StatelessService> serviceFactory;
+
+    // Test fixture
```
olegsych, line 23: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [💡](# "Consider changing") `test.instructions.md` prescribes only `// Constructor parameters` and `// Method parameters` section comments, and its example shows `fuzzy` with no section comment. Consider dropping this `// Test fixture` label for consistency.

test/Services/Runtime/StatelessServiceInstanceFactoryTest.cs
```diff
@@ -0,0 +1,128 @@
...
+        readonly Mock<StatelessService> service;
+        StatelessServiceContext actualServiceContext;
+
+        public CreateInstance()
+        {
+            service = new Mock<StatelessService>(fuzzy.StatelessServiceContext()) { DefaultValue = DefaultValue.Mock };
+
+            Mock.Get(serviceFactory)
+                .Setup(f => f(It.IsAny<StatelessServiceContext>()))
+                .Callback((StatelessServiceContext ctx) => actualServiceContext = ctx)
```
olegsych, line 73: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Use a discard for this lambda parameter since the name isn't meaningful: `_ => _(It.IsAny<StatelessServiceContext>())`. Same applies to the `c => c.Dispose()` verify in the `Dispose` test below. See `moq.instructions.md` ("Use discards to reduce noise"). Keep the meaningful `ctx => actualServiceContext = ctx` callback as-is.

olegsych (changes requested):
[:copilot:](https://docs.github.com/copilot/responsible-use/code-review) **⚠️ Needs Changes** — The core change is correct, well-scoped, and properly motivated: the factories now fail fast with a clear `InvalidOperationException` instead of a `NullReferenceException` when a user-supplied factory delegate returns null, and constructor arguments are validated. CI is green. However, there is one merge-blocking test gap (the diagnostic message the PR introduces is never asserted), plus several consistency issues with the repo's C#, test, and Moq instructions — all left as inline comments.

### Issues

#### ⚠️ Should Fix — PR description documents a change that wasn't made
The description says *"RuntimeContext.Dispose() made virtual to allow mocking in tests."* That change is not in this PR — `RuntimeContext.cs` is unchanged and `Dispose()` is still non-virtual; disposal is now verified via a mocked `CodePackageContext`. Per `CONTRIBUTING.md` (describe *what* changed and *why*), please remove that stale line. The useful "why" from your review reply — a customer factory returned null and the resulting `NullReferenceException` wasn't diagnostic enough — would be a great addition to the body.

@plave0 please take a look at the inline comments, especially the ❗ must-fix on asserting the exception message.

test/Services/Runtime/StatefulServiceReplicaFactoryTest.cs
olegsych: Optional.

test/Services/Runtime/StatefulServiceReplicaFactoryTest.cs
olegsych: This doesn't sound right.

test/Services/Runtime/StatelessServiceInstanceFactoryTest.cs
olegsych: Optional

olegsych:
I've completed another pass @plave0 and marked 👍 those comments that look legit. Could you run the `/pr-review` skill from the latest `user/olegsych/instructions` before the next review?

PR #538, opened 2026-03-03
Remove ReplicaId from System Dimensions

ilijagrbic:
Remove ReplicaId from the default system dimensions that are emitted. 
ReplicaId has a high cardinality and could cause cardinality explosion.

### DIFF
--- a/test/Diagnostics/Metrics/Implementation/MeterProviderTest.cs
+++ b/test/Diagnostics/Metrics/Implementation/MeterProviderTest.cs
@@ -39,19 +39,17 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
                 var actualSystemDimensionNames = (string[])sut.Private().Field<IEnumerable<string>>().Value;
                 var actualSystemDimensionValues = (string[])sut.Protected().Field<IEnumerable<string>>().Value;
 
-                Assert.Equal(serviceContext.ReplicaOrInstanceId.ToString(), actualSystemDimensionValues[0]);
-                Assert.Equal(serviceContext.PartitionId.ToString(), actualSystemDimensionValues[1]);
-                Assert.Equal(serviceContext.ServiceTypeName, actualSystemDimensionValues[2]);
-                Assert.Equal(serviceContext.ServiceName.ToString(), actualSystemDimensionValues[3]);
-                Assert.Equal(serviceContext.CodePackageActivationContext.ApplicationName, actualSystemDimensionValues[4]);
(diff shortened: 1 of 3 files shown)


test/Diagnostics/Metrics/Implementation/MeterProviderTest.cs
```diff
@@ -39,19 +39,17 @@ public void SetsRequiredDimensionsFromServiceContext()
...
-                Assert.Equal(serviceContext.CodePackageActivationContext.ApplicationName, actualSystemDimensionValues[4]);
-                Assert.Equal(serviceContext.CodePackageActivationContext.ApplicationTypeName, actualSystemDimensionValues[5]);
-
-                Assert.Equal(nameof(ServiceContext.ReplicaOrInstanceId), actualSystemDimensionNames[0]);
-                Assert.Equal(nameof(ServiceContext.PartitionId), actualSystemDimensionNames[1]);
-                Assert.Equal(nameof(ServiceContext.ServiceTypeName), actualSystemDimensionNames[2]);
-                Assert.Equal(nameof(ServiceContext.ServiceName), actualSystemDimensionNames[3]);
-                Assert.Equal(nameof(ServiceContext.CodePackageActivationContext.ApplicationName), actualSystemDimensionNames[4]);
-                Assert.Equal(nameof(ServiceContext.CodePackageActivationContext.ApplicationTypeName), actualSystemDimensionNames[5]);
+                Assert.Equal(serviceContext.PartitionId.ToString(), actualSystemDimensionValues[0]);
```
olegsych, line 42: Ugh, I was wondering why so many lines changed. Consider using Assert.Equals overload taking arrays or IEnumerable<T>.

https://api.xunit.net/v3/3.0.1/v3.3.0.1-Xunit.Assert.Equal.html

olegsych (approved):

plave0 (commented):
I wonder how these changes will affect the readability of the metrics.
Will we still be able to correlate to the instance or replica from which they are emitted?

ilijagrbic:
> I wonder how these changes will affect the readability of the metrics. Will we still be able to correlate to the instance or replica from which they are emitted?

We will still have **PartitionId** + **NodeName**, which would help in this regard. I'm not sure if there exists a placement constraint that would allow multiple replicas for one partition to be on the same node, but this combination should help us narrow down who emitted the metric.

edit: I would argue that we want to avoid exploding cardinality at all costs, so even if we could not fully correlate metric emission to an instance/replica, we should still remove this system dimension.

plave0 (approved):

PR #539, opened 2026-03-06
Clean up Metrics API code

ilijagrbic:
Refactored couple of things to better follow clean code practices and make the code testable
- Remove conversion from `IEnumerable` -> `Array` and use `IReadOnlyCollection` instead
- Introduce `Meter1D/2D/3D`, use them to implement `Int64/TimeSpanMeter` clases
- Remove `recordAction` from `Meter`, instead call `RecordViaNative` directly

Various tests improvements:
- Capture passed `strings` to the interop with Util.CaptureStringPointers in the tests and verify that they are as expected  (as oposed to the old way of testing where we verified that a private method is called on a `Meter` object
- Shorten some of the test classes

Bump package version `12.0.0.beta-25` -> `12.0.0.beta-26`

### DIFF
--- a/src/Diagnostics/Metrics/Implementation/Int64Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Int64Meter.cs
@@ -11,7 +11,7 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
     {
         internal Int64Meter(IFabricMeter fabricMeter, IEnumerable<string> systemDimensionValues) : base(fabricMeter, systemDimensionValues) { }
 
-        public void Record(long value)
+        public new void Record(long value)
         {
             base.Record(value);
         }
--- a/src/Diagnostics/Metrics/Implementation/Int64Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Int64Meter.cs
@@ -11,7 +11,7 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
     {
         internal Int64Meter(IFabricMeter fabricMeter, IReadOnlyCollection<string> systemDimensionValues) : base(fabricMeter, systemDimensionValues) { }
 
-        public new void Record(long value)
+        void IMeter<long>.Record(long value)
         {
             base.Record(value);
         }
--- a/src/Diagnostics/Metrics/Implementation/Int64Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Int64Meter.cs
@@ -9,7 +9,7 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 {
     sealed class Int64Meter : Meter, IMeter<long>
     {
-        internal Int64Meter(IFabricMeter fabricMeter, IEnumerable<string> systemDimensionValues) : base(fabricMeter, systemDimensionValues) { }
+        internal Int64Meter(IFabricMeter fabricMeter, IReadOnlyCollection<string> systemDimensionValues) : base(fabricMeter, systemDimensionValues) { }
 
         public new void Record(long value)
         {
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -33,7 +33,7 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
             return (long)Math.Round(value.TotalMilliseconds);
         }
 
-        protected void Record(long value, int customDimensionCount = 0, string customDimension1 = null, string customDimension2 = null, string customDimension3 = null)
+        protected void RecordOld(long value, int customDimensionCount = 0, string customDimension1 = null, string customDimension2 = null, string customDimension3 = null)
         {
             if (customDimensionCount < 0 || customDimensionCount > 3)
             {
@@ -53,6 +53,11 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
             }
 
             recordAction.Invoke(value, customDimensionCount, customDimension1, customDimension2, customDimension3);
+        }
+
+        protected void Record(long value)
+        {
+            recordAction.Invoke(value, 0, null, null, null);
         }
 
         unsafe private void RecordViaNative(long value, int customDimensionCount, string customDimension1, string customDimension2, string customDimension3)
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -67,23 +67,22 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
                 for (int i = 0; i < systemDimensionValues.Length; i++)
                 {
                     allDimensionPins[i] = GCHandle.Alloc(systemDimensionValues[i], GCHandleType.Pinned);
-                    allDimensionValuesPointers[i] = allDimensionPins[i].AddrOfPinnedObject();
                 }
 
                 if (customDimensionCount > 0)
-                {
                     allDimensionPins[systemDimensionValues.Length] = GCHandle.Alloc(customDimension1, GCHandleType.Pinned);
-                    allDimensionValuesPointers[systemDimensionValues.Length] = allDimensionPins[systemDimensionValues.Length].AddrOfPinnedObject();
-                }
+
                 if (customDimensionCount > 1)
-                {
                     allDimensionPins[systemDimensionValues.Length + 1] = GCHandle.Alloc(customDimension2, GCHandleType.Pinned);
-                    allDimensionValuesPointers[systemDimensionValues.Length + 1] = allDimensionPins[systemDimensionValues.Length + 1].AddrOfPinnedObject();
-                }
+
                 if (customDimensionCount > 2)
-                {
                     allDimensionPins[systemDimensionValues.Length + 2] = GCHandle.Alloc(customDimension3, GCHandleType.Pinned);
-                    allDimensionValuesPointers[systemDimensionValues.Length + 2] = allDimensionPins[systemDimensionValues.Length + 2].AddrOfPinnedObject();
+
+
+                for (int i = 0; i < totalDimensionCount; i++)
+                {
+                    // for strings, AddrOfPinnedObject() returns a pointer to the first character - https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.gchandle.addrofpinnedobject
+                    allDimensionValuesPointers[i] = allDimensionPins[i].AddrOfPinnedObject();
                 }
 
                 fabricMeter.Record(value, (uint)totalDimensionCount, (IntPtr)allDimensionValuesPointers);
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -19,10 +19,7 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 
         internal Meter(IFabricMeter fabricMeter, IEnumerable<string> systemDimensionValues)
         {
-            if (systemDimensionValues == null)
-            {
-                throw new ArgumentNullException(nameof(systemDimensionValues));
-            }
+            _ = systemDimensionValues ?? throw new ArgumentNullException(nameof(systemDimensionValues));
             this.fabricMeter = fabricMeter ?? throw new ArgumentNullException(nameof(fabricMeter));
             this.systemDimensionValues = systemDimensionValues.ToArray();
             this.recordAction = RecordViaNative;
@@ -33,7 +30,7 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
             return (long)Math.Round(value.TotalMilliseconds);
         }
 
-        protected void RecordOld(long value, int customDimensionCount = 0, string customDimension1 = null, string customDimension2 = null, string customDimension3 = null)
+        public void RecordOld(long value, int customDimensionCount = 0, string customDimension1 = null, string customDimension2 = null, string customDimension3 = null)
         {
             if (customDimensionCount < 0 || customDimensionCount > 3)
             {
@@ -60,8 +57,13 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
             recordAction.Invoke(value, 0, null, null, null);
         }
 
-        unsafe private void RecordViaNative(long value, int customDimensionCount, string customDimension1, string customDimension2, string customDimension3)
+        unsafe protected void RecordViaNative(long value, int customDimensionCount, string customDimension1, string customDimension2, string customDimension3)
         {
+            if (customDimensionCount < 0 || customDimensionCount > 3)
+            {
+                throw new ArgumentOutOfRangeException(nameof(customDimensionCount));
+            }
+
             int totalDimensionCount = systemDimensionValues.Length + customDimensionCount;
 
             GCHandle* allDimensionPins = stackalloc GCHandle[totalDimensionCount];
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -88,6 +88,10 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 
                 fabricMeter.Record(value, (uint)totalDimensionCount, (IntPtr)allDimensionValuesPointers);
             }
+            catch (Exception)
+            {
+                throw;
+            }
             finally
             {
                 for (int i = 0; i < totalDimensionCount; i++)
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -12,14 +12,17 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 {
     abstract class Meter
     {
-        protected readonly string[] systemDimensionValues;
         protected readonly IFabricMeter fabricMeter;
+
+        protected readonly IEnumerable<string> systemDimensionValues;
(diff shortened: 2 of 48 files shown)


test/Diagnostics/Metrics/Implementation/Meter1DTest.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            [Fact]
+            public void ThrowsArgumentNullExceptionWhenSystemDimensionsAreNull()
+            {
+                var exception = Xunit.Record.Exception(() => new Mock<Meter1D>(fabricMeter, null).Object);
+                Assert.IsType<ArgumentNullException>(exception.InnerException);
+            }
+        }
+
+        public class Record : Meter1DTest
```
plave0, line 41: This is an interesting way to test a protected method.
My approach would have been to derive a class from `Meter1D` which exposes `Record`. This would introduce some boilerplate setup code, but the tests would be more readable. 

What is the benefit of testing the protected method directly like in this test class?

test/Diagnostics/Metrics/Implementation/Meter1DTest.cs
```diff
@@ -0,0 +1,79 @@
...
+                sutMethod = sut.Protected().Method<Action<long, string>>();
+
+                // capture strings emitted to IFabricMeter.Record for assertion in tests
+                Mock.Get(fabricMeter)
+                    .Setup(m => m.Record(It.IsAny<long>(), It.IsAny<uint>(), It.IsAny<IntPtr>()))
+                    .Callback<long, uint, IntPtr>((value, count, stringPtrs) => recordedArray = Util.CaptureStringPointers(stringPtrs, count));
+            }
+
+            [Fact]
+            public void CallsRecordAction()
```
plave0: Is it reasonable to align the taming of the test `CallsRecordAction` with other tests where the same test is called `InvokesBaseRecord`?

src/Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -5,81 +5,64 @@
...
         {
-            if (systemDimensionValues == null)
-            {
-                throw new ArgumentNullException(nameof(systemDimensionValues));
-            }
+            _ = systemDimensionValues ?? throw new ArgumentNullException(nameof(systemDimensionValues));
             this.fabricMeter = fabricMeter ?? throw new ArgumentNullException(nameof(fabricMeter));
-            this.systemDimensionValues = systemDimensionValues.ToArray();
-            this.recordAction = RecordViaNative;
+            this.systemDimensionValues = systemDimensionValues;
```
olegsych: Can we combine validation of the `systemDimensionValues` parameter with the assignment of the field?

src/Diagnostics/Metrics/Implementation/Meter.cs
olegsych: I'm not sure if we arrived at any conclusion in our previous discussion, https://github.com/microsoft/service-fabric-dotnet/pull/533/changes#r2878969327. I still believe this catch serves no purpose and should be removed, but if you think otherwise, we should add a comment explaining it.

src/Diagnostics/Metrics/Implementation/Meter1D.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
+{
+    abstract class Meter1D : Meter
+    {
+        internal Meter1D(IFabricMeter fabricMeter, IReadOnlyCollection<string> systemDimensionValues) : base(fabricMeter, systemDimensionValues) { }
+
+        protected void Record(long value, string customDimension1)
+        {
+            _ = customDimension1 ?? throw new ArgumentNullException(nameof(customDimension1));
```
olegsych: This looks like we're over-specifying the expected usage because we're only passing the customDimension1 to the RecordViaNative. What would happen if we pass the null values through?

src/Diagnostics/Metrics/Implementation/Meter1D.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
+{
+    abstract class Meter1D : Meter
+    {
+        internal Meter1D(IFabricMeter fabricMeter, IReadOnlyCollection<string> systemDimensionValues) : base(fabricMeter, systemDimensionValues) { }
+
+        protected void Record(long value, string customDimension1)
```
olegsych: In MeterProvider we specify parameter names as `dimension1Name`, etc. It would be nice to use `dimension1Value` etc for clarity and consistency.

src/Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -5,81 +5,64 @@
...
-                throw new ArgumentException(nameof(customDimension1));
-            }
 
-            recordAction.Invoke(value, customDimensionCount, customDimension1, customDimension2, customDimension3);
-        }
-
-        unsafe private void RecordViaNative(long value, int customDimensionCount, string customDimension1, string customDimension2, string customDimension3)
-        {
-            int totalDimensionCount = systemDimensionValues.Length + customDimensionCount;
+            int totalDimensionCount = systemDimensionValues.Count + customDimensionCount;
```
olegsych: Aside from being inconsistent, it feels that the terms `total` and `all` are redundant in this context. Can we shorten the variable names to `dimensionCount`, `dimensionValueHandles` and `dimensionValuePointers`?

src/Diagnostics/Metrics/Implementation/Int64Meter.cs
```diff
@@ -9,9 +9,9 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 {
     sealed class Int64Meter : Meter, IMeter<long>
     {
-        internal Int64Meter(IFabricMeter fabricMeter, IEnumerable<string> systemDimensionValues) : base(fabricMeter, systemDimensionValues) { }
+        internal Int64Meter(IFabricMeter fabricMeter, IReadOnlyCollection<string> systemDimensionValues) : base(fabricMeter, systemDimensionValues) { }
 
-        public void Record(long value)
+        public new void Record(long value)
```
olegsych: I try to avoid implementing interfaces publicly, particularly in the implementation classes, like Int64Meter, which aren't meant to be used by the consumers directly. Instead, I implement interfaces explicitly by default because

- I think an explicit interface implementation more accurately conveys visibility of the method, and in particular that it's not public.
- In tests, declaring the `sut` variable of the interface type rather than of the class type expresses the design expectation explicitly, i.e. that Record method is not just a class method but that it implements the IMeter.Record.

src/Diagnostics/Metrics/Implementation/MeterProvider.cs
```diff
@@ -12,8 +12,8 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 {
     abstract class MeterProvider<TValueType> : IMeterProvider<TValueType>
     {
-        readonly IEnumerable<string> systemDimensionNames;
-        protected readonly IEnumerable<string> systemDimensionValues;
+        readonly IReadOnlyList<string> systemDimensionNames;
```
olegsych: Why do we need a broader IReadOnlyList<T> rather than the IReadOnlyCollection<T>? I only see us using the Count property which is available through the IReadOnlyCollection<T>

src/Diagnostics/Metrics/Implementation/MeterProvider.cs
```diff
@@ -44,14 +44,14 @@ protected MeterProvider(ServiceContext serviceContext = null)
             }
             else
             {
-                this.systemDimensionNames = Enumerable.Empty<string>();
-                this.systemDimensionValues = Enumerable.Empty<string>();
+                this.systemDimensionNames = Array.Empty<string>();
```
olegsych: This is cool, I remember reading about arrays implicitly converting to generic collection types, but I usually forget about it.

test/Diagnostics/Metrics/Implementation/Int64Meter1DTest.cs
```diff
@@ -41,16 +42,25 @@ public class Record : Int64Meter1DTest
             readonly long value = fuzzy.Int64();
             readonly string customDimension1 = fuzzy.String();
 
-            readonly Action<long, int, string, string, string> mockRecordAction = Mock.Of<Action<long, int, string, string, string>>();
+            protected string[] recordedArray;
```
olegsych: The term array isn't helpful - it is already self-evident from the variable type and doesn't explain its intent. Something like `actualDimensions` would be more better.

test/Diagnostics/Metrics/Implementation/Int64Meter3DTest.cs
```diff
@@ -43,16 +44,25 @@ public class Record : Int64Meter3DTest
...
+                // capture strings emitted to IFabricMeter.Record for assertion in tests
+                Mock.Get(fabricMeter)
+                    .Setup(m => m.Record(It.IsAny<long>(), It.IsAny<uint>(), It.IsAny<IntPtr>()))
+                    .Callback<long, uint, IntPtr>((value, count, stringPtrs) => recordedArray = Util.CaptureStringPointers(stringPtrs, count));
+            }
 
             [Fact]
             public void InvokesBaseRecord()
             {
+                var expectedArray = systemDimensions.Concat(new[] { customDimension1, customDimension2, customDimension3 }).ToArray();
```
olegsych: We should specify the type explicitly because ToArray is far on the right and doesn't tell us what type of items the array stores.

test/Diagnostics/Metrics/Implementation/Meter1DTest.cs
```diff
@@ -0,0 +1,79 @@
...
+using Xunit;
+
+namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
+{
+    public abstract class Meter1DTest
+    {
+        static readonly IFuzz fuzzy = new RandomFuzz(Environment.TickCount);
+
+        readonly IFabricMeter fabricMeter = Mock.Of<IFabricMeter>();
+        readonly List<string> systemDimensions = fuzzy.List(() => fuzzy.String());
```
olegsych, line 22: Can we declare the systemDimensions with the type matching the constructor parameter?

test/Diagnostics/Metrics/Implementation/Meter1DTest.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        public class Constructor : Meter1DTest
+        {
+            [Fact]
+            public void ThrowsArgumentNullExceptionWhenMeterIsNull()
+            {
+                var exception = Xunit.Record.Exception(() => new Mock<Meter1D>(null, systemDimensions).Object);
```
olegsych: Why not Assert.Throws? I didn't know Xunit had this API and don't think I've ever used it. Now that I look at the next line, it looks like it's a workaround for the constructor being protected and Moq wrapping the exception. Does Inspector's `Type<Meter1D>.New(fabricMeter, null)` preserve the original exception?

But regardless, I think it would be better to use Assert.Throws even if we need to unwrap the exception because it reports better failure messages and wouldn't result in an obscure NullReferenceException if the constructor didn't throw as expected.

test/Diagnostics/Metrics/Implementation/Meter1DTest.cs
```diff
@@ -0,0 +1,79 @@
...
+                Assert.IsType<ArgumentNullException>(exception.InnerException);
```
olegsych, line 30: It would be great to validate ArgumentException.ParamName as well.

test/Diagnostics/Metrics/Implementation/Meter3DTest.cs
```diff
@@ -0,0 +1,88 @@
...
+            readonly string customDimension2 = fuzzy.String();
+            readonly string customDimension3 = fuzzy.String();
+
+            public Record()
+            {
+                sut = new Mock<Meter3D>(fabricMeter, systemDimensions).Object;
+                sutMethod = sut.Protected().Method<Action<long, string, string, string>>();
+
+                // capture strings emitted to IFabricMeter.Record for assertion in tests
+                Mock.Get(fabricMeter)
```
olegsych, line 59: There is a fair amount of duplication in these tests, so there are probably additional opportunities to identify new abstractions either in the test code, the product code or both. I have a feeling we might be overusing inheritance to reuse the Meter logic and that composition (i.e. injecting a reusable abstraction instead of extending it) might be better.

test/Diagnostics/Metrics/Implementation/MeterTest.cs
```diff
@@ -16,110 +15,105 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
...
 
         public class Record : MeterTest
         {
-            readonly Meter sut;
+            readonly Method<Action<long>> sutMethod;
+
+            public Record() => sutMethod = sut.Protected().Method<Action<long>>();
+
+            [Fact]
+            public void CallsRecordAction()
```
olegsych: The test name feels like it's simply restating what the act section does. I would try to make the test name more descriptive, along the lines of CallsFabricMeterWithCombinedDimensions.

test/Diagnostics/Metrics/Implementation/MeterTest.cs
```diff
@@ -16,110 +15,105 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
...
+            {
+                var expectedArray = systemDimensions.ToArray();
+
+                sutMethod.Invoke(value);
```
olegsych, line 83: I try to avoid Invoke with delegates it and use the function call syntax, i.e. `sutMethod()` instead of `sutMethod.Invoke()`, because it is shorter and just as expressive. In this case it also helps to see that the field name is not very descriptive.

olegsych (approved):
Thanks @ilijagrbic. I've left a bunch of thoughts in the comments, none of them are blockers, but I feel like we're finally getting to the design aspect of TDD here. We can leave it as is, but if you still have time, I would continue iterating because the duplicated code may be indicating overuse of inheritance. 

This is a very common design problem because people trained in object-oriented design tend to use type hierarchies where an object composition may be a better option. We don't train in object composition nearly as much, but improper use of inheritance may not be immediately obvious in the beginning even when you're aware of this problem, so the apparent duplication is a symptom I try to watch for.

As I mentioned in one of the comments below, I wonder if we should reuse Meter logic through composition rather than inheritance. The TimeSpanMeter classes are probably the easiest to see this problem. Imagine if `TimeSpanMeter` received an `IMeter<long>` as a constructor parameter instead of inheriting from `Meter`. I think this would simplify its implementation and tests, although it's difficult to tell offhand without trying it out. And if the TimeSpanMeter classes don't need to inherit from the Meter classes, perhaps we could have Int64Meter classes inherit directly from Meter, eliminating the Meter*D classees. I'm not sure if the Int64Meter classes should still inherit from the Meter class at that point, perhaps there is potential to elevate it to a concrete abstraction around the IFabricMeter and inject it as well. This is something that would be easier to see once the TimeSpanMeter classes are refactored.

Anyway, I wanted to encourage you to experiment with this design a bit more to see if it could be simpler, in the four rules sense.

Also, I would love for you to try out the new .editorconfig and apply the recommendations to see if they are helpful.

src/Diagnostics/Metrics/Implementation/Meter1D.cs
ilijagrbic: That would cause an undefined behavior mostly likely. We do have tests though, that verify that this call works, and later on appropriate system dimensions are passed down to the interop.

src/Diagnostics/Metrics/Implementation/MeterProvider.cs
ilijagrbic: During first set of changes I wanted to incorporate the IReadOnlyList, before I found out Count if also present in IReadOnlyCollection, but it seems I never followed up in this specific class. I will fix this.

plave0 (approved):
Good improvements overall, I just left a couple of non-blocker comments.

test/Diagnostics/Metrics/Implementation/Meter3DTest.cs
ilijagrbic: It's always a trade off with regards to this. During our previous discussions we identified other issues with the previous which nudged us in the direction of this design. I think we can iterate on the internal design in the future, and I agree on the point you make here, but I would prefer to wrap up the implementation as soon as possible, to meet the deliverables for SF12/11.5.

test/Diagnostics/Metrics/Implementation/Meter1DTest.cs
ilijagrbic: Makes sense, I will update the naming.

test/Diagnostics/Metrics/Implementation/Meter1DTest.cs
ilijagrbic: I guess it comes down to personal preference. A derived class seemed more complex to me, but I understand the argument that usage of reflection can also be less readable. In any case I will keep it as is for now, but we can discuss this more and change it in the future.

test/Diagnostics/Metrics/Implementation/Int64Meter3DTest.cs
ilijagrbic: Makes sense, I will update this.

test/Diagnostics/Metrics/Implementation/Meter1DTest.cs
ilijagrbic: The `Type<Meter1D>.New(fabricMeter, null)` still doesn't work since this is an abstract class and can't be instantiated. But I will use the `Assert.Throws` to capture the exception and verify the inner exception separately.

src/Diagnostics/Metrics/Implementation/Int64Meter.cs
ilijagrbic: Thanks, Oleg, for the thoughts on this. I was thinking this is not as important since the class itself is internal. But regardless, your reasoning makes sense and I will update this.

plave0 (approved):
Reapprove

PR #540, opened 2026-03-06
Cleanup .editorconfig

olegsych:
To encourage succinct, expressive code and make authors think carefully about it's readability and complexity, change the `.editorconfig` to
- Reduce brace noise with file-scoped namespaces and expression-bodied members.
- Make sure `var` usage doesn't obscure variable types when we read/review code without IDE assistance.
- Reduce noise from redundant accessibility modifiers and `this.` qualifiers.
- Make fields read-only.
- camelCase private fields, parameters, locals and PascalCase everything else.
- Place using directives outside of namespace to avoid IDE suggestions that "simplify" namespaces by dropping prefixes and can make them ambiguous to human readers.

Also:
- Remove redundant settings 
- Remove C++, VB settings not applicable to this repo

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -15,49 +15,40 @@ indent_size = 4
 insert_final_newline = true
 charset = utf-8
 
-# Xml build files
-[*.builds]
-indent_size = 2
-
 # Xml files
-[*.{props,targets,ruleset,config,nuspec,resx,vsixmanifest,vsct,xml}]
-indent_size = 2
-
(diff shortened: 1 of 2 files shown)


ilijagrbic (approved):
Great changes @olegsych . I like the switch from suggestion to warning in particular for many code standards. This way we can detect discrepancies from the coding standard even outside of the immediate code we are working on and will be encouraged to remove the warnings much more.

PR #544, opened 2026-03-11
Enable deterministic dispose for IMeter and IMeterProvider 

ilijagrbic:
Enable deterministic dispose of metric related COM objects:
- `IMeter/1D/2D/3D` and `IMeterProvider` now inherit `IDisposable`
- Implement `Dispose()` method in `Meter`/`MeterProvider` classes that releases the COM object
- Throw an exception if `Meter`/`MeterProvider` are disposed while calling `Record`/`CreateMeter`
- Add `MarshalUsing(typeof(UniqueComInterfaceMarshaller<IFabricMeter>))` to the methods that get COM objects in order for `FinalReleaseComObject` to work on .NETCore.

Integrate the changes in Actors and Remoting:
- Dispose `AggregatedDiagnostics` in `OnCloseAsync`
- Dispose `MeterProvider` and `IDiagnostics` in classes where they are created (`FabricTransportServiceRemotingListener` and `FabricTransportMessageHandler`)

Other:
- Various opportunistic coding standard refactors, in line with new .editorconfig
- IFabricTransportMessageHandler now inherits from IDisposable, to make FabricTransportMessageHandler implement IDisposable exsplicitly
- Various changes to enable unit testability of the changeset, such as introduction of `IFabricTransportListener` for `FabricTransportListener`

### DIFF
--- a/src/Diagnostics/Metrics/Implementation/IFabricMeter.cs
+++ b/src/Diagnostics/Metrics/Implementation/IFabricMeter.cs
@@ -5,10 +5,9 @@
 
 using System;
 using System.Runtime.InteropServices;
-
-#if NET
 using System.Runtime.InteropServices.Marshalling;
-#else
+
+#if !NET
 using GeneratedComInterfaceAttribute = System.Runtime.InteropServices.ComImportAttribute;
 #endif
 
--- a/src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
+++ b/src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
@@ -5,10 +5,9 @@
 
 using System;
 using System.Runtime.InteropServices;
-
-#if NET
 using System.Runtime.InteropServices.Marshalling;
-#else
+
+#if !NET
 using GeneratedComInterfaceAttribute = System.Runtime.InteropServices.ComImportAttribute;
 #endif
 
@@ -19,6 +18,7 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
     [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
     partial interface IFabricMeterProvider
     {
+        [return: MarshalUsing(typeof(UniqueComInterfaceMarshaller<IFabricMeter>))]
         IFabricMeter CreateMeter([MarshalAs(UnmanagedType.LPWStr)] string metricNamespace, [MarshalAs(UnmanagedType.LPWStr)] string name, uint count, [MarshalAs(UnmanagedType.LPArray, ArraySubType = UnmanagedType.LPWStr)] string[] dimensionNames);
     }
 }
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -38,7 +38,7 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 
         protected long ConvertTimeSpanToLong(TimeSpan value) => (long)Math.Round(value.TotalMilliseconds);
 
-        unsafe protected void RecordViaNative(long value, int customDimensionCount, string dimension1Value, string dimension2Value, string dimension3Value)
+        protected unsafe void RecordViaNative(long value, int customDimensionCount, string dimension1Value, string dimension2Value, string dimension3Value)
         {
             if (disposed)
                 throw new ObjectDisposedException(nameof(Meter));
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -19,7 +19,6 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
         static Func<object, int> finalReleaseComObject = Utility.FinalReleaseComObject;
         bool disposed = false;
 
-
         internal Meter(IFabricMeter fabricMeter, IReadOnlyCollection<string> systemDimensionValues)
         {
             this.systemDimensionValues = systemDimensionValues ?? throw new ArgumentNullException(nameof(systemDimensionValues));
@@ -42,7 +41,7 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
         unsafe protected void RecordViaNative(long value, int customDimensionCount, string dimension1Value, string dimension2Value, string dimension3Value)
         {
             if (disposed)
-                throw new ObjectDisposedException(nameof(this.GetType));
+                throw new ObjectDisposedException(nameof(Meter));
             if (customDimensionCount < 0 || customDimensionCount > 3)
                 throw new ArgumentOutOfRangeException(nameof(customDimensionCount));
 
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -27,7 +27,7 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 
         public void Dispose()
         {
-            if (fabricMeter != null)
+            if (!disposed)
             {
                 finalReleaseComObject(fabricMeter);
                 disposed = true;
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -5,15 +5,20 @@
 
 using System;
 using System.Collections.Generic;
+using System.Fabric.Interop;
 using System.Runtime.InteropServices;
 
 namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 {
     abstract class Meter : IDisposable
     {
-        protected readonly IFabricMeter fabricMeter;
-
         protected readonly IReadOnlyCollection<string> systemDimensionValues;
+
+        readonly IFabricMeter fabricMeter;
+
+        static Func<object, int> finalReleaseComObject = Utility.FinalReleaseComObject;
+        bool disposed = false;
+
 
         internal Meter(IFabricMeter fabricMeter, IReadOnlyCollection<string> systemDimensionValues)
         {
@@ -21,30 +26,25 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
             this.fabricMeter = fabricMeter ?? throw new ArgumentNullException(nameof(fabricMeter));
         }
 
-        /// <summary>
-        /// Releases the native COM resources held by this meter.
-        /// </summary>
         public void Dispose()
         {
-            // TODO: Release the fabricMeter COM object
+            if (fabricMeter != null)
+            {
+                finalReleaseComObject(fabricMeter);
+                disposed = true;
+            }
         }
 
-        protected void Record(long value)
-        {
-            RecordViaNative(value, 0, null, null, null);
-        }
+        protected void Record(long value) => RecordViaNative(value, 0, null, null, null);
 
-        protected long ConvertTimeSpanToLong(TimeSpan value)
-        {
-            return (long)Math.Round(value.TotalMilliseconds);
-        }
+        protected long ConvertTimeSpanToLong(TimeSpan value) => (long)Math.Round(value.TotalMilliseconds);
 
         unsafe protected void RecordViaNative(long value, int customDimensionCount, string dimension1Value, string dimension2Value, string dimension3Value)
         {
+            if (disposed)
+                throw new ObjectDisposedException(nameof(this.GetType));
             if (customDimensionCount < 0 || customDimensionCount > 3)
-            {
                 throw new ArgumentOutOfRangeException(nameof(customDimensionCount));
-            }
 
             int dimensionCount = systemDimensionValues.Count + customDimensionCount;
 
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -9,7 +9,7 @@ using System.Runtime.InteropServices;
 
 namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 {
-    abstract class Meter
+    abstract class Meter : IDisposable
     {
         protected readonly IFabricMeter fabricMeter;
 
@@ -19,6 +19,14 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
         {
             this.systemDimensionValues = systemDimensionValues ?? throw new ArgumentNullException(nameof(systemDimensionValues));
             this.fabricMeter = fabricMeter ?? throw new ArgumentNullException(nameof(fabricMeter));
+        }
+
+        /// <summary>
+        /// Releases the native COM resources held by this meter.
+        /// </summary>
+        public void Dispose()
+        {
+            // TODO: Release the fabricMeter COM object
         }
 
         protected void Record(long value)
--- a/src/Diagnostics/Metrics/Implementation/Meter.cs
+++ b/src/Diagnostics/Metrics/Implementation/Meter.cs
@@ -14,10 +14,9 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
     {
         protected readonly IReadOnlyCollection<string> systemDimensionValues;
 
-        readonly IFabricMeter fabricMeter;
+        IFabricMeter fabricMeter;
 
         static Func<object, int> finalReleaseComObject = Utility.FinalReleaseComObject;
-        bool disposed = false;
 
         internal Meter(IFabricMeter fabricMeter, IReadOnlyCollection<string> systemDimensionValues)
         {
@@ -27,10 +26,10 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 
         public void Dispose()
         {
-            if (!disposed)
+            if (!IsDisposed())
             {
                 finalReleaseComObject(fabricMeter);
-                disposed = true;
+                fabricMeter = null;
             }
         }
 
@@ -38,9 +37,11 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 
         protected long ConvertTimeSpanToLong(TimeSpan value) => (long)Math.Round(value.TotalMilliseconds);
 
+        bool IsDisposed() => fabricMeter == null;
+
         protected unsafe void RecordViaNative(long value, int customDimensionCount, string dimension1Value, string dimension2Value, string dimension3Value)
         {
-            if (disposed)
+            if (IsDisposed())
                 throw new ObjectDisposedException(nameof(Meter));
             if (customDimensionCount < 0 || customDimensionCount > 3)
                 throw new ArgumentOutOfRangeException(nameof(customDimensionCount));
--- a/src/Diagnostics/Metrics/Implementation/MeterProvider.cs
+++ b/src/Diagnostics/Metrics/Implementation/MeterProvider.cs
@@ -60,6 +60,14 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
             return fabricMeterProvider.CreateMeter(metricNamespace, metricName, (uint)allDimensions.Length, allDimensions);
         }
 
+        /// <summary>
+        /// Releases the native COM resources held by this meter provider.
+        /// </summary>
+        public void Dispose()
+        {
+            // TODO: Release the fabricMeterProvider COM object
+        }
+
         public abstract IMeter<TValueType> CreateMeter(string metricNamespace, string name);
         public abstract IMeter1D<TValueType> CreateMeter(string metricNamespace, string name, string dimension1Name);
         public abstract IMeter2D<TValueType> CreateMeter(string metricNamespace, string name, string dimension1Name, string dimension2Name);
--- a/src/Diagnostics/Metrics/Implementation/MeterProvider.cs
+++ b/src/Diagnostics/Metrics/Implementation/MeterProvider.cs
@@ -69,7 +69,7 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 
         public void Dispose()
         {
-            if (fabricMeterProvider != null)
+            if (!disposed)
             {
                 finalReleaseComObject(fabricMeterProvider);
                 disposed = true;
--- a/src/Diagnostics/Metrics/Implementation/MeterProvider.cs
+++ b/src/Diagnostics/Metrics/Implementation/MeterProvider.cs
@@ -6,6 +6,7 @@
 using System;
 using System.Collections.Generic;
 using System.Fabric;
+using System.Fabric.Interop;
 using System.Linq;
 
 namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
@@ -14,9 +15,12 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
     {
         readonly IReadOnlyCollection<string> systemDimensionNames;
         protected readonly IReadOnlyCollection<string> systemDimensionValues;
-        protected readonly IFabricMeterProvider fabricMeterProvider;
+        readonly IFabricMeterProvider fabricMeterProvider;
+
+        bool disposed = false;
 
         static Func<IFabricMeterProvider> createFabricMeterProvider = NativeTelemetry.FabricCreateMeterProvider;
+        static Func<object, int> finalReleaseComObject = Utility.FinalReleaseComObject;
 
         protected MeterProvider(ServiceContext serviceContext = null)
         {
@@ -51,6 +55,9 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 
         protected IFabricMeter CreateNativeMeter(string metricNamespace, string metricName, IEnumerable<string> additionalDimensions)
         {
+            if (disposed)
+                throw new ObjectDisposedException(nameof(this.GetType));
+
             var allDimensionsList = new List<string>(systemDimensionNames.Count + additionalDimensions.Count());
 
             allDimensionsList.AddRange(systemDimensionNames);
@@ -60,12 +67,13 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
             return fabricMeterProvider.CreateMeter(metricNamespace, metricName, (uint)allDimensions.Length, allDimensions);
         }
 
-        /// <summary>
-        /// Releases the native COM resources held by this meter provider.
-        /// </summary>
         public void Dispose()
         {
-            // TODO: Release the fabricMeterProvider COM object
+            if (fabricMeterProvider != null)
+            {
+                finalReleaseComObject(fabricMeterProvider);
+                disposed = true;
+            }
         }
 
         public abstract IMeter<TValueType> CreateMeter(string metricNamespace, string name);
--- a/src/Diagnostics/Metrics/Implementation/MeterProvider.cs
+++ b/src/Diagnostics/Metrics/Implementation/MeterProvider.cs
@@ -56,7 +56,7 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
         protected IFabricMeter CreateNativeMeter(string metricNamespace, string metricName, IEnumerable<string> additionalDimensions)
         {
             if (disposed)
-                throw new ObjectDisposedException(nameof(this.GetType));
+                throw new ObjectDisposedException(nameof(MeterProvider<TValueType>));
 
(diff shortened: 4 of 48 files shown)


test/Diagnostics/Metrics/Implementation/MeterTest.cs
```diff
@@ -87,33 +89,38 @@ public void CallsFabricMeterWithCombinedDimensions()
...
         }
 
-        public class RecordViaNative : MeterTest
+        public class DisposeTest : MeterTest, IDisposable
+        {
+            public DisposeTest() => typeof(Meter).Field<Func<object, int>>().Set(objectParam => fuzzy.Int32());
+            public void Dispose() => typeof(Meter).Field<Func<object, int>>().Set(Utility.FinalReleaseComObject);
+
+            [Fact]
+            public void DisposesNativeFabricMeterProvider()
```
ilijagrbic: NOTE: Refactor and verify FinalReleaseComObject call

test/Diagnostics/Metrics/Implementation/MeterProviderTest.cs
```diff
@@ -55,52 +55,50 @@ public void SetsRequiredDimensionsFromServiceContext()
...
         }
 
-        public class Class : MeterProviderTest
+        public class DisposeTest : MeterProviderTest
         {
+            public DisposeTest() => typeof(MeterProvider<int>).Field<Func<object, int>>().Set(objectParam => fuzzy.Int32());
+            override public void Dispose()
+            {
+                base.Dispose();
+                typeof(MeterProvider<int>).Field<Func<object, int>>().Set(Utility.FinalReleaseComObject);
```
ilijagrbic, line 83: NOTE: Refactor and verify FinalReleaseComObject call

src/Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -5,38 +5,45 @@
...
         {
             this.systemDimensionValues = systemDimensionValues ?? throw new ArgumentNullException(nameof(systemDimensionValues));
             this.fabricMeter = fabricMeter ?? throw new ArgumentNullException(nameof(fabricMeter));
         }
 
-        protected void Record(long value)
+        public void Dispose()
         {
-            RecordViaNative(value, 0, null, null, null);
+            if (!disposed)
```
ilijagrbic: NOTE: add tests that verify idempotency of this method.

src/Diagnostics/Metrics/Implementation/MeterProvider.cs
```diff
@@ -60,6 +67,15 @@ protected IFabricMeter CreateNativeMeter(string metricNamespace, string metricNa
             return fabricMeterProvider.CreateMeter(metricNamespace, metricName, (uint)allDimensions.Length, allDimensions);
         }
 
+        public void Dispose()
+        {
+            if (!disposed)
```
ilijagrbic: NOTE: add tests that verify idempotency of this method.

src/Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -5,38 +5,46 @@
...
-        protected void Record(long value)
+        public void Dispose()
         {
-            RecordViaNative(value, 0, null, null, null);
+            if (fabricMeter != null)
+            {
+                finalReleaseComObject(fabricMeter);
+                disposed = true;
+            }
         }
```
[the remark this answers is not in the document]
olegsych: I think this is fair. Ability to safely call Dispose multiple times for the same object is what I prefer as well because when Dispose is invoked after another exception, any exceptions thrown by it would destroy the original. However, I'm not sure copilot @copilot is correct about double-releasing the COM object throwing InvalidComObjectException. We should test and find out how it works on .NET Framework and .NET Core.

I was also thinking that checking the fabricMeter for null is redundant because it should never be null thanks to the constructor check.

Ideally Dispose should be a single line unconditionally invoking the finalReleaseComObject, but if we find out that the same instance cannot be safely released multiple times, I would make the fabricMeter writeable and use a null check instead of the additional disposed field.

src/Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -5,38 +5,46 @@
...
 
-        protected long ConvertTimeSpanToLong(TimeSpan value)
-        {
-            return (long)Math.Round(value.TotalMilliseconds);
-        }
+        protected void Record(long value) => RecordViaNative(value, 0, null, null, null);
+
+        protected long ConvertTimeSpanToLong(TimeSpan value) => (long)Math.Round(value.TotalMilliseconds);
 
         unsafe protected void RecordViaNative(long value, int customDimensionCount, string dimension1Value, string dimension2Value, string dimension3Value)
```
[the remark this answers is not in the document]
olegsych: I agree. We should add this to the .editorconfig.

src/Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -5,38 +5,46 @@
...
         {
+            if (disposed)
+                throw new ObjectDisposedException(nameof(this.GetType));
```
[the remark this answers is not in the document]
olegsych: Good catch. I don't think we need to null-check the FullName because it shouldn't be null for concrete types.

src/FabricTransport/Runtime/IFabricTransportListener.cs
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.ServiceFabric.FabricTransport.Runtime
+{
+    interface IFabricTransportListener : IDisposable
```
olegsych, line 12: This is not directly related to the deterministic release of the COM objects, correct?

src/FabricTransport/Runtime/IFabricTransportMessageHandler.cs
```diff
@@ -11,7 +12,7 @@ namespace Microsoft.ServiceFabric.FabricTransport.Runtime
     /// Defines the interface that must be implemented by the ServiceRemotingListener to receive messages from the
     /// remoting transport.
     /// </summary>
-    internal interface IFabricTransportMessageHandler
+    interface IFabricTransportMessageHandler : IDisposable
```
olegsych, line 15: The FabricTransportMessageHandler is already IDisposable. I assume that making the interface disposable as well helps us make sure that the entire chain of the objects is disposable. Is that correct, could you call this out in the description?

test/Actors/Diagnostics/DiagnosticsFactoryTest.cs
```diff
@@ -87,11 +86,18 @@ public class DisposeTest : DiagnosticsFactoryTest
         {
             readonly Guid guid = Guid.NewGuid();
             readonly Mock<PerformanceCounterProviderV2> mockPerformanceCounterProviderV2;
+            readonly Mock<IMeterProvider<TimeSpan>> mockTimeSpanProvider;
+            readonly Mock<IMeterProvider<long>> mockLongProvider;
 
             public DisposeTest()
             {
                 mockPerformanceCounterProviderV2 = new Mock<PerformanceCounterProviderV2>(guid, typeInformation);
+                mockTimeSpanProvider = new Mock<IMeterProvider<TimeSpan>>() { DefaultValue = DefaultValue.Mock };
```
olegsych: The .NET guidelines recommend against Hungarian prefixes. The term mock is not really a Hungarian acronym but it still restates the information already available in the variable type. I think it's redundant here.

test/Diagnostics/Metrics/Implementation/MeterProviderTest.cs
```diff
@@ -6,38 +6,38 @@
...
 
             [Fact]
             public void SetsRequiredDimensionsFromServiceContext()
             {
-                var sut = new TestMeterProvider<int>(serviceContext);
-
-                var actualSystemDimensionNames = (string[])sut.Private().Field<IReadOnlyCollection<string>>().Value;
-                var actualSystemDimensionValues = (string[])sut.Protected().Field<IReadOnlyCollection<string>>().Value;
+                string[] actualSystemDimensionNames = (string[])sut.Private().Field<IReadOnlyCollection<string>>().Value;
+                string[] actualSystemDimensionValues = (string[])sut.Protected().Field<IReadOnlyCollection<string>>().Value;
```
olegsych: Why not var?

test/Diagnostics/Metrics/Implementation/MeterProviderTest.cs
```diff
@@ -55,52 +55,50 @@ public void SetsRequiredDimensionsFromServiceContext()
             [Fact]
             public void ThrowAnArgumentExceptionWhenServiceContextNull()
             {
-                var sut = new TestMeterProvider<int>(null);
+                var sut = new Mock<MeterProvider<int>>(null).Object;
 
-                var actualSystemDimensionNames = sut.Private().Field<IReadOnlyCollection<string>>().Value;
-                var actualSystemDimensionValues = sut.Protected().Field<IReadOnlyCollection<string>>().Value;
+                IReadOnlyCollection<string> actualSystemDimensionNames = sut.Private().Field<IReadOnlyCollection<string>>().Value;
+                IReadOnlyCollection<string> actualSystemDimensionValues = sut.Protected().Field<IReadOnlyCollection<string>>().Value;
```
olegsych: Why not var?

test/Diagnostics/Metrics/Implementation/MeterProviderTest.cs
```diff
@@ -55,52 +55,50 @@ public void SetsRequiredDimensionsFromServiceContext()
...
 
                 Assert.Empty(actualSystemDimensionValues);
                 Assert.Empty(actualSystemDimensionNames);
             }
         }
 
-        public class Class : MeterProviderTest
+        public class DisposeTest : MeterProviderTest
         {
+            public DisposeTest() => typeof(MeterProvider<int>).Field<Func<object, int>>().Set(objectParam => fuzzy.Int32());
```
olegsych: I would put the lambda body on the next line. This looks very dense and probably needs a comment explaining it.

test/Diagnostics/Metrics/Implementation/MeterProviderTest.cs
```diff
@@ -55,52 +55,50 @@ public void SetsRequiredDimensionsFromServiceContext()
...
+            override public void Dispose()
+            {
+                base.Dispose();
+                typeof(MeterProvider<int>).Field<Func<object, int>>().Set(Utility.FinalReleaseComObject);
+            }
+
             [Fact]
-            public void HasFabricCreateMeterProviderFunc()
+            public void DisposesNativeFabricMeterProvider()
```
olegsych: You mean releases, right?

test/Diagnostics/Metrics/Implementation/MeterProviderTest.cs
```diff
@@ -55,52 +55,50 @@ public void SetsRequiredDimensionsFromServiceContext()
...
         }
 
-        class TestMeterProvider<TValueType> : MeterProvider<TValueType>
+        public class CreateMeterProvider : DisposeTest
         {
-            public TestMeterProvider(ServiceContext serviceContext)
-                : base(serviceContext)
-            {
-            }
+            readonly Func<string, string, IEnumerable<string>, IFabricMeter> sutMethod;
```
olegsych: Let's name the variable consistently with the target method. This is kinda hard to read.

test/Diagnostics/Metrics/Implementation/MeterTest.cs
```diff
@@ -87,33 +89,38 @@ public void CallsFabricMeterWithCombinedDimensions()
             }
         }
 
-        public class RecordViaNative : MeterTest
+        public class DisposeTest : MeterTest, IDisposable
+        {
+            public DisposeTest() => typeof(Meter).Field<Func<object, int>>().Set(objectParam => fuzzy.Int32());
```
olegsych: Yeah, this is really hard to read without comments, it starts to feel like a missing abstraction.

test/Services.Remoting/V2/Diagnostic/TelemetryDiagnosticEventsTest.cs
```diff
@@ -109,5 +108,29 @@ public void OnCreateTransportEndObserveRequestProcessingTime()
...
+                mockRequestDeserializationTime = sut.Field<IMeter<TimeSpan>>("requestDeserializationTime").Value;
+                mockResponseSerializationTime = sut.Field<IMeter<TimeSpan>>("responseSerializationTime").Value;
+            }
+
+            [Fact]
+            public void DisposesAllMeters()
+            {
+                sut.Dispose();
+
+                Mock.Get(mockRequestProcessingTime).Verify(m => m.Dispose(), Times.Once);
```
olegsych: This is where the mock prefix becomes particularly redundant.

test/Services.Remoting/V2/FabricTransport/Runtime/FabricTransportMessageHandlerTest.cs
```diff
@@ -128,29 +127,47 @@ public async Task ShouldCallDiagnostics_InCorrectSequence()
...
+
+                Mock.Get(diagnosticEvents).Verify(d => d.Dispose(), Times.Once);
+            }
+
+            [Fact]
+            public void DisposesDisposableRemotingMessageHandler()
+            {
+                sut.Dispose();
+
+                Mock.Get(serviceRemotingMessageHandler).As<IDisposable>().Verify(d => d.Dispose(), Times.Once);
```
olegsych, line 162: Assuming Mock generates decent assertion failure messages, having separate tests for each disposable feels like an overkill.

test/Services.Remoting/V2/FabricTransport/Runtime/FabricTransportServiceRemotingListenerTest.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            sut.Field<IMeterProvider<TimeSpan>>().Set(mockMeterProvider);
+            sut.Field<IFabricTransportListener>().Set(mockFabricTransportListener);
+            sut.Field<IFabricTransportMessageHandler>().Set(mockTransportMessageHandler);
+        }
+
+        public class Dispose : FabricTransportServiceRemotingListenerTest
+        {
+            [Fact]
+            public void AbortDisposesMeterProvider()
```
olegsych: Same here, a separate test for each disposable feels like an overkill. But if you want to keep them separate, I would put the sut.Abort() call in the constructor of this test class, that way the assertions becomes a single line and we can use expression bodies for the constructor and the test methods.

test/Services.Remoting/V2/FabricTransport/Runtime/FabricTransportServiceRemotingListenerTest.cs
```diff
@@ -0,0 +1,57 @@
...
+        readonly IFabricTransportMessageHandler mockTransportMessageHandler = Mock.Of<IFabricTransportMessageHandler>();
+
+        public FabricTransportServiceRemotingListenerTest()
+        {
+            sut.Field<IMeterProvider<TimeSpan>>().Set(mockMeterProvider);
+            sut.Field<IFabricTransportListener>().Set(mockFabricTransportListener);
+            sut.Field<IFabricTransportMessageHandler>().Set(mockTransportMessageHandler);
+        }
+
+        public class Dispose : FabricTransportServiceRemotingListenerTest
```
olegsych: This class should be called Abort correct?

olegsych (commented):
Thanks @ilijagrbic I love how thoroughly we're testing this. Look forward to the next iteration.

src/FabricTransport/Runtime/IFabricTransportListener.cs
ilijagrbic: Yes, this is what enabled us to test the `FabricTransportServiceRemotingListener` and `FabricTransportMessageHandler` changes. These classes are used to emit the metrics from the Remoting stack.

src/FabricTransport/Runtime/IFabricTransportMessageHandler.cs
ilijagrbic: `FabricTransportMessageHandler` didn't actually implement the IDisposable interface, but it did have a `Dispose()` method implemented in it. I wanted to ensure I can verify these calls, so that's why I added the IDisposable interface to the interface contract. I think it should be added regardless of the tests, since it's weird that this class had defined its own `Dispose()` method.

test/Actors/Diagnostics/DiagnosticsFactoryTest.cs
ilijagrbic: Makes sense, I will rename them.

test/Diagnostics/Metrics/Implementation/MeterProviderTest.cs
ilijagrbic: I think for both this one and the previous instance, there was a suggestion to not use var, probably because of the new .editorconfig and I didn't think this through too much, just took the editor suggestion. I was trying to clean up as many violation of the new rules as possible.

ilijagrbic: I will revert the change.

test/Diagnostics/Metrics/Implementation/MeterProviderTest.cs
ilijagrbic: Yup, thanks for noticing this.

test/Diagnostics/Metrics/Implementation/MeterProviderTest.cs
ilijagrbic: Perhaps in this specific case it is a bit harder. But in general, I try to keep the fields under test names sut (or sutMethod), since for example here, the class name is the name of the method being tested. I will rename it, just giving my thoughts.

test/Diagnostics/Metrics/Implementation/MeterTest.cs
ilijagrbic: Ordinarily I would agree. This clunky way of mocking is the result of `FinalReleaseComObject` being a static extension function. Static functions are notoriously difficult or inconvenient to test. So, in this case we don't have the choice to refactor it, since it would cost way more than we would gain. I will add some comments to hopefully make the tests more understandable, and if this issue persist, as we write more unit tests that verify `FinalReleaseComObject`, we can think of a better solution repository wide.

test/Services.Remoting/V2/Diagnostic/TelemetryDiagnosticEventsTest.cs
ilijagrbic: Interestingly, this test was generated by Copilot, and the reason it generated it this way could be because it tried to mimic the same style as already present in the document (where we had similar naming for previous tests). Goes to show that the better is our code, the better will be whatever we use Copilot for.

ilijagrbic: I will rename.

test/Services.Remoting/V2/FabricTransport/Runtime/FabricTransportMessageHandlerTest.cs
ilijagrbic: I was thinking about this and going back and forth. I will unify the tests.

src/Diagnostics/Metrics/Implementation/Meter.cs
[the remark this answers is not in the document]
ilijagrbic: I have tested this, and I do get failures on the double release occasionally on, namely `System.AccessViolationException` when I try the release step-by-step. So far, I only got these failures during debugging and on new8.0, but regardless, I don't see any downsides to releasing only once anyways, other than slightly more complex logic.

By the way, great suggestion, testing this manually made me realize we are missing `MarshalUsing(typeof(UniqueComInterfaceMarshaller<IFabricMeter>))` from the methods that return the COM objects. I have also added this to the change set, and will call it out in the description.

src/Diagnostics/Metrics/Implementation/IFabricMeter.cs
```diff
@@ -5,10 +5,9 @@
 
 using System;
 using System.Runtime.InteropServices;
-
-#if NET
 using System.Runtime.InteropServices.Marshalling;
-#else
+
```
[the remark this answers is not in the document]
olegsych: @copilot This is incorrect. We define shim types in .NET Framework version of System.Fabric.dll to reduce conditional compilation.

src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
```diff
@@ -5,10 +5,9 @@
 
 using System;
 using System.Runtime.InteropServices;
-
-#if NET
 using System.Runtime.InteropServices.Marshalling;
-#else
+
```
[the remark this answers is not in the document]
olegsych: @copilot This is incorrect. We define shim types in .NET Framework version of System.Fabric.dll to reduce conditional compilation.

src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
```diff
@@ -19,6 +18,7 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
     [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
     partial interface IFabricMeterProvider
     {
+        [return: MarshalUsing(typeof(UniqueComInterfaceMarshaller<IFabricMeter>))]
         IFabricMeter CreateMeter([MarshalAs(UnmanagedType.LPWStr)] string metricNamespace, [MarshalAs(UnmanagedType.LPWStr)] string name, uint count, [MarshalAs(UnmanagedType.LPArray, ArraySubType = UnmanagedType.LPWStr)] string[] dimensionNames);
```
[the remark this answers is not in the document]
olegsych: @copilot This is incorrect. We define shim types in .NET Framework version of System.Fabric.dll to reduce conditional compilation.

src/Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -5,38 +5,46 @@
...
+        protected long ConvertTimeSpanToLong(TimeSpan value) => (long)Math.Round(value.TotalMilliseconds);
+
+        bool IsDisposed() => fabricMeter == null;
 
-        unsafe protected void RecordViaNative(long value, int customDimensionCount, string dimension1Value, string dimension2Value, string dimension3Value)
+        protected unsafe void RecordViaNative(long value, int customDimensionCount, string dimension1Value, string dimension2Value, string dimension3Value)
         {
+            if (IsDisposed())
+                throw new ObjectDisposedException(nameof(Meter));
             if (customDimensionCount < 0 || customDimensionCount > 3)
```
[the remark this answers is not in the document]
olegsych: @ilijagrbic I think this is a legitimate problem, but the recommended solution is not going to work either because even if the local meter variable won't be null, it could still be pointing to a released COM object, so I don't know if it's worth it.

src/Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -5,38 +5,46 @@
...
         {
             this.systemDimensionValues = systemDimensionValues ?? throw new ArgumentNullException(nameof(systemDimensionValues));
             this.fabricMeter = fabricMeter ?? throw new ArgumentNullException(nameof(fabricMeter));
         }
 
-        protected void Record(long value)
+        public void Dispose()
         {
-            RecordViaNative(value, 0, null, null, null);
+            if (!IsDisposed())
```
olegsych, line 29: I've looked it up in the .NET source code

https://github.com/dotnet/dotnet/blob/b0f34d51fccc69fd334253924abd8d6853fad7aa/src/runtime/src/libraries/System.Runtime.InteropServices/src/System/Runtime/InteropServices/Marshalling/ComObject.cs#L86

In .NET 10, ComObject.FinalRelease can be safely called multiple times, but not in .NET 8.  

https://github.com/dotnet/dotnet/blob/86c1b454f17232e86cccbd509c734630eb55afef/src/runtime/src/libraries/System.Runtime.InteropServices/src/System/Runtime/InteropServices/Marshalling/ComObject.cs#L80

Let's add a comment to remind ourselves to remove this in the next major version of the SF SDK.

src/Diagnostics/Metrics/Implementation/NativeTelemetry.cs
```diff
@@ -4,7 +4,9 @@
 // ------------------------------------------------------------
 
 using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.Marshalling;
 using HRESULT = System.Int32;
```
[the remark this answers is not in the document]
olegsych: @copilot This is incorrect. We define shim types in .NET Framework version of System.Fabric.dll to reduce conditional compilation.

test/Actors/Diagnostics/DiagnosticsFactoryTest.cs
```diff
@@ -86,20 +85,36 @@ public void InstantiatesMeterProviders()
...
+            readonly Mock<PerformanceCounterProviderV2> performanceCounterProviderV2;
+            readonly Mock<IMeterProvider<TimeSpan>> timeSpanProvider;
+            readonly Mock<IMeterProvider<long>> longProvider;
 
             public DisposeTest()
             {
-                mockPerformanceCounterProviderV2 = new Mock<PerformanceCounterProviderV2>(guid, typeInformation);
-                sut.Field<PerformanceCounterProviderV2>().Set(mockPerformanceCounterProviderV2.Object);
+                performanceCounterProviderV2 = new Mock<PerformanceCounterProviderV2>(guid, typeInformation);
+                timeSpanProvider = new Mock<IMeterProvider<TimeSpan>>() { DefaultValue = DefaultValue.Mock };
```
olegsych, line 95: I think the `performanceCounterProviderV2` needs to be initialized by the constructor because it takes instance fields as parameters, but the meter provider mocks should be possible to initialize inline.

test/Actors/Diagnostics/DiagnosticsFactoryTest.cs
```diff
@@ -86,20 +85,36 @@ public void InstantiatesMeterProviders()
...
             public void DisposesPerformanceCounterProvider()
             {
                 sut.Dispose();
 
-                mockPerformanceCounterProviderV2.Verify(p => p.Dispose(), Times.Once);
+                performanceCounterProviderV2.Verify(p => p.Dispose(), Times.Once);
+            }
+
+            [Fact]
+            public void DisposesMeterProviders()
```
olegsych: Combine with the test above?

olegsych (approved):
Thanks again @ilijagrbic Everything looks good, feel free to merge.

test/Actors/Diagnostics/DiagnosticsFactoryTest.cs
ilijagrbic: Maybe I'm missing something here, but I think `new Mock<PerformanceCounterProviderV2>(guid, typeInformation)` calls the constructor for the provider.

src/Diagnostics/Metrics/Implementation/Meter.cs
ilijagrbic: Thanks for looking this up @olegsych . This is consistent with the observations I had during testing, where the sporadic failures occurred only on .NET 8.

test/Actors/Diagnostics/DiagnosticsFactoryTest.cs
olegsych: That's right, I was saying that the timeSpanProvider and longProvider fields can be initialized inline instead of the test class constructor.

test/Diagnostics/Metrics/Implementation/MeterProviderTest.cs
```diff
@@ -6,56 +6,58 @@
...
-            }
+            typeof(MeterProvider<int>).Field<Func<IFabricMeterProvider>>().Set(() => fabricMeterProvider);
+            sut = new Mock<MeterProvider<int>>(serviceContext).Object;
+        }
 
-            public void Dispose()
-            {
-                typeof(MeterProvider<int>).Field<Func<IFabricMeterProvider>>().Set(NativeTelemetry.FabricCreateMeterProvider);
-            }
+        public virtual void Dispose() => typeof(MeterProvider<int>).Field<Func<IFabricMeterProvider>>().Set(NativeTelemetry.FabricCreateMeterProvider);
```
olegsych, line 33: I like expression-bodied methods, but still feel that they are more readable when the expression body is on a separate line, particularly when they are long enough to wrap around in the PR UX.

olegsych (approved):
:shipit:

src/Diagnostics/Metrics/Implementation/Meter.cs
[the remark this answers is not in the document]
ilijagrbic: Thanks @olegsych , as discussed, I will tackle this in a separate PR, but I won't treat is as a priority right now.

PR #546, opened 2026-03-13
Integrate FabricTransport tests

olegsych:
- Bump version of the `Microsoft.ServiceFabric` packages still produced from the internal `WindowsFabric` repo to `12/9.0.892-beta` to restore `FabricTransport.Tests.csproj` access to internals that was lost after the change from full signing with test key to partial signing with product key during the initial port to GitHub.
- Introduce `WindowsOnlyAttribute` in the `TestFramework.csproj` to help with skipping of tests that can't run on Linux yet. 
- Cleanup `FabricTransport.Tests.csproj`
  - Split `ServiceCommunication.Test.cs` into `FabricTransportSettingsTest.cs` and `FabricTransportListenerSettingsTest.cs`
  - Remove unnecessary, duplicated code and configuration
- Add `FabricTransport.Tests.csproj` to the `code.slnx`


test/FabricTransport/FabricTransportSettingsTest.cs
```diff
@@ -0,0 +1,84 @@
...
+public abstract class FabricTransportSettingsTest
+{
+    const string onLinux = "Can't load FabricCommon.dll";
+
+    readonly string settingsFile = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "ServiceCommunicationTestSettings.xml");
+
+    public sealed class LoadFrom: FabricTransportSettingsTest
+    {
+        [Fact(Skip=onLinux, SkipUnless=nameof(TestEnvironment.IsWindows), SkipType=typeof(TestEnvironment))]
+        public void ReadTransportSettingsFromConfig()
```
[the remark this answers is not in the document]
olegsych: @copilot what you said in this comment is incorrect.

test/FabricTransport/NativeMessageStreamTest.cs
```diff
@@ -6,179 +6,94 @@
...
 {
-    public class NativeMessageStreamTest
+    readonly NativeMessageStream sut;
+
+    // Constructor parameters
+    readonly List<Tuple<uint, nint>> bufferList;
+
+    // Test fixture
+    static readonly IFuzz fuzzy = new RandomFuzz(Environment.TickCount);
+    readonly PinCollection pins = [];
```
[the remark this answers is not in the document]
olegsych: @copilot what you said in this comment is incorrect.

test/FabricTransport/FabricTransportListenerSettingsTest.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Microsoft.ServiceFabric.FabricTransport;
+
+public abstract class FabricTransportListenerSettingsTest
+{
+    const string onLinux = "Can't load FabricCommon.dll";
+
+    public sealed class LoadFrom: FabricTransportListenerSettingsTest
+    {
+        [Fact(Skip=onLinux, SkipUnless=nameof(TestEnvironment.IsWindows), SkipType=typeof(TestEnvironment))]
+        public void ThrowsExceptionWhenCodePackageDoesNotExist() =>
```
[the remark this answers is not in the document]
olegsych: @copilot what you've said in this comment is incorrect.

test/FabricTransport/FabricTransportListenerSettingsTest.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Microsoft.ServiceFabric.FabricTransport;
+
+public abstract class FabricTransportListenerSettingsTest
+{
+    const string onLinux = "Can't load FabricCommon.dll";
+
+    public sealed class LoadFrom: FabricTransportListenerSettingsTest
+    {
+        [Fact(Skip=onLinux, SkipUnless=nameof(TestEnvironment.IsWindows), SkipType=typeof(TestEnvironment))]
```
ilijagrbic: This is really a nit, but can we use [`SkipWhen`](https://api.xunit.net/v3/3.0.0/v3.3.0.0-Xunit.FactAttribute.SkipWhen.html#v3_3_0_0_Xunit_FactAttribute_SkipWhen) here? I think this would be more readable than a negative of `IsWindows`, especially considering the other condition is `Skip=onLinux`.

We can add a `IsLinux` property in the `TestEnvironment` that itself is a negation of `IsWindows`.

test/FabricTransport/FabricTransportListenerSettingsTest.cs
```diff
@@ -0,0 +1,33 @@
...
+        public void ThrowsExceptionWhenCodePackageDoesNotExist() => 
+            Assert.Throws<ArgumentException>(() => FabricTransportListenerSettings.LoadFrom("TestServiceListener", "Config1"));
+
+        [Fact(Skip=onLinux, SkipUnless=nameof(TestEnvironment.IsWindows), SkipType=typeof(TestEnvironment))]
```
ilijagrbic: Consider introducing an attribute to skip entire test class on Linux, since all the tests here are skipped because of the same reason and I imagine if we end up fixing this, they will be fixed at the same time. Also, I think this is not the first time we have to skip test cases/test classes on Linux. I briefly looked into this, and it looks like entire class can be easily skipped by doing the following:
```
[AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
public sealed class SkipOnLinuxAttribute : BeforeAfterTestAttribute
{
    public override void Before(System.Reflection.MethodInfo methodUnderTest, IXunitTest test)
    {
        if (isLinux) throw SkipException.ForSkip("Placeholder message");
    }
}
```

After that we just need to annotate the classes we need to skip:
```
[SkipOnLinux]
public abstract class FabricTransportListenerSettingsTest
{
...
}
```

test/FabricTransport/NativeMessageStreamTest.cs
```diff
@@ -6,179 +6,94 @@
...
-using System.Text;
-using System.Xml;
-using Microsoft.ServiceFabric.FabricTransport;
+using Fuzzy;
 using Xunit;
 
-namespace Microsoft.ServiceFabric.Services.Messaging.Test
+namespace Microsoft.ServiceFabric.FabricTransport;
+
+public abstract class NativeMessageStreamTest: IDisposable
```
ilijagrbic, line 15: I like how much code we deleted from this class, it feels much more readable and useful as a test.

ilijagrbic (approved):
Hey @olegsych, the changes look good, no blockers. I had some minor suggestions around the way we are skipping the tests on Linux, but other than that it's good to go. :shipit:

test/FabricTransport/FabricTransportListenerSettingsTest.cs
olegsych: That was my first thought as well. Unfortunately, there is no API to positively determine Linux on .NET Framework.

test/FabricTransport/FabricTransportListenerSettingsTest.cs
olegsych: Great idea. I didn't know this existed. In all discussions I came across where people asked for a class-level skip attribute in xUnit, Brad Wilson rejected the idea because he strongly feels suppressions should be temporary.

ilijagrbic (approved):

PR #556, opened 2026-03-18
Dispose Diagnsotics in ActorService Abort

ilijagrbic:
Diagnostic related objects are already being disposes in ActorService `OnCloseAsync`. 
Introduce the same dispose in `OnAbort` code path, as it was missed during the initial handling of deterministic release.

Opportunistically fix some of the .editorconfig warnings.

### DIFF
--- a/src/Actors/Runtime/ActorService.cs
+++ b/src/Actors/Runtime/ActorService.cs
@@ -18,374 +18,307 @@ using Microsoft.ServiceFabric.Services;
 using Microsoft.ServiceFabric.Services.Communication.Runtime;
 using Microsoft.ServiceFabric.Services.Runtime;
 
-namespace Microsoft.ServiceFabric.Actors.Runtime
+namespace Microsoft.ServiceFabric.Actors.Runtime;
+
+/// <summary>
+/// Represents the base class for Microsoft Service Fabric based reliable actors service.
+/// </summary>
+/// <remarks>
+/// Derive from this class to implement your own custom actor service if you want to override
+/// any service level behavior for your actors.
+/// </remarks>
+public class ActorService : StatefulServiceBase, IActorService
 {
+    const string traceType = "ActorService";
+
+    readonly ActorTypeInformation actorTypeInformation;
+    readonly IActorStateProvider stateProvider;
+    readonly ActorServiceSettings settings;
+    readonly IActorActivator actorActivator;
+    readonly ActorManagerAdapter actorManagerAdapter;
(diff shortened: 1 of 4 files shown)


test/Actors/Runtime/ActorServiceTest.cs
```diff
@@ -14,85 +14,92 @@
...
-            }
+            sut.DeclaredBy(typeof(ActorService)).Method<Func<CancellationToken, Task>>("OnCloseAsync").Invoke(TestContext.Current.CancellationToken);
+
+            Mock.Get(diagnostics).Verify(d => d.Dispose(), Times.Once);
+            Mock.Get(diagnosticsFactory).Verify(d => d.Dispose(), Times.Once);
         }
 
-        public class OnRoleChange : ActorServiceTest
+        [Fact]
+        public void IsDisposedByAbort()
```
ilijagrbic, line 67: I haven't separated the test .editorconfig refactor, so calling out the only added test in the diff here.

src/Actors/Runtime/IActorStateProviderInternal.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.ServiceFabric.Actors.Runtime
     internal interface IActorStateProviderInternal
     {
         /// <summary>
-        /// Gets TraceType used for tracing.
+        /// Gets traceType used for tracing.
```
olegsych, line 17: This is unexpected change. Great example of why we shouldn't add unnecessary doc comments :)

src/Actors/Runtime/ActorService.cs
```diff
@@ -18,366 +18,307 @@
 using Microsoft.ServiceFabric.Services.Communication.Runtime;
 using Microsoft.ServiceFabric.Services.Runtime;
 
-namespace Microsoft.ServiceFabric.Actors.Runtime
+namespace Microsoft.ServiceFabric.Actors.Runtime;
```
olegsych, line 21: Ugh... This is rough. We should definitely separate style and functional changes in the future.

olegsych (approved):
I _think_ this is OK, but the style changes made it too hard to review.

src/Actors/Runtime/ActorService.cs
ilijagrbic: Thanks for the feedback and for the review! I will separate them in the future. In particular the namespace refactor is the worst offender from the perspective of the diff, while being not as important even as a style change.

PR #558, opened 2026-03-20
Update .editorconfig and copilot instructions

olegsych:
- Expand `copilot-instructions.md` with details on principles, style, etc.
- Extract from `copilot-instructions.md` a new `instructions/test.instructions.md` and expand it with more specifics and highlights.
- Change `instructions/inspector.instructions.md` based on Copilot's suggestions to a small API cheat sheet and pull the full source code from the original GitHub repo.
- Create `instructions/fuzzy.instructions.md` similar to the new `inspector.instructions.md`
- Update `.editorconig` with remainig .NET code styles that weren't incorporated before. 
- Make Copilot rewrite `OrdinalStringTest` using prompt `Discard the previous session, start from scratch and rewrite this test class based on latest instructions`


.github/instructions/test.instructions.md
```diff
@@ -0,0 +1,308 @@
...
+    {
+        [Fact]
+        public void ReturnsBuz()
+        {
+            string result = sut.Bar();
+            Assert.Equal("Buz", result);
+        }
+    }
+}
+```
```
ilijagrbic, line 230: Why not add these examples to the first example of the test structure?
It seems like we are repeating a lot of similar examples there the main goal is to better explain 1 concept to the copilot. Perhaps we can just add a bullet point on the arrange/act and give an example as one of the test cases.

.github/instructions/test.instructions.md
```diff
@@ -0,0 +1,308 @@
...
+// Simple mock (most cases)
+readonly ICommunicationListener listener = Mock.Of<ICommunicationListener>();
+
+// Mock needing configuration
+readonly IServiceInstance instance = new Mock<IServiceInstance> { DefaultValue = DefaultValue.Mock }.Object;
+
+// Setup and verify via Mock.Get
+Mock.Get(listener).Setup(l => l.OpenAsync(It.IsAny<CancellationToken>())).ReturnsAsync("endpoint");
+Mock.Get(listener).Verify(l => l.OpenAsync(It.IsAny<CancellationToken>()), Times.Once);
+```
```
ilijagrbic, line 265: We don't have to change this right away, but I'm curious if copilot would know this by itself already. We are explaining the Fuzzy/Inspector usage since they are less widely known libraries, but there should be tons of example on Moq online, and I would assume they were used in the training data for the newest models. 

So, for this case I think it would make more sense to define our preferences by just specifying that we prefer `Mock.Of<T>()` if possible and then let the model decide what is the best choice for the rest of the use-cases. 

I would also add that we prefer to do these kinds of setup in constructors, to avoid repeated similar setups for various test cases. Perhaps that would be even more generalized guideline, not only for Mocks.

.github/instructions/test.instructions.md
```diff
@@ -0,0 +1,308 @@
...
+using Inspector;
+
+Bar value = sut.Field<Bar>().Value; // Access private fields
+Foo foo = Type<Foo>.New(); // Create types with private constructors
+Foo foo = Type<Foo>.Uninitialized(); // Skip constructor
+```
+
+## Platform-Specific Tests
+
+Use `[WindowsOnly("reason")]` from `TestFramework` to skip tests on Linux.
```
ilijagrbic, line 299: We should keep an eye on whether the copilot uses other features from the TestFramework. 
Perhaps we could add a generalized message to "look at the TestFramework for shared test utilities if possible" and if this doesn't prove to be useful, we can specify it in more detail.

.github/copilot-instructions.md
```diff
@@ -1,48 +1,107 @@
...
+
+- Reduce potential merge conflicts in the future
+  - Keep package, project and assembly references in separate groups, sorted alphabetically in the project and solution files.
+  - Keep using directives sorted alphabetically in the C# files, with `System` assemblies above others.
+
+- Read and follow the guidelines codified in the [.editorconfig](../.editorconfig) before writing or reviewing code
+  - Pay attention to the `file_header_template` setting; existing files may have legacy headers that differ from the template
+
+### Documentation
+- Add doc-comments for public APIs only
```
ilijagrbic, line 92: This was a much-needed instruction. :)

ilijagrbic (approved):
Looks good @olegsych! These all look like great improvements on the instruction set we have in the repo. 
I look forward to seeing them in practice and future iteration. The `OrdinalStringTest` example seems like it worked great.
I wonder how well all of these instructions are used by the Copilot on GitHub especially during reviews. According to [this](https://docs.github.com/en/copilot/tutorials/use-custom-instructions), it should work, but I want to see it in practice as well.

.github/instructions/test.instructions.md
olegsych: This is a good point. I just started learning how to author Copilot instructions, so this will take a few more iterations. I'll try to do this on one of my small repos first and circle back to update this.

.github/instructions/test.instructions.md
olegsych: Good points. Let's iterate on this in follow-up PRs. I feel that Moq can be used in too many different ways and we still need instructions specific to our repository because we know enough about it to have specific preferences.

.github/instructions/test.instructions.md
olegsych: Good point. There is general description of the TestFramework in the copilot-instructions.md, so I think it already does. But then the question becomes about completeness, duplication, conflicts, etc. from both human and copilot perspective. The key technique learned about recently is to iterate with copilot on both the actions it is taking and the instructions we're providing, ask why it did or didn't do something, ask how the instructions can be improved, adjust instructions for readability by humans, ask to review the instructions, and then repeat the process from scratch with updated instructions.

ilijagrbic (approved):

PR #559, opened 2026-03-21
Switch from hard-coded versions to Nerdbank.GitVersioning

olegsych:
The `Nerdbank.GitVersioning` package is maintained by the .NET Foundation. When referenced by a C# project, it dynamically generates version numbers based on the settings in the `version.json` file and the number of commits in the Git history since it was updated, a.k.a. "height".

Instead of hard-coding versions in the `service_fabric*.props` files we now specify only the base version number in the `version.json`. The number will be automatically incremented for each new commit. The last hard-coded version was `12.0.0-beta.28` and the first Git-based version produced from develop should be `12.0.0-beta.29`.

For branches other than `develop`, Nerdbank will automatically append the Git commit to differentiate versions with the same height but different branches. For example, the PR validation build produced version `12.0.0-beta.30.g6cd49399a9.` because it checked out the pull request merge commit `6cd49399a930344cc22bcfab02155a61164a3d0c`. Also note that because of the extra merge commit, the build version is 30 instead of 29 that we should have once the changes are merged into develop as a single squash commit.

The `build.yml` was modified to checkout the Git repository with full commit history instead of the shallow 1-commit history checked out by default. Although fetching the full commit history is an overkill, our repository is small enough that it didn't produce a noticeable increase in the duration of the checkout step.

The `build.yml` actions `checkout` and `upload-artifact` were also upgraded to their latest versions to avoid deprecation warnings about older Node.js versions they used.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,13 +23,15 @@ jobs:
   test-windows:
     runs-on: windows-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
       - run: .\SkipStrongName.ps1
       - run: .\UninstallSfSdkFromGac.ps1
       - run: dotnet build -bl:"$env:Logs/build.binLog"
       - run: dotnet test --no-build --logger trx -bl:"$env:Logs/test.binLog"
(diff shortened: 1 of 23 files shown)


ilijagrbic (approved):
Great changes @olegsych ! I was guilty of forgetting to bump versions in my PRs multiple times, so this is a welcomed change. :)

PR #560, opened 2026-03-24
[ImpactApproval] [SelfReconfig] HTTP Code generation for SRS and IA APIs

iYzeee23:
**Summary**
- Generated HTTP client and PowerShell cmdlet code for the `ClearImpactObjects`, `StartImpactAction` and `CompleteSelfReconfiguration` internal APIs
- These are DRI mitigation tools for unblocking stuck impact approval objects and stuck self-reconfigurations. The code was generated from the Swagger spec defined in the SF-Swagger PR to evaluate the generated output quality before deciding on public visibility

**Old WindowsFabric PRs**
- [PR 14707667: ClearImpactObjects and StartImpactAction](https://dev.azure.com/msazure/One/_git/WindowsFabric/pullrequest/14707667)
- [PR 14631993: CompleteSelfReconfiguration](https://dev.azure.com/msazure/One/_git/WindowsFabric/pullrequest/14631993)

**New WindowsFabric PR**
- [PR 15225207: Changes in naming convention](https://dev.azure.com/msazure/One/_git/WindowsFabric/pullrequest/15225207)

**Related ServiceFabricSwagger PR**
- [PR 15158930: Swagger spec](https://dev.azure.com/msazure/One/_git/SF-Swagger/pullrequest/15158930)

**TSG followed**:
- [Generate HTTP code](https://msazure.visualstudio.com/One/_wiki/wikis/ServiceFabric/967066/Generate-HTTP-Code)


src/Client.Http/Common/Generated/ImpactActionKind.cs
```diff
@@ -0,0 +1,23 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Microsoft.ServiceFabric.Common
+{
+    /// <summary>
+    /// Defines values for ImpactActionKind.
+    /// </summary>
+    public enum ImpactActionKind
```
olegsych, line 11: I think the term Kind is redundant in this context. In combination with the ImpactOperationKind it can also be confusing in some contexts. Can we remove it?

I'm not sure if the term Action is precise enough. If this enum is meant to be used for approvals only, Can we use the term Approval instead? Perhaps ImpactApprovalAction, consistently with the ImpactApprovalStatus?

src/Client.Http/Common/Generated/ImpactActionKind.cs
```diff
@@ -0,0 +1,23 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+namespace Microsoft.ServiceFabric.Common
+{
+    /// <summary>
+    /// Defines values for ImpactActionKind.
```
olegsych, line 9: Please add a real summary to the Swagger definition.

src/Powershell.Http/Generated/CompleteSelfReconfigurationCmdlet.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Microsoft.ServiceFabric.Powershell.Http
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Management.Automation;
+    using Microsoft.ServiceFabric.Common;
+
+    /// <summary>
+    /// Force-completes a stuck self-reconfiguration on a partition.
```
olegsych: Could you replace all instances of "force-completes" with proper English, something along the lines "Forces completion of a..."

src/Client.Http/Client/Generated/IClusterClient.cs
```diff
@@ -797,6 +797,68 @@ Task ToggleVerboseServicePlacementHealthReportingAsync(
...
+        /// this parameter is 60 seconds.</param>
+        /// <param name ="cancellationToken">Cancels the client-side operation.</param>
+        /// <returns>
+        /// A task that represents the asynchronous operation.
+        /// </returns>
+        /// <exception cref="InvalidCredentialsException">Thrown when invalid credentials are used while making request to cluster.</exception>
+        /// <exception cref="ServiceFabricRequestException">Thrown when request to Service Fabric cluster failed due to an underlying issue such as network connectivity, DNS failure or timeout.</exception>
+        /// <exception cref="ServiceFabricException">Thrown when the requested operation failed at server. Exception contains Error code <see cref="FabricError.ErrorCode"/>, message indicating the failure. It also contains a flag wether the exception is transient or not, client operations can be retried if its transient.</exception>
+        /// <exception cref="OperationCanceledException">Thrown when cancellation is requested for the cancellation token.</exception>
+        Task StartImpactActionAsync(
```
olegsych: We only have two action kinds, Unknown and Approve. What does it mean to start impact action of kind "Unknown"? Do we expect additional action kinds?

src/Client.Http/Common/Generated/ImpactSelector.cs
```diff
@@ -0,0 +1,83 @@
...
+        public ImpactApprovalStatus? StatusFilter { get; }
+
+        /// <summary>
+        /// Gets only clear impact objects of this operation kind (e.g., Restart, Remove, Add). Possible values include:
+        /// 'Unknown', 'Restart', 'Remove', 'Add'
+        /// 
+        /// The type of operation associated with the impacted instances (e.g., restarting, removing, or adding the impacted
+        /// instance).
+        /// </summary>
+        public ImpactOperationKind? KindFilter { get; }
```
olegsych: I think the term "KindFilter" is ambiguous because we have multiple Impact-related Kind enums. As I mentioned earlier, the term Kind is also redundant. We should use the term Operation instead.

olegsych (changes requested):
Thanks for the PR. We'll need to iterate on the API. Please update the Swagger definitions and start a WindowsFabric PR to implement the updated API. We'll need to review all three pull requests at the same time.

src/Client.Http/Common/Generated/ImpactSelector.cs
iYzeee23: Changed this

src/Client.Http/Client/Generated/IClusterClient.cs
iYzeee23: We do expect. Most probably, it's going to be something like `Cancel` and it should lend in next month [or the one after it]. Specifically, about `Unknown`, it doesn't mean anything. Implementation should auto-reject it. I changed the description explaining that this is the default value. Let me know if this one requires additional changes

src/Powershell.Http/Generated/CompleteSelfReconfigurationCmdlet.cs
iYzeee23: Changed. Thanks for catching this

src/Client.Http/Common/Generated/ImpactActionKind.cs
iYzeee23: Added summary, please verify

src/Client.Http/Common/Generated/ImpactActionKind.cs
iYzeee23: Already talked about this one on the mail thread. so, we kind of explained why it's "impossible" to change them. We agreed on leaving ImpactOperationKind as it is, as it doesn't hurt. However, we need justification for ImpactActionKind. Now, not sure if you meant on leaving it in the docs or just in the PR comments, so I'll explain here:

1. `Cancel` action kind is coming soon, hence the Enum isn't approval-only, but rather an action dispatcher. Once `Cancel` is added, `ImpactActionKind` with values {`Approve`, `Cancel`} makes perfect sense. Renaming to ImpactApprovalAction now would be wrong within months

2. The PS cmdlet is no longer `Start`, but `Invoke` now. With `Invoke-ServiceFabricImpactAction`, the dispatch pattern is idiomatic:
   - `Invoke-ServiceFabricImpactAction -ImpactId <guid> -ActionKind Approve`
   - `Invoke-ServiceFabricImpactAction -ImpactId <guid> -ActionKind Cancel`

3. Also, the Enum name is consistent with its native-level counterpart. `ImpactActionKind` maps directly to `FABRIC_IMPACT_ACTION` in the IDL, which is consumed by `IM`. Keeping the managed name aligned with the native name avoids confusion for anyone debugging across the C++/C# boundary...

iYzeee23: Let me know if you have any other concern

src/Client.Http/Common/Generated/ImpactActionKind.cs
olegsych: Have you regenerated it? It still restates the ImpactActionKind.

src/Powershell.Http/Generated/NewImpactActionCmdlet.cs
```diff
@@ -0,0 +1,52 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Management.Automation;
+    using Microsoft.ServiceFabric.Common;
+
+    /// <summary>
+    /// Approves or acts on an impact approval object.
+    /// </summary>
+    [Cmdlet(VerbsCommon.New, "SFImpactAction")]
+    public partial class NewImpactActionCmdlet : CommonCmdletBase
```
olegsych: Could you double-check this? The class name and the verb use the term New instead of Invoke. We may have a problem with the generator.

src/Powershell.Http/Generated/NewImpactActionCmdlet.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// </summary>
+        [Parameter(Mandatory = true, ValueFromPipelineByPropertyName = true, Position = 0)]
+        public Guid? ImpactId { get; set; }
+
+        /// <summary>
+        /// Gets or sets ImpactActionKind. The action to perform on the impact object. Possible values include: 'Unknown',
+        /// 'Approve'
+        /// </summary>
+        [Parameter(Mandatory = true, ValueFromPipelineByPropertyName = true, Position = 1)]
+        public ImpactActionKind? ImpactActionKind { get; set; }
```
olegsych: Can we drop Kind from the property name?

src/Client.Http/Common/Generated/ImpactActionKind.cs
olegsych: I don't think this is any better, to be honest. InvokeImpactAction is so abstract, I don't think anyone can understand it without reading the documentation. As I've explained it in the email, this should have been simply ApproveImpact, but I understand it would take more work to go back and fix and I'm tired of debating this.

olegsych (changes requested):
Let's just make sure the PowerShell commandlet is correct.

src/Client.Http/Common/Generated/ImpactActionKind.cs
iYzeee23: I do understand the comment, but as already said, it won't be just "Approve" as an action, but also "Cancel" as well. Therefore, term "ImpactAction" should definitely be a part of the name. As this is internal cmdlet, I think it's okay for somebody to go through TSG and understand what the catch is with this command.

src/Client.Http/Common/Generated/ImpactActionKind.cs
iYzeee23: As explained in the mail, we'll stick with ImpactActionKind. We cannot introduce backwards incompatibility, as this name is widely used in internal ecosystem [ImpactManager user component].

src/Powershell.Http/Generated/NewImpactActionCmdlet.cs
iYzeee23: I forgot to update the Swagger PR. Now, it should be fine.

src/Powershell.Http/Generated/NewImpactActionCmdlet.cs
iYzeee23: Dropped it, and adjusted other PRs as well.

olegsych:
@iYzeee23 Could you update this PR with code generated from your Swagger PR? Let's get it merged.

iYzeee23:
> @iYzeee23 Could you update this PR with code generated from your Swagger PR? Let's get it merged.

@olegsych Yes, this is up-to-date. I just run the commands from SF Wiki and it didn't produce any new files.

olegsych (approved):

PR #562, opened 2026-03-26
Update createMeter API to pass fixed/system dimension values

ilijagrbic:
- Change `IFabricMeterProvider.CreateMeter` method to pass fixed dimension values on creation, to improve performance issues. Note that this is safe because `systemDimensions` are fixed and are stored when `MeterProvider` is created. Interface change itself is safe since the feature is still in development and we don't have to worry about the backwards comparibility.
- Pass only customDimension values to the interop in `Meter.Record`, since fixed (system) dimensions values are already known on creation.
- Delete `systemDimensionValues` field from `Meter `since it is now unused

Native optimization changes that require this PR to integrate: [PR](https://msazure.visualstudio.com/One/_git/WindowsFabric/pullrequest/15120943?path=/src/prod/src/Telemetry/Native/FabricMeter.cpp)


ilijagrbic:
@copilot Can you revert all of the license header changes from this PR? My local Copilot added them, and I don't want them right now in this PR before I understand why they occurred.

src/Diagnostics/Metrics/Implementation/IFabricMeter.cs
```diff
@@ -18,6 +18,6 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
     [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
     partial interface IFabricMeter
     {
-        void Record(long value, uint count, IntPtr dimensionValues);
+        void Record(long value, uint variableDimensionCount, IntPtr variableDimensionValues);
```
olegsych, line 21: Good idea to change the name of the count parameter. It could have been confused as the number of pre-aggregated metric values.

olegsych, line 21: I wonder if we should stick to the term "fixed" and continue using the term "dimension" without a qualifier to refer to the "variable" or "custom" dimensions.

src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
```diff
@@ -19,6 +19,6 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
     partial interface IFabricMeterProvider
     {
         [return: MarshalUsing(typeof(UniqueComInterfaceMarshaller<IFabricMeter>))]
-        IFabricMeter CreateMeter([MarshalAs(UnmanagedType.LPWStr)] string metricNamespace, [MarshalAs(UnmanagedType.LPWStr)] string name, uint count, [MarshalAs(UnmanagedType.LPArray, ArraySubType = UnmanagedType.LPWStr)] string[] dimensionNames);
+        IFabricMeter CreateMeter([MarshalAs(UnmanagedType.LPWStr)] string metricNamespace, [MarshalAs(UnmanagedType.LPWStr)] string name, uint count, [MarshalAs(UnmanagedType.LPArray, ArraySubType = UnmanagedType.LPWStr)] string[] dimensionNames, uint fixedDimensionCount, [MarshalAs(UnmanagedType.LPArray, ArraySubType = UnmanagedType.LPWStr)] string[] fixedDimensionValues);
```
olegsych: Let's rename the `count` here to be more precise as well.

src/Diagnostics/Metrics/Implementation/Int64Meter.cs
```diff
@@ -3,13 +3,11 @@
...
 // ------------------------------------------------------------
 
-using System.Collections.Generic;
-
 namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation
 {
     sealed class Int64Meter : Meter, IMeter<long>
     {
-        internal Int64Meter(IFabricMeter fabricMeter, IReadOnlyCollection<string> systemDimensionValues) : base(fabricMeter, systemDimensionValues) { }
+        internal Int64Meter(IFabricMeter fabricMeter) : base(fabricMeter) { }
```
olegsych, line 10: Love the simplification!

src/Diagnostics/Metrics/Implementation/Meter.cs
```diff
@@ -46,18 +42,12 @@ protected unsafe void RecordViaNative(long value, int customDimensionCount, stri
             if (customDimensionCount < 0 || customDimensionCount > 3)
                 throw new ArgumentOutOfRangeException(nameof(customDimensionCount));
 
-            int dimensionCount = systemDimensionValues.Count + customDimensionCount;
-
-            GCHandle* dimensionPins = stackalloc GCHandle[dimensionCount];
-            IntPtr* dimensionValuesPointers = stackalloc IntPtr[dimensionCount];
+            GCHandle* dimensionPins = stackalloc GCHandle[customDimensionCount];
```
olegsych: Let's use consistent and more precise terms here - i.e. "variable" instead of "custom" or drop the term altogether depending on the other discussion. Right now, we still have the mental baggage of the old implementation. But in this context, I feel this will be clear once the change is completed.

src/Diagnostics/Metrics/Implementation/MeterProvider.cs
```diff
@@ -26,28 +26,28 @@ protected MeterProvider(ServiceContext serviceContext = null)
 
             if (serviceContext != null)
             {
-                systemDimensionNames = new[]
-                {
+                systemDimensionNames =
```
olegsych: Let's use consistent terminology here, this should be "fixed" based on the current `IFabricMeterProvider`, I think.

src/Diagnostics/Metrics/Implementation/MeterProvider.cs
```diff
@@ -26,28 +26,28 @@ protected MeterProvider(ServiceContext serviceContext = null)
...
                     serviceContext.CodePackageActivationContext.ApplicationName,
                     serviceContext.CodePackageActivationContext.ApplicationTypeName
-                };
+                ];
             }
             else
             {
-                systemDimensionNames = Array.Empty<string>();
-                systemDimensionValues = Array.Empty<string>();
+                systemDimensionNames = [];
```
olegsych: Nice! Did VS suggest this based on the .editorconfig?

olegsych: I wonder if there was a rule to initialize variables inline

src/Diagnostics/Metrics/Implementation/MeterProvider.cs
olegsych: Below we also have "additional", which should be "variable"

src/Diagnostics/Metrics/Implementation/MeterProvider.cs
```diff
@@ -56,15 +56,17 @@ protected MeterProvider(ServiceContext serviceContext = null)
         protected IFabricMeter CreateNativeMeter(string metricNamespace, string metricName, IEnumerable<string> additionalDimensions)
         {
             if (IsDisposed())
-                throw new ObjectDisposedException(nameof(MeterProvider<TValueType>));
+                throw new ObjectDisposedException(nameof(MeterProvider<>));
 
-            var allDimensionsList = new List<string>(systemDimensionNames.Count + additionalDimensions.Count());
+            var allDimensionsNameList = new List<string>(systemDimensionNames.Count + additionalDimensions.Count());
```
olegsych: I love the new collection expressions. I think we can simplify this further, specifically the `allDimensionsNameList` could be eliminated.
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/collection-expressions#spread-element

test/Diagnostics/Metrics/Implementation/Meter2DTest.cs
```diff
@@ -19,22 +17,14 @@ public abstract class Meter2DTest
...
-                Assert.IsType<ArgumentNullException>(exception.InnerException);
-            }
-
-            [Fact]
-            public void ThrowsArgumentNullExceptionWhenSystemDimensionsAreNull()
-            {
-                var exception = Assert.Throws<TargetInvocationException>(() => new Mock<Meter2D>(fabricMeter, null).Object);
-                Assert.IsType<ArgumentNullException>(exception.InnerException);
+                var exception = Assert.Throws<TargetInvocationException>(() => new Mock<Meter2D>(null).Object);
+                _ = Assert.IsType<ArgumentNullException>(exception.InnerException);
```
olegsych, line 27: It would be great to verify the `ParamName` as well and not discard the exception instance. Try using Inspector's `.Constructor().Parameter()` methods. You might need to move the `sut` to the base class, I think it needs an instance.

test/Diagnostics/Metrics/Implementation/Meter3DTest.cs
```diff
@@ -80,10 +70,7 @@ public void CallsFabricMeterWithCombinedDimensions()
             [InlineData(null, "dimension2Value", null)]
             [InlineData(null, null, "dimension3Value")]
             [InlineData(null, null, null)]
-            public void ThrowsExceptionIfCustomDimensionIsNull(string dimension1Value, string dimension2Value, string dimension3Value)
-            {
-                Assert.Throws<ArgumentNullException>(() => sutMethod.Invoke(value, dimension1Value, dimension2Value, dimension3Value));
-            }
+            public void ThrowsExceptionIfCustomDimensionIsNull(string dimension1Value, string dimension2Value, string dimension3Value) => Assert.Throws<ArgumentNullException>(() => sutMethod.Invoke(value, dimension1Value, dimension2Value, dimension3Value));
```
olegsych: Let's put expression bodies like this on a new line. I feel this is too wide even for my 4K screen 😄

src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
olegsych: With 7 parameters, the `IFabricMeterProvider.CreateMeter` is at the limit of method complexity. In the context of the .NET SDK it's still OK because customers won't be calling it manually. But as a COM API for the native SDK consumers, it's pretty bad.

With this last-minute decision to optimize it, I also wonder if it is as stable as we think it is. I certainly didn't expect this modification, and it makes me worried that we'll need more, particularly with the ongoing discussion of instruments.

Given the borderline complexity and the risk of needing to extend it in the near future, I think we should follow the advice I've been giving to other teams and introduce a FABRIC_METER_DESCRIPTOR struct. It would make the API simpler to call now and allow us to avoid introducing a new interface in the future, more SOLID so to speak.

https://dev.azure.com/msazure/One/_wiki/wikis/ServiceFabric/978764/Design?anchor=do-use-parameter-objects

olegsych (changes requested):
This is awesome @ilijagrbic! I love how much simpler the design has become overall. The only blocking issue is the consistency of the terminology, which we already discussed in scrum.

sridmad:
**Cross-repo coordination note**

I reviewed the companion native PR ([ADO PR 15120943](https://msazure.visualstudio.com/One/_git/WindowsFabric/pullrequest/15120943)). The `FabricMeter` class on that branch already accepts `fixedDimensionsCount`/`fixedDimensionValues` in its constructor and handles them correctly (cached as UTF-8 pairs at construction, inserted into the attribute list on each `Record()`).

However, the COM bridge layer (`ComFabricMeterProvider`, `ComFabricMeter`, IDL, prebuilt header) still has the old 5-parameter `CreateMeter` signature. The managed side here sends 6 parameters (adding `fixedDimensionCount` + `fixedDimensionValues`), so both PRs need the COM layer updated before they can integrate. I left a comment on the ADO PR with the specific files that need updating.

On terminology -- the native side uses `fixedDimensionsCount`/`fixedDimensionValues` and `variableDimensionCount`/`variableDimensionValues`, which aligns with what @olegsych suggested. Once the naming is settled here, it should be mirrored on the native side.

src/Diagnostics/Metrics/Implementation/MeterProvider.cs
ilijagrbic: Yeah, this was one of the changes suggested by the editor, I like it as well. :)

test/Diagnostics/Metrics/Implementation/Meter3DTest.cs
ilijagrbic: Ah, apologies, the "Meter3D" went out of hand, fixed. :)

src/Diagnostics/Metrics/Implementation/IFabricMeter.cs
ilijagrbic: Good question. I think I like "variable" more than "custom", so I will make changes accordingly.

src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
ilijagrbic: I completely agree in principle with you @olegsych . Main issue is that we are asked to ship this (with all of the perf changes) by 31/3, for the next 11.5 RC. While adding a struct seems simple enough, I would have a higher level of expected testing that would need to be done in that case, since I remember that string marshalling was handled differently in the structs, especially for a variable length array. I'm not sure exactly how to implement that, I can't remember offhand, but I don't think it was outright supported or there was some complication regarding it. That was one of the reasons we chose to use a string array instead of a struct in the first place. And while I think all of this would be ultimately resolved, I'm not sure if we have bandwidth for this right now, especially since I'm on-call during the entire week. Also, it seems to me that every day we delay we also block perf changes further, which is the one task I'm most concerned about, as there are the most unknowns in it.

I particularly don't like this kind of last-minute decisions, as they could lead to long term costs for short term gains (in this case gains in time). But this change looks low risk to me, I have confidence it functionally works, since we are using the same Marshalling we already tested and know that it works. On the other hand, main issue is the number of parameter and extensibility of the design, which don't seem like blockers, especially for an internal API, only used by our team.

Let me know what you think @olegsych , I don't want to push in these changes at any cost, but if 11.5 is to be targeted, I think we will need the changes asap, otherwise we might have yet another uncomfortable delay for this feature.

edit: By the way, I have addressed the naming comments. :)

test/Diagnostics/Metrics/Implementation/Meter2DTest.cs
ilijagrbic: As this is a test improvement, I will take note and work on it in a separate PR.

src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
olegsych: Hey @ilijagrbic, I hear you, but I would push back on 11.5 because of the perf risk anyway. I won't throw a fit over this, but I wanted to see if Copilot could do it for us.

It actually did a pretty good job and I only had to correct it a few times and had it update the instructions to avoid these mistakes in the future.

I accidentally pushed the commit into your branch. I meant to create a PR instead, but now that it's there, you can review it in place and get rid of it if you prefer. I already pushed it into my own `users/olegsych/updateCreateMeterAPI` branch, so we can create a separate PR instead.

I'll try to generate a similar change in the native code.

olegsych: Here is the ADO PR https://dev.azure.com/msazure/One/_git/WindowsFabric/pullrequest/15206851

olegsych (approved):

src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
ilijagrbic: Hey @olegsych , thank you for the feedback and for working on this. I reverted the commit in this branch just to enable Priyal to build and integrate my branch locally, sorry for not explaining this on Friday, but it was really late and I just wanted to unblock her. 
I reviewed the Copilot changes, they are interested in a sense that they avoid the default Marshalling completely and just use `IntPtr`. Default marshalling of `string[]` was exactly what was causing issues for me. However, I feel this is not a small change, especially compared to this PR. Also, we wanted to avoid `IntPtr` in the past and use actual types whenever possible, so I think both options have tradeoffs. Anyways, to not go too long on this, I will merge this PR, and I will attend the SDK scrum, so we can discuss this further if needed.

PR #584, opened 2026-04-14
Split copilot-instructions.md and create review skill

olegsych:
Decompose the monolithic `copilot-instructions.md` into focused, scoped instruction files
to reduce per-session context noise and make each concern independently maintainable.

### New files
- `coding.instructions.md` — language-agnostic coding guidelines (`applyTo: **`)
- `csharp.instructions.md` — C#-specific rules (`applyTo: **/*.cs`)
- `interop.instructions.md` — COM interop conventions (`applyTo: **/*.cs`)
- `skills/review/SKILL.md` — code review skill with multi-model support
- `agents/{codex,gemini,opus}.agent.md` — sub-agent definitions for multi-model reviews

### Moved content
- Project structure table and build conventions → `CONTRIBUTING.md`
- Instruction file conventions → `knowledge.instructions.md`
- PR workflow guidance (commit guidelines, thread resolution, maintainer process) → `CONTRIBUTING.md`

### Other changes
- Narrow test instruction `applyTo` from `test/**` to `test/**/*.cs`
- Broaden `docs.instructions.md` scope to cover NuGet package descriptions and README files
- Add `applyTo` to `knowledge.instructions.md`
- Fix "For for" typo and heading style in `README.md`
- Reduce `copilot-instructions.md` to minimal discovery pointers


.github/instructions/interop.instructions.md
```diff
@@ -0,0 +1,17 @@
...
+
+- Product code must work identically on `net462` (legacy COM interop) and `net8.0+` (ComWrappers/`[GeneratedComInterface]`).
+  Prefer a single implementation over `#if`-separated code paths.
+- For COM method parameters that contain strings or string arrays, prefer passing a blittable struct via `IntPtr` rather
+  than using `[MarshalAs]` attributes. Marshalling attributes behave differently between legacy COM and ComWrappers,
+  but a blittable `IntPtr` passes through both runtimes without any marshalling.
+- Define native structs with `[StructLayout(LayoutKind.Sequential, Pack = 8)]` using only `IntPtr` and `uint` fields.
+  Include an `IntPtr Reserved` field for future extensibility (allowing `_EX1`, `_EX2` extensions).
+- Use `GCHandle.Alloc(value, GCHandleType.Pinned)` + `stackalloc` to pin strings and build pointer arrays on the caller
+  side. See `Meter.RecordViaNative` and `MeterProvider.CreateNativeMeter` for the canonical pattern.
```
ilijagrbic, line 16: `Meter`/`MeterProvider` implementations decisions were heavily influenced by the performance needs for that API (performance for the `Record()` call and more complex API needed for `CreateMeter()` to achieve this). I think it is fine to keep most of these instructions, with the clarification to apply the IntPtr/struct/GCHandle pattern only for the string arrays. But I think we will definitely need to iterate on this, in order to converge to an acceptable practice for a general-purpose API.

In particular, I don't think every API should use stack allocated `GCHandles`, unless it is needed from the performance perspective, especially for transferring singular strings, which `MarshalAs` would handle just fine, and there would be no overhead for calling such methods. I would also add a guideline to use structs in general, if number of parameters becomes large enough (>3-4) or when extensibility might be needed in the API.

CONTRIBUTING.md
```diff
@@ -114,7 +142,18 @@ sure you're building with latest package versions and assemblies.
...
+  - Title should use the imperative mood and be limited to 72 characters.
+  - Don't include textual tags `[MyFeature]` in the title.
 - Make sure the validation build completes successfully.
 - Link the pull request from the work item/issue you've created.
 - Publish the pull request and address the Copilot review feedback.
+- Resolve review comment threads after addressing the feedback.
 - Tag the person you were discussing your proposal to review the PR.
+
+## Maintainers
+- Ask Copilot `/review PR #N with multiple models and post feedback to GitHub`.
```
ilijagrbic, line 157: Is there a way to utilize the same approach (get feedback from Copilot with multiple models) automatically when requesting review from the Copilot coding agent on GitHub, eliminating the need to manually ask local Copilot for the review? Also, I would like to know which comments are human written vs AI generated, but as far as I see this is already the way the comments work with the :copilot: emoji at the beginning.

.github/skills/review/SKILL.md
```diff
@@ -0,0 +1,215 @@
...
+5. **Reconcile your assessment with the author's claims.** Where your independent reading of the code disagrees with
+   the PR description, investigate further — do not defer to the author's framing.
+6. **Build a thread inventory**: For each unresolved and resolved thread, record the file, line, severity, what was
+   requested, the thread ID, and whether it is resolved or unresolved. This is your verification checklist.
+7. **Update your assessment** if the additional context genuinely changes your evaluation. Do not soften findings just
+   because the PR description sounds reasonable.
+8. **Validate the PR title and description** based on the detailed analysis, make sure it is accurate and meets the
+   `CONTRIBUTING.md` guidelines. Report violation as a separate finding.
+
+## Step 3: Verify Addressed Feedback
```
ilijagrbic, line 100: This seems to be a step the agent would take after the initial pass/comments have been posted. Does this mean this skill should (ideally) be run in the same Copilot session, to preserve the original context/intent behind the posted comments?

.github/skills/review/SKILL.md
```diff
@@ -0,0 +1,215 @@
...
+2. **Consider collateral damage.** For every changed code path, brainstorm: what other scenarios, callers, or inputs
+   flow through this code? Could any of them break after this change? Surface plausible risks so the author can
+   evaluate.
+3. **Be specific and actionable.** Every comment should tell the author exactly what to change and why. Include
+   evidence of how you verified the issue is real.
+4. **Flag severity clearly:**
+   - `[❌](# "Must fix before merge")` — Bugs, security issues, test gaps for behavior changes.
+   - `[⚠️](# "Should fix")` — Missing validation, inconsistency with established patterns.
+   - `[💡](# "Consider changing")` — Readability wins, minor improvements.
+5. **Don't pile on.** If the same issue appears many times, flag it once with a note listing all affected
```
ilijagrbic, line 130: Perhaps we should add an exception to this point in case of some high severity **"Must fix"** issues? The way I see it, each security concern should be flagged.

.github/skills/review/SKILL.md
```diff
@@ -0,0 +1,215 @@
...
+   - If unsure, investigate further or surface it explicitly as a low-confidence question.
+   - Trust the author's context. If a pattern is consistent with the repo, assume it's intentional.
+   - Never assert that something "does not exist" or "is unavailable" based on training data alone. Ask rather
+     than assert.
+9. **Ensure code suggestions are valid.** Any code you suggest must be syntactically correct and complete.
+10. **Label in-scope vs. follow-up.** Distinguish between issues the PR should fix and out-of-scope improvements.
+
+## Multi-Model Review (Optional)
+
+When requested, use the `opus`, `gemini` and `codex` sub-agents to get diverse perspectives. Different models catch different
```
ilijagrbic, line 162: I wonder how different of a response would the same model provide when queries multiple times. If there is enough variability, perhaps it makes sense to include an option to run the review by using multiple models multiple times.

ilijagrbic (approved):
Thank you @olegsych , these are great changes. 
I left some comments/questions, none of which are blocking.

.github/instructions/interop.instructions.md
olegsych: It's a good point. We'll continue iterating. Plus these aren't new instructions. They were moved here from copilot-instructions.md

.github/skills/review/SKILL.md
olegsych: I don't think so, at least that's not how I tested it. It should be able to re-establish the context from scratch.

.github/skills/review/SKILL.md
olegsych: You don't think follow-up reviews would catch them? Most of this section is a verbatim copy from the dotnet/runtime skill, so I'm hesitant to change it without testing. Explosion of review comments was a real issue in testing.

.github/skills/review/SKILL.md
olegsych: There is definitely some variability because subsequent iterations reported additional issues. I think it should be already possible by changing the prompt because the argument-hint defined in the frontmatter doesn't have to be followed to the letter. For example, I've seen it posting feedback to GitHub when I said `and post response`.

CONTRIBUTING.md
olegsych: I've asked Copilot multiple times about this and the answer seems to be no. The dotnet/runtime repo has extra files that look like there may be undocumented instructions for the code review agent, but nothing official. I didn't want to copy&paste undocumented/unverified instructions.

olegsych: And yes, the review skill should be starting each review and each comment with the :copilot: emoji linked to the copilot code review docs.

olegsych:
Thanks @ilijagrbic Could you re-approve? I've made some changes to reduce verbosity of the reviews and try to prevent common mistakes I see Copilot make repeatedly.

ilijagrbic (approved):

.github/skills/review/SKILL.md
ilijagrbic: They probably will, that's a good point.

PR #587, opened 2026-04-21
Formalize, reorganize package dependencies, prepare for 12.0.0 release

olegsych:
- Simplify dependency analysis and prevent unintended application of transitive pins.
  - Split `Directory.Packages.props` into 3 files to separate common, `src/` and `test/` dependencies.
  - Separate framework-specific dependencies into conditional `<ItemGroups/>`.
  - Add end-of-line comments explaining non-trivial package versions.

- Minimize the dependency graph for consumers of Service Fabric packages.

**Instructions**
- Expand `nuget.instructions.md` (formerly `package.instructions.md`) with detailed guidance and tooling verified in this PR.
- Add `Get-TransitivePackage` helper to `NuGetHelpers.ps1`.
- Add NuGet MCP server to `.vscode/mcp.json`.
- Coding agent MCP configuration on GitHub has already been updated.

**Prepare Package References for 12.0.0 Release Based on New Guidance**

_Upgraded_
- `Newtonsoft.Json` `13.0.3` → `13.0.4` - Direct reference, latest stable.

_Downgraded_
- `Microsoft.Bcl.AsyncInterfaces` `10.0.5` → `8.0.0` - pin, test-only, 
- `System.Collections.Immutable` `10.0.5` → `8.0.0` - pin, net4-test-only, 
- `System.Diagnostics.DiagnosticSource` `10.0.5` → `8.0.1` - pin, **common,** net4-only

_Added_
- `System.Reflection.Metadata` `8.0.1` — pin, net4-test-only, 
- `System.Threading.Tasks.Extensions` `4.6.0` — pin, **common,** net4-only

_Removed_
- `System.Reflection.Emit` `4.7.0` — Unnecessary Remoting v1 reference
- `System.Reflection.Emit.ILGeneration` `4.7.0` — Unnecessary Remoting v1 reference
- `System.ValueTuple` `4.6.1` — Unnecessary pin

**Also**
- Clarify `pwsh.instructions.md` regarding naming of user-invokable functions like `Get-TransitivePackage`


test/Directory.Packages.props
```diff
@@ -0,0 +1,21 @@
...
+    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" /> <!-- x.v.c->6.0.0; S.T.J->8.0.0 -->
+    <PackageVersion Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="3.1.12" /> <!-- 3.2.2 breaks SemVer: S.C.I 8.0.0->9.0.8 -->
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="18.4.0" />
+    <PackageVersion Include="Moq" Version="4.20.72" />
+    <PackageVersion Include="xunit.v3" Version="3.2.2" />
+    <PackageVersion Include="xunit.v3.assert" Version="3.2.2" />
+    <PackageVersion Include="xunit.v3.extensibility.core" Version="3.2.2" />
+    <PackageVersion Include="xunit.runner.visualstudio" Version="3.1.5" />
+  </ItemGroup>
+  <ItemGroup Condition="$(TargetFramework.StartsWith('net4'))">
```
ilijagrbic, line 16: nit: I would be nice to set a property (let's say: NetFXTarget, NetCoreTarget etc.) depending on which target framework is used and then use that for conditions like this. It would be more readable and easier to maintain in the future.

eng/NuGetHelpers.ps1
```diff
@@ -38,3 +38,35 @@ function AddNuGetPackageSource([string] $configPath, [string] $key, [string] $pa
     Set-Content $configPath $content -NoNewline
     Write-Host "Updated $configPath"
 }
+
+function Get-TransitivePackage([Parameter(Mandatory)] [string] $package, [string] $project, [string] $framework) {
```
ilijagrbic, line 42: Perhaps I missed something with these helper methods, but to my knowledge, we have them in order to enable easier building of runtime and SDK together. I just haven't figured out what is the specific use-case for this function?

ilijagrbic (approved):
Thank you for the changes @olegsych , they look good. :)

eng/NuGetHelpers.ps1
olegsych: The new commandlet is explained in the `nuget.instructions.md`. Try running the examples, they are pretty self-explanatory once you see the output. 

I don't know if this is a good long-term place for it. This is the place copilot picked and wouldn't have been my first choice, but does makes sense from the standpoint of keeping the number of files minimal. I was thinking about converting the nuget.instructions.md into a nuget/SKILL.md and moving the commandlet into it.

test/Directory.Packages.props
olegsych: I go back and forth about MSBuild properties, specifically on whether they are better than simple literals given that MSBuild doesn't enforce correctness and properties are easily broken by typos. Are you thinking about conditions only in `Directory.Packages.props` or other MSBuild files too?

PR #591, opened 2026-05-10
[Armada] HTTP Code generation for PerNodePlacement

iYzeee23:
- **Summary**
  - Regenerates the Service Fabric HTTP client and generated HTTP PowerShell surface for opt-in node placement properties
  - This picks up the Swagger change that adds `IncludePlacementProperties` to the node query REST APIs and exposes the returned `NodeInfo.PlacementProperties` map in generated client models
- **Changes**
  - Added `includePlacementProperties` to generated node client methods:
     - `GetNodeInfoListAsync`
     - `GetNodeInfoAsync`
  - Updated generated HTTP requests to send `IncludePlacementProperties`
  - Added `NodeInfo.PlacementProperties` as `IReadOnlyDictionary<string, string>`
  - Updated `NodeInfoConverter` to deserialize and serialize the placement properties dictionary
  - Added `IncludePlacementProperties` to generated `Get-SFNode` HTTP PowerShell
  - Updated generated HTTP PowerShell help XML
- **Notes**
  - This is generated code only. The runtime behavior and Swagger source are handled in the corresponding Service Fabric runtime and Swagger PRs
  - `PlacementProperties` is populated only when callers opt in with `IncludePlacementProperties=true`; existing callers remain source-compatible because the new generated parameter has a default value
- **SF and Swagger**
  - [WindowsFabric](https://dev.azure.com/msazure/One/_git/WindowsFabric/pullrequest/15490913)
  - [Swagger](https://dev.azure.com/msazure/One/_git/SF-Swagger/pullrequest/15686092)


src/Client.Http/Client/Generated/INodeClient.cs
```diff
@@ -48,6 +51,7 @@ Task<PagedData<NodeInfo>> GetNodeInfoListAsync(
             ContinuationToken continuationToken = default(ContinuationToken),
             NodeStatusFilter? nodeStatusFilter = NodeStatusFilter.Default,
             long? maxResults = 0,
+            bool? includePlacementProperties = false,
```
olegsych: Why do we need an option to include placement properties?

Discuss here: https://msazure.visualstudio.com/DefaultCollection/One/_git/WindowsFabric/pullRequest/15490913?discussionId=243673147#1779076171

olegsych: We'll need to regenerate this PR when/if the Swagger is updated.

olegsych (commented):
We'll need to close on bool vs enum before merging

olegsych (approved):

src/Client.Http/Client/Generated/INodeClient.cs
olegsych: The implementation has been updated. Once the Swagger is updated, this PR needs to be re-generated.

iYzeee23: @olegsych Updated the DotNet PR as well. Please take a look

olegsych (changes requested):
Asked a question about this on the Swagger PR - why are we defining options as an int? I expected it to be an enum

olegsych (approved):

PR #596, opened 2026-05-22
Change IFabricMeterProvider interface version

ilijagrbic:
This is a proposal for a solution regarding the [incident ](https://portal.microsofticm.com/imp/v5/incidents/details/802511150/summary)

Short summary of the issue:
- Customer app build with SDK 11.4 contains a old version of **IFabricMeterProvider** interface. This version passes a string as the first parameter - `CreateMeter(IntPtr metricNamespace, IntPtr name, uint count, IntPtr dimensionNames);`
- FabricTelemetry.dll implementation for the same interface has a diferent signiture of the method - `CreateMeter(IntPtr meterDescription);`
- String that is passed as the namespace gets interpreted as a struct, and dereferences pointers based on the raw binary value of the string - causing Access Violation exception, crashing the user process.

Proposed fix:
- Instead of changing the method signature, introduce a new method - `CreateMeter2(IntPtr meterDescription);`, while the implementation for the old method will throw proper exceptions, ensuring we catch them.
- This change will be coupled with an [appropriate change in FabricTelemetry.dll](https://dev.azure.com/msazure/One/_git/WindowsFabric/pullrequest/15838625)

### DIFF
--- a/src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
+++ b/src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
@@ -16,14 +16,19 @@ namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation;
 [GeneratedComInterface]
 [Guid("89462876-f11e-41c6-bd99-c933b46c5e66")]
 [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
-partial interface IFabricMeterProvider
+partial interface IFabricMeterProvider2
 {
-    // Vtable slot 3: inherited from old IFabricMeterProvider (GUID 15AD37D2-F641-4188-824B-0D68CB4F6C17).
-    // Declared for vtable alignment only; not called by managed code.
+    #region IFabricMeterProvider
+
     [return: MarshalUsing(typeof(UniqueComInterfaceMarshaller<IFabricMeter>))]
     IFabricMeter CreateMeter(IntPtr metricNamespace, IntPtr name, uint count, IntPtr dimensionNames);
 
-    // Vtable slot 4: new method on IFabricMeterProvider2.
+    #endregion
+
+    #region IFabricMeterProvider2
+
     [return: MarshalUsing(typeof(UniqueComInterfaceMarshaller<IFabricMeter>))]
     IFabricMeter CreateMeter2(IntPtr meterDescription);
+
+    #endregion
 }
--- a/src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
+++ b/src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
@@ -14,10 +14,16 @@ using GeneratedComInterfaceAttribute = System.Runtime.InteropServices.ComImportA
 namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation;
 
 [GeneratedComInterface]
-[Guid("15AD37D2-F641-4188-824B-0D68CB4F6C17")]
+[Guid("89462876-f11e-41c6-bd99-c933b46c5e66")]
 [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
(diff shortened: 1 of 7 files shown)


src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
```diff
@@ -14,10 +14,16 @@
...
-[Guid("15AD37D2-F641-4188-824B-0D68CB4F6C17")]
+[Guid("89462876-f11e-41c6-bd99-c933b46c5e66")]
 [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
 partial interface IFabricMeterProvider
 {
+    // Vtable slot 3: inherited from old IFabricMeterProvider (GUID 15AD37D2-F641-4188-824B-0D68CB4F6C17).
+    // Declared for vtable alignment only; not called by managed code.
     [return: MarshalUsing(typeof(UniqueComInterfaceMarshaller<IFabricMeter>))]
-    IFabricMeter CreateMeter(IntPtr meterDescription);
+    IFabricMeter CreateMeter(IntPtr metricNamespace, IntPtr name, uint count, IntPtr dimensionNames);
```
olegsych, line 24: If we're doing this, we shouldn't have the IFabricMeterProvider2 inherit from IFabricMeterProvider, it should be a clean interface with a new guid.

But before we do this, let's make sure this is really what the leadership wants to do.

olegsych (changes requested):

src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
ilijagrbic: Hey @olegsych , I think this is what Gaurav pretty much explicitly said wants this to be done. I'm only concerned about if it works at this point.

About the interface inheritance - I will open a PR for 11/main with the same changes and add you there so we can discuss. I assume this is possible because C++ supports multiple inheritance, and we would be able to implement both interfaces in a single COM object. If this is not the case, I try to avoid implementing a different COM object for the new interface.

ilijagrbic: Also, to my understanding, we would also need to change the FabricCreateMeterProvider P/Invoke function, since that one returns a `IFabricMeterProvider`. If the only reason is to keep the code clean, I would reconsider the priority of that.

src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
```diff
@@ -14,10 +14,16 @@
 namespace Microsoft.ServiceFabric.Diagnostics.Metrics.Implementation;
 
 [GeneratedComInterface]
-[Guid("15AD37D2-F641-4188-824B-0D68CB4F6C17")]
+[Guid("89462876-f11e-41c6-bd99-c933b46c5e66")]
 [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
 partial interface IFabricMeterProvider
```
olegsych: We should maintain consistent naming - `IFabricMeterProvider2`

olegsych (changes requested):

src/Diagnostics/Metrics/Implementation/IFabricMeterProvider.cs
olegsych: Also, we should use comments/regions consistent with what we do in the System.Fabric code.

https://msazure.visualstudio.com/One/_git/WindowsFabric?path=/src/prod/src/managed/Api/src/System/Fabric/Interop/NativeClient.cs&version=GBdevelop&_a=contents

i.e. the VTable slot comments are replaced by regions.

ilijagrbic: Thank you @olegsych , I have addressed this. :)

olegsych (approved):

PR #603, opened 2026-06-02
Fix double wrapping in AggregateException in WCF

gobradovic:
### Problem

When a service operation invoked over WCF remoting threw an `AggregateException`,
the client received it wrapped in an *additional* `AggregateException`. The caller
had to unwrap two layers (`AggregateException` → `AggregateException` → actual
exception) instead of one, which is inconsistent with how the non-WCF (Fabric
transport) remoting client behaves and breaks code that relies on a single
`Flatten()` / `InnerException` access.

### Root cause

In `WcfServiceRemotingClient.RequestResponseAsync`, the
`FaultException<RemoteException2>` handler unconditionally wrapped the
deserialized exception:

```csharp
throw new AggregateException(exception);
```

When `exception` was already an `AggregateException` (because the service-side
threw one and it round-tripped through `RemoteException2` deserialization), this
produced the double-wrapping.

### Fix

Re-throw the deserialized exception as-is when it is already an
`AggregateException`; otherwise wrap it as before. This matches the pattern
already used by `ExceptionDeserializer.DeserializeRemoteExceptionAndThrowAsync`
in the Fabric transport path.

```csharp
throw exception is AggregateException aggregateException
    ? aggregateException
    : new AggregateException(exception);
```

### Test

Added `WcfServiceRemotingClientTest.RequestResponseAsync.DoesNotDoubleWrapAggregateException`,
which serializes an `AggregateException(NotImplementedException)` through the
same `ExceptionSerializer` the runtime uses on the wire, feeds it into the
client via a faulted `FaultException<RemoteException2>`, and asserts the caught
`AggregateException` has exactly one inner `NotImplementedException` — proving
no extra wrapping layer was added.

Also fixed a pre-existing typo in the enclosing nested test class name
(`RequestResnposeAsync` → `RequestResponseAsync`).

### Compatibility

Behavior changes only for the previously broken case (remote
`AggregateException`). All other exception types continue to be wrapped exactly
as before, so callers of WCF remoting that already handled `AggregateException`
with `Flatten()` are unaffected.


test/Services.Wcf/WcfServiceRemotingClientTest.cs
```diff
@@ -114,6 +114,36 @@ public async Task ThrowsServiceExceptionForUnknownExceptions()
...
+
+                IServiceRemotingRequestMessage requestMessageMock = Mock.Of<IServiceRemotingRequestMessage>();
+
+                Mock.Get(requestMessageMock)
+                    .Setup(m => m.GetHeader())
+                    .Throws(faultException);
+
+                AggregateException exception = await Assert.ThrowsAsync<AggregateException>(() => sut.RequestResponseAsync(requestMessageMock));
+                Exception innerException = Assert.Single(exception.InnerExceptions);
+                Assert.IsType<NotImplementedException>(innerException);
```
olegsych, line 144: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") `Assert.IsType<T>()` contradicts [`.github/instructions/test.instructions.md`](.github/instructions/test.instructions.md): _"Don't use `Assert.IsAssignableFrom<T>()` or `Assert.IsType<T>()`. Compiler verification of inheritance is sufficient in most cases… To test SUT's inherited APIs, use type casting instead."_

Since the test cares about the exact deserialized type:

```suggestion
                Assert.Equal(typeof(NotImplementedException), innerException.GetType());
```

The pre-existing `Assert.IsType<T>` usages elsewhere in this file are out of scope.

src/Services.Wcf/Remoting/V2/Wcf/Client/WcfServiceRemotingClient.cs
```diff
@@ -119,7 +119,9 @@ public async Task<IServiceRemotingResponseMessage> RequestResponseAsync(IService
                 RemoteException2 remoteException2 = faultException.Detail;
 
                 Exception exception = exceptionDeserializer.ConvertRemoteException(remoteException2);
-                throw new AggregateException(exception);
+                throw exception is AggregateException aggregateException
+                    ? aggregateException
```
olegsych, line 123: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [📝](# "Note") The fix relies on `ExceptionDeserializer.ConvertRemoteException` always returning either an `AggregateException` or something that can be wrapped in one — behavior that isn't documented on `ConvertRemoteException`. If a future change in the deserializer stops auto-wrapping, this site will silently switch to "no wrap" for non-aggregate exceptions and break callers that expect an `AggregateException`. Consider a short comment here referencing `ConvertRemoteException`'s wrapping contract, or extracting a shared helper used by both deserialization paths. Not blocking — the fix as-is matches the existing pattern in `DeserializeRemoteExceptionAndThrowAsync`.

test/Services.Wcf/WcfServiceRemotingClientTest.cs
```diff
@@ -114,6 +114,36 @@ public async Task ThrowsServiceExceptionForUnknownExceptions()
...
+                RemoteException2 remoteException = exceptionSerializer.BuildRemoteException(
+                    new AggregateException(new NotImplementedException(errorMessage)));
+
+                var faultException = new FaultException<RemoteException2>(remoteException);
+
+                IServiceRemotingRequestMessage requestMessageMock = Mock.Of<IServiceRemotingRequestMessage>();
+
+                Mock.Get(requestMessageMock)
+                    .Setup(m => m.GetHeader())
+                    .Throws(faultException);
```
olegsych, line 140: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") The new `Setup()` chain isn't assigned to a discard `_`, which contradicts [`.github/instructions/moq.instructions.md`](.github/instructions/moq.instructions.md): _"Assign `Setup()` chains to a discard `_` variable. This explains that the expression value is not used intentionally and suppresses the `IDE0058` warnings."_

```suggestion
                _ = Mock.Get(requestMessageMock)
                    .Setup(m => m.GetHeader())
                    .Throws(faultException);
```

The two pre-existing instances earlier in this file are out of scope per the "Limit scope of changes" guideline; only this new code is in scope.

olegsych (approved):
[:copilot:](https://docs.github.com/copilot/responsible-use/code-review) **✅ Looks Good** — The fix correctly resolves the double-wrapping of `AggregateException` in the WCF remoting path by mirroring the existing pattern in `DeserializeRemoteExceptionAndThrowAsync`. CI is green and a focused regression test exercises the real serializer round trip. Inline comments capture non-blocking suggestions for test hygiene and a docs/clarity note.

### Detailed Assessment

**Motivation.** `WcfServiceRemotingClient.RequestResponseAsync` caught `FaultException<RemoteException2>`, called `ExceptionDeserializer.ConvertRemoteException`, then unconditionally rethrew as `new AggregateException(exception)`. `ConvertRemoteException` already normalizes its result with `ex is AggregateException ? ex : new AggregateException(ex)`. When the remote side threw an `AggregateException`, callers observed `AggregateException → AggregateException → original`. The non-WCF path (`DeserializeRemoteExceptionAndThrowAsync`) did not have this bug, so WCF was the outlier.

**Approach.** Add a conditional throw: if `ConvertRemoteException` already returned an `AggregateException`, rethrow it as-is; otherwise wrap. Minimal correct change, no API change, parity with the non-WCF path. The new test `DoesNotDoubleWrapAggregateException` exercises the real serializer round trip (`ExceptionSerializer.BuildRemoteException` → `FaultException<RemoteException2>` → `ConvertRemoteException`) rather than mocking the deserializer, which would have hidden the bug. `Assert.Single(exception.InnerExceptions)` is the right "no double wrap" assertion. The bundled rename of the nested test class `RequestResnposeAsync` → `RequestResponseAsync` brings the class name into compliance with the nested-test-naming rule in [`.github/instructions/test.instructions.md`](.github/instructions/test.instructions.md) and is appropriately scoped because the same class is being substantially modified.

**Cost-benefit / Risk.** Tiny code change, low risk. Existing tests (`ThrowsActualExceptionForKnownExceptions`, `ThrowsServiceExceptionForUnknownExceptions`) continue to pass because their payloads are not `AggregateException`. A regression would be caught by `Assert.Single(exception.InnerExceptions)`. All CI checks green.

### Issues

#### [⚠️](# "Should fix") PR description — empty body violates [`CONTRIBUTING.md`](CONTRIBUTING.md)

The PR body is empty. `CONTRIBUTING.md` under "All Contributors" says: _"Follow the git commit guidelines to author the PR title and description. Describe what it changes and why, not how it was developed or which prior PRs it addresses."_ The title is fine (imperative, ≤72 chars, no tags), but please add a brief description:

1. **What:** WCF remoting path double-wraps `AggregateException` because `ExceptionDeserializer.ConvertRemoteException` already wraps non-aggregate exceptions.
2. **Why:** It produces `AggregateException → AggregateException → original`, breaking callers that handle `AggregateException` once via `.Flatten()` / `InnerException`.

#### [📝](# "Note") Related pattern in `CommunicationClientFactoryBase`

[`src/Services/Communication/Client/CommunicationClientFactoryBase.cs`](src/Services/Communication/Client/CommunicationClientFactoryBase.cs#L499) has `throw new AggregateException(actualException);` where `actualException` is the original caught exception (not normalized by `ConvertRemoteException`). If the underlying operation already throws an `AggregateException` (e.g., from a `Task.Wait()` upstream), this site can produce the same double-wrap shape. Out of scope for this PR — flagging for the author to confirm and file separately if appropriate.

### ✅ Things that look right

- New code is minimal, ternary-only.
- The new test exercises the real round trip rather than mocking the deserializer, which would have hidden the bug.
- `Assert.Single(exception.InnerExceptions)` is the right "no double wrap" assertion.
- Rename `RequestResnposeAsync` → `RequestResponseAsync` aligns the nested test class name with the SUT-member naming rule and is appropriately scoped.
- PR title is imperative, ≤72 chars, no tags.
- All CI checks green.

@gobradovic — please take a look at the inline suggestions and consider adding a PR description before merge.

test/Services.Wcf/WcfServiceRemotingClientTest.cs
olegsych: Nah, this is not the right suggestion. The reason why Assert.IsType shouldn't be used is because it doesn't provide complete test coverage. Normally, we would want to assert on the exception identity, but since it's not preserved through the serialization/deserialization process, we'd assert on expected Message after downcasting to the NotImplementedException. Optional.

test/Services.Wcf/WcfServiceRemotingClientTest.cs
olegsych: Optional. We'll take care of this when we dotnet format this code.

olegsych:
@gobradovic The PR looks good, the only thing I would ask you to address is the PR descriptions. All comments are optional.

PR #609, opened 2026-06-18
Add APIs to disable and enable services

vpavlovicMSFT:
Disabling a service scales it down to zero replicas or instances while
keeping its identity, so it can be enabled again later. Today that
only works through the native API - this adds the HTTP and PowerShell
forms so it can be done from any client.

I'm putting this up next to two sibling PRs, the runtime change and the
REST spec, so the public API shape gets reviewed as a whole instead of
drifting apart between surfaces.

On the create side I replaced the IsCreateAsDisabled boolean with a
ServiceOptions flags enum (None, InitiallyDisabled). That enum is the
part worth a close look - it's the extension point for future creation
options, so it's better to get it right now than to break the signature
later.

The C# is generated from the SF-Swagger spec, so the API design itself
lives there - the spec PR is the right place to review shape and naming,
not the generated code here.

Related:
- Runtime: https://msazure.visualstudio.com/One/_git/WindowsFabric/pullrequest/16134176
- Spec: https://msazure.visualstudio.com/One/_git/SF-Swagger/pullrequest/15062355
- Dotnet: https://github.com/microsoft/service-fabric-dotnet/pull/609
- Docs: https://learn.microsoft.com/en-us/azure/service-fabric/disable-enable-service


vpavlovicMSFT (commented):
[:copilot:](https://docs.github.com/copilot/responsible-use/code-review) **✅ Looks Good**

The PR updates internal Service Fabric types by adding \EnableService\ and \DisableService\ HTTP/Powershell APIs. Since all the changes are within \Generated/\ folders, the C# code accurately reflects the upstream specification updates. As noted by the author, these files are auto-generated from the SF-Swagger spec, which is reviewed separately.

### Detailed Assessment

- **Motivation**: This PR introduces the generated API surface (both HTTP client and PowerShell cmdlets) for the capability to scale a service to zero without deleting it. This completes the feature mapping from internal / native implementations to the HTTP/PowerShell APIs.
- **Approach**: All modified C# and XML code are artifacts generated from the Swagger specification.
- **Cost-benefit/Risk Analysis**: Minimal risk within this repository since the changes are purely generated scaffolding. Any design issues with the extensions belong to the swagger specification review.

### Issues

No issues found in the generated code.

vpavlovicMSFT (commented):
[:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") **Needs Changes** — well-scoped, CI-green generated-client change that faithfully mirrors the Service Fabric scale-to-zero feature (api-version, error codes, status values, and the destructive-cmdlet pattern all check out). Two API-shape / wire-behavior items below should be corrected in the SF-Swagger spec and regenerated before shipping; two items initially considered blockers were cleared after verification.

_Note: this consolidated review supersedes the earlier automated "✅ Looks Good" comment on this PR, which was posted before the deeper cross-check below._

## ⚠️ Should Fix

### 1. `ServiceOptions` is exposed as raw `int?`, not the intended typed flags enum
The description says `IsCreateAsDisabled` was replaced with a `ServiceOptions` flags enum (`None`, `InitiallyDisabled`), but the generated C# never defines the enum — the property, both `*ServiceDescription` constructors, the serializers, and `New-SFService -ServiceOptions` are all `int?`:
- `src/Client.Http/Common/Generated/ServiceDescription.cs` — `int? serviceOptions = 0`, `public int? ServiceOptions { get; }`
- `src/Client.Http/Common/Generated/StatefulServiceDescription.cs`, `.../StatelessServiceDescription.cs`
- `src/Powershell.Http/Generated/NewServiceCmdlet.cs` — `public int? ServiceOptions { get; set; } = 0;`

Consumers must pass magic numbers (`-ServiceOptions 1`) with no type safety or discoverability of `InitiallyDisabled`, and shipping `int?` now then promoting it to an enum later is itself the breaking change called out in the description. `DisableServiceFlag` in this same PR *is* generated as a proper enum + converter, so the two new flag concepts are modeled inconsistently. **Suggest** defining the flags enum in the SF-Swagger spec so the generator emits a typed enum + converter here.

### 2. `serviceOptions` defaults to `0` (always serialized) instead of `null`
Every other optional description field defaults to `null` and is omitted from the wire, but `serviceOptions` defaults to `0` and the converters write it whenever it is non-null (`if (obj.ServiceOptions != null) ...` in `StatefulServiceDescriptionConverter.cs` / `StatelessServiceDescriptionConverter.cs`). So **every** `CreateService` request — including those still using `api-version=6.0` — now emits `"ServiceOptions": 0`, pushing a new field onto older clusters and diverging from the established optional-field pattern. **Suggest** defaulting to `null` in the description constructors and `NewServiceCmdlet` so omission is the default; callers can still pass `0` explicitly.

## 📝 Notes

- **Positional / binary-compat:** `serviceOptions` is inserted mid-list (after `serviceTags`) in the public derived constructors and shifts `Flags` / later positional params in `New-SFService`. Adding an optional ctor param is a binary break and can rebind positional callers — but this is in-policy for a `12.0.0-beta` major pre-release and matches prior regenerations (e.g. the `serviceTags` consolidation). Worth confirming it's acceptable under the package's SemVer policy; any compat-preserving fix belongs in the spec/generator.
- **Dead code:** `src/Client.Http/Client.Http/Generated/Serialization/DisableServiceFlagConverter.cs` is unused — the flag is sent via `.ToString()` in the query string and never deserialized. Benign generator artifact.
- **Typo:** the new `DisableServiceAsync` / `EnableServiceAsync` `<exception>` comments in `IServiceClient.cs` say "a flag _wether_ the exception is transient" (→ "whether"). Pre-existing boilerplate.

## ✅ Investigated and cleared — no action
- **Disable transport (query vs. body):** `DisableServiceAsync` sends `DisableServiceFlag` / `ForceDisable` as query params with no request body. This matches the documented sibling **Delete Service** operation (query `ForceRemove`, bodyless POST), and the spec emits no `DisableServiceDescription` body model, so it is consistent with convention. A maintainer with internal SF-Swagger access could confirm the operation models these as `in: query`.
- **`New-SFService` help XML:** `-ServiceOptions` is absent from the `<command:syntax>` block, but so are all common (non-parameter-set-specific) optional parameters; the diff shows a full position-renumber cascade (i.e. full regeneration), and the cmdlet-vs-help position offset is a uniform, pre-existing convention. No defect.

vpavlovicMSFT:
[:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") **Needs Changes** — well-scoped, CI-green generated-client change that faithfully mirrors the Service Fabric scale-to-zero feature (api-version, error codes, status values, and the destructive-cmdlet pattern all check out). Two API-shape / wire-behavior items below should be corrected in the SF-Swagger spec and regenerated before shipping; two items initially considered blockers were cleared after verification.

_Note: this consolidated review supersedes the earlier automated "✅ Looks Good" comment on this PR, which was posted before the deeper cross-check below._

## ⚠️ Should Fix

### 1. `ServiceOptions` is exposed as raw `int?`, not the intended typed flags enum
The description says `IsCreateAsDisabled` was replaced with a `ServiceOptions` flags enum (`None`, `InitiallyDisabled`), but the generated C# never defines the enum — the property, both `*ServiceDescription` constructors, the serializers, and `New-SFService -ServiceOptions` are all `int?`:
- `src/Client.Http/Common/Generated/ServiceDescription.cs` — `int? serviceOptions = 0`, `public int? ServiceOptions { get; }`
- `src/Client.Http/Common/Generated/StatefulServiceDescription.cs`, `.../StatelessServiceDescription.cs`
- `src/Powershell.Http/Generated/NewServiceCmdlet.cs` — `public int? ServiceOptions { get; set; } = 0;`

Consumers must pass magic numbers (`-ServiceOptions 1`) with no type safety or discoverability of `InitiallyDisabled`, and shipping `int?` now then promoting it to an enum later is itself the breaking change called out in the description. `DisableServiceFlag` in this same PR *is* generated as a proper enum + converter, so the two new flag concepts are modeled inconsistently. **Suggest** defining the flags enum in the SF-Swagger spec so the generator emits a typed enum + converter here.

### 2. `serviceOptions` defaults to `0` (always serialized) instead of `null`
Every other optional description field defaults to `null` and is omitted from the wire, but `serviceOptions` defaults to `0` and the converters write it whenever it is non-null (`if (obj.ServiceOptions != null) ...` in `StatefulServiceDescriptionConverter.cs` / `StatelessServiceDescriptionConverter.cs`). So **every** `CreateService` request — including those still using `api-version=6.0` — now emits `"ServiceOptions": 0`, pushing a new field onto older clusters and diverging from the established optional-field pattern. **Suggest** defaulting to `null` in the description constructors and `NewServiceCmdlet` so omission is the default; callers can still pass `0` explicitly.

## 📝 Notes

- **Positional / binary-compat:** `serviceOptions` is inserted mid-list (after `serviceTags`) in the public derived constructors and shifts `Flags` / later positional params in `New-SFService`. Adding an optional ctor param is a binary break and can rebind positional callers — but this is in-policy for a `12.0.0-beta` major pre-release and matches prior regenerations (e.g. the `serviceTags` consolidation). Worth confirming it's acceptable under the package's SemVer policy; any compat-preserving fix belongs in the spec/generator.
- **Dead code:** `src/Client.Http/Client.Http/Generated/Serialization/DisableServiceFlagConverter.cs` is unused — the flag is sent via `.ToString()` in the query string and never deserialized. Benign generator artifact.
- **Typo:** the new `DisableServiceAsync` / `EnableServiceAsync` `<exception>` comments in `IServiceClient.cs` say "a flag _wether_ the exception is transient" (→ "whether"). Pre-existing boilerplate.

## ✅ Investigated and cleared — no action
- **Disable transport (query vs. body):** `DisableServiceAsync` sends `DisableServiceFlag` / `ForceDisable` as query params with no request body. This matches the documented sibling **Delete Service** operation (query `ForceRemove`, bodyless POST), and the spec emits no `DisableServiceDescription` body model, so it is consistent with convention. A maintainer with internal SF-Swagger access could confirm the operation models these as `in: query`.
- **`New-SFService` help XML:** `-ServiceOptions` is absent from the `<command:syntax>` block, but so are all common (non-parameter-set-specific) optional parameters; the diff shows a full position-renumber cascade (i.e. full regeneration), and the cmdlet-vs-help position offset is a uniform, pre-existing convention. No defect.

vpavlovicMSFT:
> ## ⚠️ Should Fix
> ### 1. `ServiceOptions` is exposed as raw `int?`, not the intended typed flags enum
> The description says `IsCreateAsDisabled` was replaced with a `ServiceOptions` flags enum (`None`, `InitiallyDisabled`), but the generated C# never defines the enum — the property, both `*ServiceDescription` constructors, the serializers, and `New-SFService -ServiceOptions` are all `int?`:
> 
> * `src/Client.Http/Common/Generated/ServiceDescription.cs` — `int? serviceOptions = 0`, `public int? ServiceOptions { get; }`
> * `src/Client.Http/Common/Generated/StatefulServiceDescription.cs`, `.../StatelessServiceDescription.cs`
> * `src/Powershell.Http/Generated/NewServiceCmdlet.cs` — `public int? ServiceOptions { get; set; } = 0;`
> 
> Consumers must pass magic numbers (`-ServiceOptions 1`) with no type safety or discoverability of `InitiallyDisabled`, and shipping `int?` now then promoting it to an enum later is itself the breaking change called out in the description. `DisableServiceFlag` in this same PR _is_ generated as a proper enum + converter, so the two new flag concepts are modeled inconsistently. **Suggest** defining the flags enum in the SF-Swagger spec so the generator emits a typed enum + converter here.
> ### 2. `serviceOptions` defaults to `0` (always serialized) instead of `null`
> Every other optional description field defaults to `null` and is omitted from the wire, but `serviceOptions` defaults to `0` and the converters write it whenever it is non-null (`if (obj.ServiceOptions != null) ...` in `StatefulServiceDescriptionConverter.cs` / `StatelessServiceDescriptionConverter.cs`). So **every** `CreateService` request — including those still using `api-version=6.0` — now emits `"ServiceOptions": 0`, pushing a new field onto older clusters and diverging from the established optional-field pattern. **Suggest** defaulting to `null` in the description constructors and `NewServiceCmdlet` so omission is the default; callers can still pass `0` explicitly.

On # 1 (typed enum): `ServiceOptions` follows the exact same pattern SF uses for every REST bitmask flag, please see `HealthStateFilter` (`Default`/`None`/`Ok`/`Warning`/`Error`/`All`): a documented `integer` that generates as `int?` and preserves bitwise composition. There are no typed flags enums anywhere in the generated HTTP client, so modeling `ServiceOptions` as a typed enum would be unprecedented in the SF REST surface; it would break bitwise composition (or require a gateway rewrite), **and isn't producible by the codegen for integer enums anyway**. So `int?` is the right choice here and I'm keeping it.

On # 2 (default `0`): good catch, fixed in cf0a930. The body-property bitmasks like `ApplicationHealthStateFilter` do default to `0`, but on the PowerShell surface every bitmask cmdlet parameter (and every other `New-SFService` parameter) is initializer-free, so `-ServiceOptions = 0` was the lone outlier. I've dropped `default: 0` from the spec and regenerated, so `ServiceOptions` is now a plain `int?` that's omitted from the wire when unset, no more `"ServiceOptions": 0` on every create.

olegsych (approved):

vpavlovicMSFT:
@olegsych Related PRs have all been merged, can we close this one as well?

PR #610, opened 2026-06-26
Add InternalsVisibleTo for FabricDCA in Microsoft.ServiceFabric.Diagnostics

junkaigao-ms:
Adds InternalsVisibleTo for FabricDCA (prod + test keys) to the Microsoft.ServiceFabric.Diagnostics assembly (src/Diagnostics/AssemblyInfo.cs), granting FabricDCA access to the internal metrics API. Mirrors the existing entries for FabricBRS, Microsoft.ServiceFabric.Data.Impl, and FabricIS.parallel.


olegsych (changes requested):
@junkaigao-ms Could you update the PR title/description to be more concise and accurate?

We don't need to know which metrics FabricDCA will be implementing here, but we do need to know to which assembly we're adding the InternalsVisibleTo attributes.

junkaigao-ms:
> @junkaigao-ms Could you update the PR title/description to be more concise and accurate?
> 
> We don't need to know which metrics FabricDCA will be implementing here, but we do need to know to which assembly we're adding the InternalsVisibleTo attributes.

updated. does it look correct now?

olegsych (approved):

PR #612, opened 2026-07-01
Add `IHostBuilder` overload for `UseServiceFabricIntegration`

gobradovic:
On .NET 10, `WebHostBuilder`, `IWebHost`, and `IWebHostBuilder` are
obsolete (ASPDEPR004, ASPDEPR008). `KestrelCommunicationListener` already
supported building an `IHost`, but Service Fabric integration
(`UseServiceFabricIntegration`) was only available through the
`IWebHostBuilder` path. This left modern host-based apps without a
warning-free way to configure Service Fabric middleware and unique service
URLs.

## Fix
Added `UseServiceFabricIntegration` for `IHostBuilder` in the new
`IHostBuilderExtensions` class.

The new host-based extension mirrors existing Service Fabric integration
behavior:

- Validates `hostBuilder` and `listener` arguments (fail fast).
- Uses `hostBuilder.Properties` for idempotency (generic-host equivalent
  of `UseSetting`/`GetSetting`).
- Applies `UseUniqueServiceUrl` to the listener when requested.
- Registers `ServiceFabricSetupFilter` through `ConfigureServices`.

The existing `IWebHostBuilder` extension path remains in place for
compatibility.


test/AspNetCore/HostBuilderServiceFabricExtensionTests.cs
```diff
@@ -0,0 +1,124 @@
...
+using Microsoft.ServiceFabric.Services.Communication.AspNetCore;
+using Moq;
+using Xunit;
+
+namespace Microsoft.ServiceFabric.AspNetCore.Tests
+{
+    /// <summary>
+    /// Test class for WebHostBuilderServiceFabricExtension with IHostBuilder.
+    /// </summary>
+    public class HostBuilderServiceFabricExtensionTests
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") This adds a second top-level test class for the existing product type `WebHostBuilderServiceFabricExtension`, which already has `WebHostBuilderServiceFabricExtensionTests`. `test.instructions.md` requires one test class per top-level product type, with each overloaded member in a separate nested class disambiguated by its BCL parameter type (e.g. `IWebHostBuilder`/`IHostBuilder`). Consider folding this coverage into the existing test class as a nested class. The pre-existing sibling already deviates, so a clean fix touches both files — a judgment call for the human reviewer.

test/AspNetCore/HostBuilderServiceFabricExtensionTests.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// Used by tests to check if services were configured by WebHostBuilderServiceFabricExtension.UseServiceFabricIntegration.
+        /// </summary>
+        private int configureServicesCount;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="HostBuilderServiceFabricExtensionTests"/> class.
+        /// </summary>
+        public HostBuilderServiceFabricExtensionTests()
+        {
+            var context = TestMocksRepository.GetMockStatelessServiceContext();
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") `var` hides the declared type `StatelessServiceContext`, which isn't obvious from the right-hand side. `csharp.instructions.md` and `.editorconfig` say to specify the type explicitly when it isn't apparent from the RHS.

test/AspNetCore/HostBuilderServiceFabricExtensionTests.cs
```diff
@@ -0,0 +1,124 @@
...
+            this.builder = new Mock<IHostBuilder>();
+            this.builder.SetupGet(_ => _.Properties).Returns(new Dictionary<object, object>());
+            this.builder
+                .Setup(_ => _.ConfigureServices(It.IsAny<Action<HostBuilderContext, IServiceCollection>>()))
+                .Callback<Action<HostBuilderContext, IServiceCollection>>(action =>
+                {
+                    this.capturedConfigureServices = action;
+                    this.configureServicesCount++;
+                })
+                .Returns(this.builder.Object);
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") These Moq `SetupGet`/`Setup` chains are used as flat expression statements. `moq.instructions.md` says to assign `Setup()` chains to a `_` discard to make the intentional discard explicit and suppress unused-value warnings. Prefix both the `SetupGet(...)` line and the `this.builder.Setup(...)` chain with `_ =`.

test/AspNetCore/HostBuilderServiceFabricExtensionTests.cs
```diff
@@ -0,0 +1,124 @@
...
+            var exception = Assert.Throws<ArgumentNullException>(
+                () => this.builder.Object.UseServiceFabricIntegration(null, ServiceFabricIntegrationOptions.None));
+            Assert.Equal("listener", exception.ParamName);
+        }
+
+        /// <summary>
+        /// Verify WebHostBuilderExtension for ServiceFabricIntegrationOptions.None.
+        /// </summary>
+        [Fact]
+        public void VerifyWithServiceFabricIntegrationOptions_None()
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Test names should describe the SUT's observable behavior as an English sentence, not the test mechanics. `test.instructions.md` — prefer e.g. `RegistersStartupFilterWithoutUrlSuffixWhenOptionsIsNone` over `VerifyWithServiceFabricIntegrationOptions_None`. The `ThrowsArgumentNullExceptionWhen...` and `RegistersStartupFilter` names in this same file are good examples of the target style.

test/AspNetCore/HostBuilderServiceFabricExtensionTests.cs
```diff
@@ -0,0 +1,124 @@
...
+            this.builder.Object.UseServiceFabricIntegration(this.listener, ServiceFabricIntegrationOptions.None);
+            Assert.Equal(1, this.configureServicesCount);
+            Assert.Empty(this.listener.UrlSuffix);
+        }
+
+        /// <summary>
+        /// Verify WebHostBuilderExtension for ServiceFabricIntegrationOptions.UseUniqueServiceUrl.
+        /// </summary>
+        [Fact]
+        public void VerifyWithServiceFabricIntegrationOptions_UseUniqueServiceUrl()
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") Same as above: prefer a behavior-describing name such as `RegistersStartupFilterWithUrlSuffixWhenOptionsIsUseUniqueServiceUrl` over `VerifyWithServiceFabricIntegrationOptions_UseUniqueServiceUrl`.

test/AspNetCore/HostBuilderServiceFabricExtensionTests.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            var services = new ServiceCollection();
+            this.capturedConfigureServices(null, services);
+
+            Assert.Single(services, d => d.ServiceType == typeof(IStartupFilter));
+        }
+
+        private IHost BuildFunc(string url, AspNetCoreCommunicationListener listener)
+        {
+            return new Mock<IHost>().Object;
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") This returns a pass-through `IHost` that is never used with `Setup`/`Verify`/`Mock.Get`, so `moq.instructions.md` prefers the more concise `Mock.Of<IHost>()` over `new Mock<IHost>().Object`.

src/AspNetCore/README.md
```diff
@@ -20,8 +20,8 @@ Derive from `AspNetCoreCommunicationListener` to create a custom communication l
 ```csharp
 class MyCommunicationListener : AspNetCoreCommunicationListener
 {
-    public MyCommunicationListener(ServiceContext context, Func<string, AspNetCoreCommunicationListener, IWebHost> build)
-        : base(context, build) { }
+  public MyCommunicationListener(ServiceContext context, Func<string, AspNetCoreCommunicationListener, IHost> build)
+            : base(context, build) { }
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") The `IWebHost`→`IHost` change is correct, but the indentation regressed: `public` now sits at 2 spaces and `: base(...)` at 12, where the example previously used 4/8. Since this C# example renders in the NuGet package README, `docs.instructions.md` applies and `.editorconfig` specifies `indent_size = 4`. Please restore 4-space indentation.

src/AspNetCore/WebHostBuilderServiceFabricExtension.cs
```diff
@@ -55,5 +56,50 @@ public static IWebHostBuilder UseServiceFabricIntegration(this IWebHostBuilder h
...
             return hostBuilder;
         }
+
+        /// <summary>
+        /// Configures the Service to use ServiceFabricMiddleware and tells the listener that middleware is configured for the service so that it can
+        /// suffix PartitionId and ReplicaOrInstanceId to url before providing it to Service Fabric Runtime.
+        /// </summary>
+        /// <param name="hostBuilder">The Microsoft.Extensions.Hosting.IHostBuilder to configure.</param>
+        /// <param name="listener">The <see cref="AspNetCoreCommunicationListener"/> to configure.</param>
+        /// <param name="options">Options to configure ServiceFabricMiddleware and AspNetCoreCommunicationListener.</param>
```
olegsych: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") The `<summary>`, `<param name="listener">`, and `<param name="options">` here are byte-for-byte identical to the existing overload, and the `hostBuilder`/`returns` text uses plaintext fully-qualified type names. `docs.instructions.md` recommends reusing shared docs via `<inheritdoc cref="..."/>` (path-based, e.g. `<inheritdoc path="/param[@name='listener']" cref="..."/>`, since the overloads differ in `hostBuilder`/`returns`) and using `<see cref="IHostBuilder"/>` for navigable references instead of raw `Microsoft.Extensions.Hosting.IHostBuilder` text. The namespace is already imported, so `<see cref="IHostBuilder"/>` resolves without new usings.

olegsych (changes requested):
[:copilot:](https://docs.github.com/copilot/responsible-use/code-review) **⚠️ Needs Changes** — The production change is correct, minimal, and well-motivated: it faithfully mirrors the existing `IWebHostBuilder` overload onto the modern `IHostBuilder` abstraction and even improves on it (explicit `listener` null-check via `nameof`). No behavioral bugs were found and CI is green. All requested changes are in the new test file and docs — none are merge-blocking bugs, so a human reviewer may reasonably accept the test-file structure to stay consistent with its pre-existing sibling.

Reviewed independently by three models, then cross-checked; all findings received unanimous agreement.

### Notes (not blocking)

- 📝 The existing `IWebHostBuilder` overload doesn't null-check `listener`, while the new `IHostBuilder` code does (correctly using `nameof`) — a minor parity gap on the old route, out of scope here.
- 📝 The HttpSys README's generic-host sample still goes straight from `Host.CreateDefaultBuilder()` to `ConfigureWebHostDefaults(...)`; since `HttpSysCommunicationListener` also has the `IHost` constructor, aligning it with the new Kestrel example may be worth a follow-up.
- 📝 `WebHostBuilderServiceFabricExtension` now also carries `IHostBuilder` extensions; renaming a public type is a breaking change, so not actionable here — the PR appropriately updated the class `<summary>` to mention both builders.

@gobradovic please take a look at the inline comments.

src/AspNetCore/WebHostBuilderServiceFabricExtension.cs
olegsych: I don't think this is a good idea because we will likely remove the IWebHostBuilder extension in the next major SDK release.

src/AspNetCore/WebHostBuilderServiceFabricExtension.cs
```diff
@@ -55,5 +56,50 @@ public static IWebHostBuilder UseServiceFabricIntegration(this IWebHostBuilder h
...
+        /// <returns>The Microsoft.Extensions.Hosting.IHostBuilder.</returns>
+        public static IHostBuilder UseServiceFabricIntegration(this IHostBuilder hostBuilder, AspNetCoreCommunicationListener listener, ServiceFabricIntegrationOptions options)
```
olegsych: Could you move this method to a new class - `IHostBuilderExtensions`? The old class name is already inaccurate and will be incorrect once the IWebHostBuilder extension method is removed. By putting the new overload in a clean separate class we'll enable simply deleting the old class entirely in the future.

test/AspNetCore/HostBuilderServiceFabricExtensionTests.cs
olegsych: This class needs to be renamed to `IHostBuilderExtensionsTest` once the new method is moved to a new class `IHostBuilderExtensions`.

test/AspNetCore/HostBuilderServiceFabricExtensionTests.cs
olegsych: Could you switch to fuzzy.StatelessServiceContext()? I think we should be able to use it instead of the legacy TestMocksRepository.

test/AspNetCore/HostBuilderServiceFabricExtensionTests.cs
```diff
@@ -0,0 +1,124 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.ServiceFabric.Services.Communication.AspNetCore;
+using Moq;
+using Xunit;
+
+namespace Microsoft.ServiceFabric.AspNetCore.Tests
+{
+    /// <summary>
```
olegsych: Could you remove the doc comments? We shouldn't add them to the tests.

src/AspNetCore.Kestrel/README.md
```diff
@@ -23,13 +23,16 @@ class MyService : StatelessService
...
                     Host.CreateDefaultBuilder()
+                        .UseServiceFabricIntegration(listener, ServiceFabricIntegrationOptions.UseUniqueServiceUrl)
                         .ConfigureWebHostDefaults(webBuilder => webBuilder.UseStartup<Startup>().UseUrls(url))
                         .Build()))
         ];
     }
 }
 ```
 
+This host-based path avoids obsolete `WebHostBuilder` APIs on newer ASP.NET Core versions.
```
olegsych: Could you remove this comment - I think it's redundant. The example above doesn't even mention the IWebHostBuilder.

src/AspNetCore.Kestrel/README.md
```diff
@@ -23,13 +23,16 @@ class MyService : StatelessService
             new ServiceInstanceListener(context =>
                 new KestrelCommunicationListener(context, "ServiceEndpoint", (url, listener) =>
                     Host.CreateDefaultBuilder()
+                        .UseServiceFabricIntegration(listener, ServiceFabricIntegrationOptions.UseUniqueServiceUrl)
```
olegsych, line 26: I think we need to add this to the `AspNetCore.HttpSys/README.md` as well.

olegsych (changes requested):
A few more requests from manual review.

olegsych (approved):
[:copilot:](https://docs.github.com/copilot/responsible-use/code-review) Approving — the implementation correctly mirrors the existing `IWebHostBuilder` behavior onto `IHostBuilder`, resolving the .NET 10 obsolescence warnings, and the prior review feedback was addressed. CI is green and all tests pass across `net8.0`/`net9.0`/`net10.0`/`net472`. The remaining items below are minor (a stale PR description, a docs convention, and two small test-coverage improvements) and are left to the author's discretion.

### Detailed Assessment

The change adds an `IHostBuilder` overload of `UseServiceFabricIntegration` in the new `IHostBuilderExtensions` class, matching the semantics of the existing `IWebHostBuilder` extension: it registers the `ServiceFabricSetupFilter` startup filter, is idempotent via a `Properties` key guard, and returns the builder for chaining. The approach is sound and consistent with the existing pattern.

### Issues

#### ⚠️ Should Fix — PR description names the wrong class

The PR body says the overload was added in `WebHostBuilderServiceFabricExtension`, but the implementation lives in the new `IHostBuilderExtensions` class (`WebHostBuilderServiceFabricExtension.cs` is not in the diff). Per the CONTRIBUTING guidance, please update the description to reflect the actual change.

#### ⚠️ Should Fix — Use `<see cref>` instead of plaintext type names in XML docs

In `src/AspNetCore/IHostBuilderExtensions.cs` the XML comments refer to the host builder as plaintext `Microsoft.Extensions.Hosting.IHostBuilder` (class summary, `hostBuilder` param, returns). `docs.instructions.md` prefers `<see cref="..."/>` over plain text for first-class abstractions. The `Microsoft.Extensions.Hosting` namespace is already imported, so `<see cref="IHostBuilder"/>` resolves without new usings.

#### 💡 Consider — Cover the idempotent early-return branch explicitly

`RegistersStartupFilterWithoutUrlSuffixWhenOptionsIsNone` and its `UseUniqueServiceUrl` sibling each call `UseServiceFabricIntegration` twice and assert `configureServicesCount == 1`. Per `test.instructions.md` (a separate test per logical branch), folding the `Properties.ContainsKey(SettingName)` early-return out into a dedicated idempotency test would map more cleanly to the SUT.

#### 💡 Consider — Assert the returned builder for chaining

No test verifies that `UseServiceFabricIntegration` returns the same `IHostBuilder` it was given, which is part of the fluent contract. An `Assert.Same(builder.Object, result)` would document it.

#### 📝 Note — `SettingName` could be `const string` / `nameof`

`private static readonly string SettingName = "UseServiceFabricIntegration";` is a compile-time literal — `const string` is more appropriate, and `nameof(UseServiceFabricIntegration)` would keep it refactor-safe. Carried over verbatim from the legacy class; out of scope, FYI.

olegsych (approved):

PR #616, opened 2026-07-08
Remove NetDataContractSerializer from WCF exception handling

olegsych:
The `NetDataContractSerializer` is inherently unsafe during deserialization because it relies on the type information embedded in the serialized data. We're removing it from the WCF stack to eliminate this security vulnerability.

On the server side, `WcfGlobalErrorHandler` will use only `DataContractSerializer` for exception serialization. This keeps it backward compatible with the older WCF clients that fallback from `NetDataContractSerializer` to `DataContractSerializer`. 

On the client side, `WcfExceptionHandler` will parse the full name of the exception type from the XML generated by the `DataContractSerializer` and, for backward compatibility with the older servers, the `NetDataContractSerializer` as well.

The `WcfRemoteExceptionInformation` type was reduced to fault codes. Exception serialization logic was moved to the `WcfGlobalErrorHandler`. Exception deserialization logic was replaced by XML parsing in the `WcfExceptionHandler`.

The `Microsoft.ServiceFabric.Services.Communication.Client.ExceptionHandlingRetryResult` was modified to expose an internal constructor taking `string exceptionId` instead of the `Exception` parameter of the public constructor.

All classes except `WcfRemoteExceptionInformation` were put under test.

The `<RootNamespace>` of `.Services.Wcf.Tests.csproj` was updated the actual namespace of the product project to make sure the newly added test files have no style warnings. The product project wasn't updated because that would require more changes to disambiguate the `SR` classes that would have the same full name as the one in  `.Services.dll`.


src/Services.Wcf/Communication/Wcf/WcfRemoteExceptionInformation.cs
```diff
@@ -1,158 +1,19 @@
...
-                    CultureInfo.CurrentCulture,
-                    Microsoft.ServiceFabric.Services.Wcf.SR.ErrorExceptionSerializationFailed2,
-                    exception);
+using System;
+using System.Globalization;
+using System.IO;
+using System.Runtime.Serialization;
+using System.ServiceModel;
+using System.Text;
+using System.Xml;
```
gobradovic: `System`, `System.Globalization`, `System.IO`, `System.Runtime.Serialization`, `System.Text` and `System.Xml` seem unused.

gobradovic:
❗ PR description incorrectly claims `ExceptionId` is "used only for tracing"

The PR description states: *"The ExceptionId is used only for tracing."* This is incorrect. `ExceptionId` is used in `Utility.cs` `ShouldRetryOperation` to track retry counts per exception type, called from `ServicePartitionClient.cs`. The retry logic compares consecutive `ExceptionId` values — if they differ, the retry counter resets.

The old `ExceptionId` was a type name (e.g. `"System.InvalidOperationException"` or `"ServiceException"` on fallback). The new value is a full serialized XML blob of `ServiceExceptionData`, which includes the exception message and stack trace via `exception.ToString()`. This changes the retry tracking granularity:
- If the exception message or stack trace varies between retries (e.g. includes timestamps, request IDs, or varying inner exceptions), the counter resets every time, effectively bypassing `MaxRetryCount`.
- For stable errors (same code path, same message), the behavior is equivalent.

gobradovic (changes requested):

src/Services.Wcf/Communication/Wcf/WcfRemoteExceptionInformation.cs
olegsych: Removed

src/Services.Wcf/Communication/Wcf/Client/WcfExceptionHandler.cs
```diff
@@ -8,6 +8,8 @@ namespace Microsoft.ServiceFabric.Services.Communication.Wcf.Client
     using System;
     using System.ServiceModel;
```
olegsych, line 9: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") This file was substantially modified but keeps the legacy `{}`-scoped namespace with `using` directives inside it, while its sibling [WcfGlobalErrorHandler.cs](https://github.com/microsoft/service-fabric-dotnet/blob/develop/src/Services.Wcf/Communication/Wcf/Runtime/WcfGlobalErrorHandler.cs) was converted to a file-scoped namespace. [csharp.instructions.md](https://github.com/microsoft/service-fabric-dotnet/blob/develop/.github/instructions/csharp.instructions.md) prefers file-scoped namespaces, `.editorconfig` flags both file-scoped namespaces and outside-namespace usings as warnings, and [coding.instructions.md](https://github.com/microsoft/service-fabric-dotnet/blob/develop/.github/instructions/coding.instructions.md) applies style to "code being substantially modified." Converting it would make the two files consistent (at the cost of a larger diff).

src/Services.Wcf/Communication/Wcf/Runtime/WcfGlobalErrorHandler.cs
```diff
@@ -3,47 +3,54 @@
...
         }
+    }
 
-        public bool HandleError(Exception error)
-        {
-            if (error is FaultException)
-            {
-                return false;
-            }
+    static FaultReason FaultReason(Exception exception)
```
olegsych, line 41: [:copilot:](https://docs.github.com/copilot/responsible-use/code-review) [⚠️](# "Should fix") The method name `FaultReason` shadows the `System.ServiceModel.FaultReason` type it constructs and returns, so the signature reads `FaultReason FaultReason(Exception exception)` and the call site `new(FaultReason(error), ...)` is easy to misread. Consider a distinct name such as `CreateFaultReason`, consistent with the "make the code as informative and intuitive as possible" guidance in [coding.instructions.md](https://github.com/microsoft/service-fabric-dotnet/blob/develop/.github/instructions/coding.instructions.md).

olegsych (commented):
[:copilot:](https://docs.github.com/copilot/responsible-use/code-review) This is a well-executed security fix that removes the insecure `NetDataContractSerializer` from the WCF fault path; the approach is sound, backward-compatible in both directions, CI is green, and the tests are thorough. The earlier "make `TryParseExceptionId` static" finding is now addressed. Two minor style/readability findings remain, posted inline. @olegsych please take a look.

src/Services.Wcf/Communication/Wcf/Client/WcfExceptionHandler.cs
olegsych: Unlike the WcfGlobalErrorHandler.cs, which was essentially rewritten, this file was only edited. I think this is out of scope.

src/Services.Wcf/Communication/Wcf/Runtime/WcfGlobalErrorHandler.cs
olegsych: Create/New/etc. are mere filler words because a non-void method is expected to return something. It is also very close to the type-casting syntax `(FaultReason)error`, in fact C++ supports this style of casting directly and I think it is very readable. So no, I don't want to do this.

gobradovic (approved):

PR #620, opened 2026-09-02
User/orepetrovi/capacity release http apis

orepetrovi:
## Summary

Adds generated HTTP client APIs, models, serializers, and PowerShell cmdlets for Service Fabric Capacity Release operations.

## What changed

- Added APIs for retrieving the current cluster capacity-release level.
- Added paged capacity-release estimation APIs with continuation-token and maximum-result support using the standard paged client type.
- Added the cluster-wide operation for setting the capacity-release level.
- Added generated models and JSON converters for capacity-release levels, actions, and estimates.
- Documented the service availability and state implications of each capacity-release level across generated client and PowerShell surfaces.
- Projected capacity-release configuration through generated service create, update, query, serialization, and PowerShell surfaces.
- Updated generated service-tag dependencies used by the service create and update cmdlets.
- Targeted Service Fabric REST API version 12.0.

## PowerShell safety

`Set-SFCapacityReleaseLevel` is declared with high confirmation impact and supports `ShouldProcess`. It presents level-specific action text against the Service Fabric cluster target:

- `None` restores configured replica, instance, minimum, and auxiliary targets.
- `Minor` reduces DropToZero services to configured minimum and auxiliary targets to zero, warning of greater quorum-loss exposure.
- `Major` reduces DropToMin services to configured minimum, DropToZero services to zero, and auxiliary targets to zero, warning that services at zero become unavailable and stateful services at zero permanently lose data.

`-WhatIf` and declined confirmation return before the cluster client is accessed, preventing the cluster-wide POST. Automation can explicitly use `-Confirm:$false`.

## Validation

- Swagger expansion, semantic validation, C# and Python generation, and client-library generation completed successfully.
- Release graph rebuild of `Microsoft.ServiceFabric.Powershell.Http.csproj`: 0 warnings and 0 errors.
- Generated output scope verified: exactly 28 generated paths.
- Generated files match the authoritative artifact except for the intentional `Set-SFCapacityReleaseLevel` confirmation-safety delta.


src/Client.Http/Client.Http/Generated/ClusterClient.cs
```diff
@@ -826,6 +826,95 @@ HttpRequestMessage RequestFunc()
             return this.httpClient.SendAsyncGetResponse(RequestFunc, url, ClusterLoadInfoConverter.Deserialize, requestId, cancellationToken);
         }
 
+        /// <inheritdoc />
+        public Task<CapacityReleaseLevelResult> GetCapacityReleaseLevelAsync(
```
olegsych, line 830: This is the extra object wrapping level I mentioned in the Swagger PR.

src/Client.Http/Common/Generated/CapacityReleaseEstimate.cs
```diff
@@ -0,0 +1,70 @@
...
+            this.TotalCapacity = totalCapacity;
+        }
+
+        /// <summary>
+        /// Gets the capacity release scenario whose projected impact this estimate represents. Possible values include:
+        /// 'None', 'Minor', 'Major'
+        /// 
+        /// The level of capacity release applied to the cluster.
+        /// </summary>
+        public CapacityReleaseLevel? Level { get; }
```
olegsych, line 52: Could you double-check whether this code was generated after the properties were defined as required in the swagger? I don't think they should be nullable.

src/Client.Http/Common/Generated/CapacityReleaseLevelResult.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Microsoft.ServiceFabric.Common
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// The current capacity release level for the cluster.
+    /// </summary>
+    public partial class CapacityReleaseLevelResult
```
olegsych, line 14: I hope we can eliminate this class.

olegsych (changes requested):
Could you update the PR title and remove agent-generated filler from the description? It should explain what is changing and why, not how the code was built or tested.

